PR #7, opened 2026-04-10
Fix Office startup readiness checks

free2518:
## Summary
- wait for the local Office page to become reachable before reporting Claw3D as running
- open the embedded webview at `/office` directly instead of the root route
- write Hermes-specific gateway defaults so first-run Office launches do not fall back to OpenClaw defaults

## Why
On my setup, the Office tab could open to a blank page on first launch and only render after a manual refresh. I cannot confirm how broadly this reproduces, so this may be somewhat environment-specific.

## Validation
- `npm run typecheck`
- verified the desktop now waits for `/office` readiness before marking the service as running

## Notes for Reviewers
- This patch was prepared with AI assistance.
- Because the original bug may be environment-specific, please inspect the readiness and config changes carefully before merging.

### DIFF
--- a/src/main/claw3d.ts
+++ b/src/main/claw3d.ts
@@ -1,4 +1,5 @@
 import { spawn, ChildProcess, execSync } from "child_process";
+import { request as httpRequest } from "http";
 import {
   existsSync,
   readFileSync,
@@ -20,6 +21,9 @@ const WS_URL_FILE = join(HERMES_HOME, "claw3d-ws-url");
 const DEFAULT_PORT = 3000;
 const DEFAULT_WS_URL = "ws://localhost:18789";
 const CLAW3D_SETTINGS_DIR = join(homedir(), ".openclaw", "claw3d");
+const CLAW3D_HTTP_READY_TIMEOUT_MS = 45_000;
+const CLAW3D_HTTP_STATUS_TIMEOUT_MS = 1_500;
+const CLAW3D_HTTP_POLL_INTERVAL_MS = 500;
 
 let devServerProcess: ChildProcess | null = null;
 let adapterProcess: ChildProcess | null = null;
@@ -27,6 +31,10 @@ let devServerLogs = "";
 let adapterLogs = "";
 let devServerError = "";
 let adapterError = "";
+
+function isRecord(value: unknown): value is Record<string, unknown> {
+  return Boolean(value && typeof value === "object" && !Array.isArray(value));
+}
 
 function getSavedPort(): number {
   try {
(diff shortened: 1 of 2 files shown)


fathah:
Thanks for the patch! Reviewed and tested locally. Findings:

**Looks good ✅**
- Typecheck and production build are clean.
- ESLint introduces no new errors on the touched files (just 2 minor prettier whitespace warnings — feel free to run `npx prettier --write src/main/claw3d.ts` to clear them).
- The new HTTP readiness probe (`probeHttpOfficeReady`) and `waitForClaw3dReady` logic look correct, including the redirect-tolerant 2xx/3xx check and the local-only gateway probe.
- IPC handler at `src/main/index.ts:519` works fine with `startAll()` becoming async — `ipcMain.handle` auto-awaits the returned Promise, no caller change needed.
- Backwards-compat preservation of existing `gateway.token` and merging of `lastKnownGood` is handled cleanly.

**Please address before I merge ❌**
1. **Rebase needed** — `package.json` rolls the version back from `0.2.1` → `0.2.0`. This branch was based on `0.2.0`, so a rebase on latest `main` will fix it automatically.
2. **(Nice-to-have)** During the up-to-45s `waitForClaw3dReady`, the renderer just shows "starting" with no progress feedback. Could you either (a) lower the default timeout, or (b) emit an IPC progress event during the wait so the UI can show "waiting for Office page (Xs)…"? Not a blocker, but feels like a long silent wait on slow machines.
3. **(Question)** The settings migration is one-way — `delete settings.adapter; delete settings.url; delete settings.token;` strips top-level keys forever. Intended? If a user downgrades to a previous build, they'd lose their settings. Worth a `schemaVersion` field in `settings.json` to make this explicit.

Once rebased I'll re-run the smoke test and merge.

pmos69:
Thank you for opening this and for pointing at the right class of problem. I took a careful look because Office has moved quite a bit since this branch was opened, and parts of this PR were already handled elsewhere while other parts were still very relevant.

I opened a fresh replacement PR here: #437

The new PR preserves the still-useful pieces from this one, but rebuilds them against current `main` and current Hermes Office behavior:

- Desktop now writes `CLAW3D_GATEWAY_ADAPTER_TYPE=hermes` for Office.
- Desktop now writes the modern nested `gateway` settings shape used by current Hermes Office.
- Desktop opens Office directly at `/office`.
- Desktop waits for the Office `/office` route before treating local Office startup as successful.
- Desktop remounts the Office webview on stop/start to avoid a stale blank webview.

During live testing we also found a couple of related edge cases that the replacement PR handles: stale nested `gateway.adapterType` / `gateway.lastKnownGood` values could keep Office on the manual Connect panel, and stop/start could show Desktop as `Running` while the webview was blank.

I am closing this PR as superseded by #437, not because the report was wrong. The original direction was useful; the implementation just needed to be rebuilt against the current Office startup/settings flow. Thanks again for the contribution.

PR #10, opened 2026-04-12
fix: recognize oauth profiles and authenticate desktop chat to local api server

bharatindia123-a11y:
Title: fix: recognize oauth profiles and authenticate desktop chat to local api server

Summary
- treat OAuth-backed providers such as `openai-codex` as configured in the desktop install gate
- resolve `auth.json` against the active Hermes profile instead of only root `~/.hermes`
- support credential lookup from both `providers` token state and `credential_pool`
- when the local Hermes API server is protected by `API_SERVER_KEY`, send `Authorization: Bearer <API_SERVER_KEY>` from the desktop app's local API client

Problem
Hermes Desktop currently treats the installation as "configured" only when:
- a supported API key exists in `.env`, or
- the active provider is a local/custom endpoint

That blocks already-authenticated OAuth-backed Hermes setups (for example `openai-codex`) and incorrectly sends users into the API-key setup flow even though Hermes is already ready to chat.

A second issue appears even after setup passes:
- if the local Hermes API server is protected by `API_SERVER_KEY`, desktop chat requests to `http://127.0.0.1:8642` fail with `Invalid API key`

Reproduction
1. Log into Hermes CLI with an OAuth-backed provider such as `openai-codex`
2. Store the active session under a named profile (for example `~/.hermes/profiles/generic-core/auth.json`)
3. Enable the local API server with `API_SERVER_KEY`
4. Launch Hermes Desktop
5. Observe either:
   - the desktop app routes the user into setup/API-key onboarding instead of treating the install as ready, or
   - chat fails with `Invalid API key`

Expected behavior
- if the active Hermes profile already has valid OAuth credentials, Hermes Desktop should treat the install as configured and proceed normally
- if the local API server requires `API_SERVER_KEY`, Hermes Desktop should authenticate to it automatically

Root cause
- `checkInstallStatus()` in `src/main/installer.ts` only checks `.env` API keys and local providers
- auth lookup in `src/main/config.ts` only reads `~/.hermes/auth.json`, not the active profile's `auth.json`
- the local API fast-path in `src/main/hermes.ts` calls `http://127.0.0.1:8642/v1/*` without sending the configured `API_SERVER_KEY`

What changed
1. `src/main/config.ts`
- made `auth.json` path resolution profile-aware
- added `getActiveProfileNameSync()`
- updated credential pool helpers to accept an optional profile
- added `hasOAuthCredentials(provider, profile?)`
- added `getApiServerKey(profile?)`
- `hasOAuthCredentials()` checks:
  - `providers[provider].access_token`
  - `providers[provider].refresh_token`
  - `providers[provider].api_key`
  - `credential_pool[provider][].{api_key,access_token,refresh_token}`
- for named profiles, it also falls back to root `~/.hermes/auth.json`
- `getApiServerKey()` checks profile config/env first, then root config/env

2. `src/main/installer.ts`
- `checkInstallStatus()` now resolves the active profile first
- considers the active profile's `.env` and `auth.json`
- accepts OAuth-backed providers via `hasOAuthCredentials()`
- returns the active profile in the status payload for easier debugging

3. `src/main/hermes.ts`
- reads `API_SERVER_KEY` from config / env using the active profile when possible
- adds `Authorization: Bearer <API_SERVER_KEY>` on local requests to `http://127.0.0.1:8642`
- uses the same auth headers for the non-streaming probe request

Why this is safe
- existing API-key flows are unchanged
- local/custom providers still bypass API-key requirements as before
- the new logic broadens the definition of "configured" to include already-authenticated OAuth-backed providers
- API server auth is only forwarded to the local Hermes API fast-path and only when `API_SERVER_KEY` is configured
- fallback to root auth/config preserves compatibility for installs where shared state still lives under `~/.hermes`

Verification
- `npm run typecheck:node`
- `npm run typecheck`
- `npm run build`
- `npm run build:unpack`
- verified against a live machine state where:
  - active profile = `generic-core`
  - provider = `openai-codex`
  - OAuth credentials exist in `auth.json`
  - `API_SERVER_KEY` is enabled in Hermes config
  - the patched install gate evaluates to pass
  - an authenticated local request to `http://127.0.0.1:8642/v1/chat/completions` succeeds

Notes
This was validated against a local source-built macOS app bundle and the installed app was replaced with the source-built binary to confirm parity.

🤖 This PR was created by an AI coding agent
🤖 This was written by an AI coding agent

### DIFF
--- a/src/main/config.ts
+++ b/src/main/config.ts
@@ -222,6 +222,22 @@ export function getHermesHome(profile?: string): string {
   return profilePaths(profile).home;
 }
 
+export function getApiServerKey(profile?: string): string {
+  const candidates = [
+    getConfigValue("API_SERVER_KEY", profile),
+    profile && profile !== "default" ? getConfigValue("API_SERVER_KEY") : null,
+    readEnv(profile).API_SERVER_KEY || null,
+    profile && profile !== "default" ? readEnv().API_SERVER_KEY || null : null,
+  ];
+
+  for (const candidate of candidates) {
+    const value = String(candidate || "").trim();
+    if (value) return value;
+  }
+
+  return "";
+}
+
 // ── Platform enabled/disabled in config.yaml ────────────
 
 const SUPPORTED_PLATFORMS = ["telegram", "discord", "slack", "whatsapp", "signal"];
--- a/src/main/config.ts
+++ b/src/main/config.ts
@@ -305,20 +305,34 @@ export function setPlatformEnabled(
   safeWriteFile(configFile, content);
 }
 
-// ── Credential Pool (auth.json) ──────────────────────────
+// ── Credential Pool / OAuth store (auth.json) ─────────────────────────
 
-function authFilePath(): string {
-  return join(HERMES_HOME, "auth.json");
+function getActiveProfileNameSync(): string {
+  try {
+    const activeFile = join(HERMES_HOME, "active_profile");
+    if (!existsSync(activeFile)) return "default";
+    const name = readFileSync(activeFile, "utf-8").trim();
+    return name || "default";
+  } catch {
+    return "default";
+  }
+}
+
+function authFilePath(profile?: string): string {
+  return join(profileHome(profile || getActiveProfileNameSync()), "auth.json");
 }
 
 interface CredentialEntry {
-  key: string;
-  label: string;
+  key?: string;
+  api_key?: string;
+  access_token?: string;
+  refresh_token?: string;
+  label?: string;
 }
 
-function readAuthStore(): Record<string, unknown> {
+function readAuthStore(profile?: string): Record<string, unknown> {
   try {
-    const p = authFilePath();
+    const p = authFilePath(profile);
     if (!existsSync(p)) return {};
     return JSON.parse(readFileSync(p, "utf-8"));
   } catch {
@@ -326,12 +340,14 @@ function readAuthStore(): Record<string, unknown> {
   }
 }
 
-function writeAuthStore(store: Record<string, unknown>): void {
-  safeWriteFile(authFilePath(), JSON.stringify(store, null, 2));
+function writeAuthStore(store: Record<string, unknown>, profile?: string): void {
+  safeWriteFile(authFilePath(profile), JSON.stringify(store, null, 2));
 }
 
-export function getCredentialPool(): Record<string, CredentialEntry[]> {
-  const store = readAuthStore();
+export function getCredentialPool(
+  profile?: string,
+): Record<string, CredentialEntry[]> {
+  const store = readAuthStore(profile);
   const pool = store.credential_pool;
   if (!pool || typeof pool !== "object") return {};
   return pool as Record<string, CredentialEntry[]>;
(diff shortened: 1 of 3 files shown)


fathah:
Thanks — reviewed and the static checks pass. Want to run the live tests on my end before merging since this touches both the OAuth gate and the API-server auth path.

**Static review ✅**
- Typecheck and production build are clean.
- Lint adds no new errors (just 8 prettier whitespace warnings — `npx prettier --write` will clear them).
- `Authorization: Bearer <API_SERVER_KEY>` is correctly scoped to `127.0.0.1:8642` only and is skipped when the key is empty, so existing setups without `API_SERVER_KEY` are unaffected.
- Backwards-compat looks good: existing function signatures already accepted `profile?`, `CredentialEntry` was widened with all-optional fields, and the install gate's API-key path is unchanged.

**Could you address before merge ❌**

1. **`getActiveProfileNameSync()` is now defined three times** — `installer.ts:44`, `config.ts:326`, and there's already an async version in `profiles.ts:87`. Could you extract one shared helper (sync) and import it from the others? Less drift risk down the line.

2. **`InstallStatus & { activeProfile?: string }` intersection** at `installer.ts:67` — please add `activeProfile?: string` to the `InstallStatus` interface itself (and to the matching declaration in `src/preload/index.d.ts:3-8`). Otherwise the renderer can't see the new field via TypeScript and the type intersection feels like a workaround.

3. **Implicit behavior change in `getCredentialPool()` / `setCredentialPool()`** — these now default to the *active profile's* `auth.json` instead of root `~/.hermes/auth.json`. For default-profile users it's a no-op (because `profileHome("default") === HERMES_HOME`), but for users on a named profile, the Settings UI will now read/write a different file than before. Could you call this out in the PR description? Also, the IPC handler at `src/main/index.ts:467` doesn't pass profile, so the renderer is implicitly bound to "whatever profile is active right now" — please confirm that's intended.

4. **(Optional perf)** `getApiServerKey()` is called in `sendMessageViaApi` on every chat message and every probe, and it re-reads `config.yaml` + `.env` each time. Could you route it through the existing 5s TTL cache in `config.ts`? Minor, but you already pay that cost in `readEnv()`.

**End-to-end tests I'm running locally before merging**
- OAuth-only profile (`openai-codex` under a named profile) → install gate passes without prompting for API key
- `API_SERVER_KEY` enabled → desktop chat succeeds against local API server (vs. prior "Invalid API key")
- Existing API-key flow (no OAuth, no `API_SERVER_KEY`) → still works, no regression
- Credential pool UI on both default and named profiles

I'll merge after addressing 1–3 and the live tests pass.

YotsuBotfds:
Confirming this still affects Hermes Desktop v0.3.4 on macOS Apple Silicon.

I have Hermes CLI working with `openai-codex` OAuth / ChatGPT subscription auth, but Desktop stayed stuck on the provider/API-key setup screen because the first-run install/config check did not treat `openai-codex` as configured.

Local validation:

- Hermes CLI `openai-codex` auth worked before and after the Desktop patch.
- Desktop v0.3.4 kept routing to provider setup with `provider: openai-codex` in `~/.hermes/config.yaml` and valid OAuth credentials in `auth.json`.
- Locally patching the Desktop install gate so the configured-provider list includes `openai-codex` let Desktop proceed without requiring an OpenAI API key.
- No dummy `OPENAI_API_KEY` was needed once the install gate recognized `openai-codex`.

So this PR appears to address a real current blocker for users relying on ChatGPT/Codex subscription auth rather than direct OpenAI API keys. Thanks for the work on it.

pmos69:
Hi @bharatindia123-a11y — I noticed this PR has been inactive for a while and is currently in CONFLICTING state against `main`. To unblock the fixes (both of which I needed locally), I've opened #241 which rebases both your commits onto current `main` and addresses the outstanding review feedback (1–4). You are credited via `Co-authored-by` on both commits.

If you'd prefer to land this from your own branch, the work in #241 is happy to be closed in favor of yours — just leaving a heads-up so neither effort gets duplicated.

pmos69:
Hi @bharatindia123-a11y, thanks again for this PR. The work here was rebased, updated for current `main`, and merged in #241, with your commits credited via `Co-authored-by`.

Since this branch is now superseded by the merged PR, I’m going to close this one to keep the PR list tidy. Appreciate you identifying and fixing the OAuth/profile + API server auth path.

PR #11, opened 2026-04-13
Add shared i18n infrastructure

foqianyilishazi:
## Summary
- add a shared i18n runtime, locale service, and React provider/hook for Hermes Desktop
- add Vitest-based i18n tests and wire locale APIs through main, preload, and renderer
- keep English as the default runtime language with no end-user language behavior change

## Test Plan
- [x] npm run test
- [x] npm run typecheck


foqianyilishazi:
Thanks for taking a look.

I split this work into two PRs to keep review scope smaller:

- #11 adds the shared i18n infrastructure only
- #12 adds the zh-CN localization and Chinese docs on top of it

For this first PR, I intentionally kept English as the default runtime language so there is no end-user behavior change in the base app. The goal here is just to introduce the localization foundation across renderer, preload, and main process in a way that future language packs can build on cleanly.

fathah:
Thanks for this clean, well-scoped foundation PR.

**Looks good ✅**
- Typecheck, build, and the new vitest harness all pass (3/3 tests).
- New deps install cleanly. No new high-severity vulns introduced (the 2 existing ones — `vite` and `lodash` — are already on `main`).
- IPC wiring (`get-locale` / `set-locale`) lines up across main, preload, and renderer.
- No actual UI strings migrated to `t()` yet, so end-user behavior really is unchanged — good call to land this separately from the localization additions in #12.

**Please address before merge ❌**

1. **Lint error in `I18nProvider.tsx:18`** — `react-refresh/only-export-components`. Move `I18nContext` (and the `I18nContextValue` type) into its own file (e.g. `I18nContext.ts`) and import it from both `I18nProvider.tsx` and `useI18n.ts`. Otherwise HMR full-reloads the app on every edit to that file.

2. **Drive-by reformatting in `src/main/index.ts` and `src/preload/index.ts`** — the cron-job, send-message, and set-env handlers are reformatted with no logic change. Could you either revert those hunks or split them into a separate format-only PR? It'll cause noisy merge conflicts otherwise.

   Side note: the `if ((isGatewayRunning() && key.endsWith("_API_KEY")) || ...)` change at `index.ts:223` is just adding parens — `&&` already binds tighter than `||`, so behavior is identical. (The original condition does have a latent bug — the `_TOKEN` / `HF_TOKEN` branches skip the `isGatewayRunning()` check — but that's out of scope for this PR.)

**Nice to have for the follow-up (#12)**
- `setAppLocale` only mutates an in-memory variable — no persistence. Fine for now since only `en` exists, but once a real second locale lands, you'll want to write it to disk (e.g., `~/.hermes/desktop_locale` or `localStorage`) and read it on startup.
- Renderer bundle grew +98 KB (+8.4%) from `i18next` + `react-i18next`. Worth keeping an eye on as locales are added — consider lazy-loading non-default-locale resource bundles.

After (1) and (2) I'll smoke-test the renderer (DevTools `getLocale()`/`setLocale("en")` round-trip + visual check that all English strings look the same) and merge. Will then review #12 on top.

foqianyilishazi:
Updated based on the review:

1. Moved  and  into a dedicated  file so  no longer exports non-component state.
2. Reverted the unrelated formatting churn in , , and , and kept only the locale wiring changes in those files.

Also propagated the same cleanup to the stacked localization branch so #12 stays aligned.

foqianyilishazi:
Updated based on the review:

1. Moved the i18n context and its value type into a dedicated I18nContext.ts file so I18nProvider.tsx no longer exports non-component state.
2. Reverted the unrelated formatting churn in src/main/index.ts, src/preload/index.ts, and src/preload/index.d.ts, and kept only the locale wiring changes in those files.

Also propagated the same cleanup to the stacked localization branch so #12 stays aligned.

PR #15, opened 2026-04-16
feat: complete Chinese (zh-CN) i18n translation

liudashuang:
## Summary

This PR completes the full Chinese (zh-CN) internationalization for Hermes Desktop.

## Changes

- **Complete Chinese translations**: Added comprehensive zh-CN translations for all UI components
- **Replaced hardcoded strings**: Updated all React components to use i18n t() function calls
- **Updated translation files**: Both common and domain-specific translation files updated for complete coverage

## Impact

- Makes Hermes Desktop accessible to Chinese-speaking users
- Maintains English as the default language for backward compatibility
- All user-facing text is now properly internationalized


fathah:
Thanks for the thorough zh-CN translation work — the locale content itself looks great and the keys line up cleanly with the English counterparts. Before we can merge, a few things need to be addressed:

**Blockers**

1. **`.eslintcache` is committed in this PR.** It contains your local absolute paths (`/Users/a1/Desktop/docker/www/hermes-desktop/...`). Please remove it from the diff and add `.eslintcache` to `.gitignore`.
2. **`src/shared/i18n/index.ts:94` drops optional chaining** — `resources[lang]?.translation` was changed to `resources[lang].translation`. This will throw if a persisted locale value ever doesn't match a loaded resource (older configs, migrations). Please restore the `?.`.

**Please revert (unrelated to i18n)**

3. The Prettier/formatter reflows in `src/main/index.ts` (around L237-240), `src/preload/index.ts`, `src/preload/index.d.ts`, and the `let` → `const` / regex / signature changes in `src/main/session-cache.ts` that aren't i18n-related. They're semantically fine but make the diff hard to review. Keep only the actual i18n wiring changes (the `t("sessions.newConversation", getAppLocale())` calls etc.).

**Nit**

4. In `src/shared/i18n/index.test.ts`, the existing "falls back to key when missing" test was replaced rather than kept alongside the new zh-CN test. Could you add that case back so we don't lose coverage of the fallback path?

The `tools.ts` refactor to `labelKey`/`descriptionKey` + `localizeToolDefs()` is a nice change and the keys are present in both locales — that one's good to keep.

Once the two blockers are fixed and the unrelated formatting churn is reverted, happy to merge. Thanks again!

liudashuang:
Thank you for your guidance. I have reworked the irrelevant changes and formatting issues.

PR #29, opened 2026-04-29
fix: add lock to prevent race condition in session cache

Ricardo-M-L:
## Summary

- Add `_cacheLock` boolean to serialize writes between `syncSessionCache()` and `updateSessionTitle()`
- Both functions read `sessions.json`, modify in memory, then write back — concurrent calls could overwrite each other's changes
- Also adds smoke test coverage for both functions

## Test plan

- [x] `pnpm test` (smoke test passes — real implementation requires HERMES_HOME/state.db)
- [x] Code review: lock is set before read, released in `finally` block in both functions

### DIFF
--- a/src/main/session-cache.ts
+++ b/src/main/session-cache.ts
@@ -76,83 +76,91 @@ function getDb(): Database.Database | null {
 
 // Sync from hermes DB to local cache — only fetches new/updated sessions
 export function syncSessionCache(): CachedSession[] {
-  const cache = readCache();
-  const db = getDb();
-  if (!db) return cache.sessions;
-
+  _cacheLock = true;
   try {
-    // Fetch sessions newer than last sync, or all if first sync
-    const rows = db
-      .prepare(
-        `SELECT s.id, s.started_at, s.source, s.message_count, s.model, s.title
-         FROM sessions s
-         WHERE s.started_at > ?
-         ORDER BY s.started_at DESC`,
-      )
-      .all(cache.lastSync > 0 ? cache.lastSync - 300 : 0) as Array<{
-      id: string;
-      started_at: number;
-      source: string;
-      message_count: number;
-      model: string;
-      title: string | null;
-    }>;
-
-    const existingIds = new Set(cache.sessions.map((s) => s.id));
-    let newSessions: CachedSession[] = [];
-
-    for (const row of rows) {
-      if (existingIds.has(row.id)) {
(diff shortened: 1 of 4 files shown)


fathah:
Merge Conflict

Ricardo-M-L:
On a closer re-review I want to flag that the \`_cacheLock\` boolean here doesn't actually prevent the race I claimed in the description, so I'd like your call on what to do.

Both \`syncSessionCache()\` and \`updateSessionTitle()\` are fully synchronous (\`better-sqlite3\` is sync, \`readCache\`/\`writeCache\` are sync \`fs\` calls, no \`await\` anywhere in the path). In single-threaded Node within a single Electron process there's no point at which one function can be interrupted mid-flight by the other, so a JS-level boolean flag doesn't add real protection — and the patch only sets the flag, never blocks/queues on it.

If the maintainer's mental model of the bug is **cross-process** (renderer triggering \`updateSessionTitle\` over IPC while main is in \`syncSessionCache\`), then a JS variable in main can't help — that would need either file-level locking on \`sessions.json\` or making both functions go through a serialized queue.

Options:

1. **Close this PR** — agreed there's no real race here, ditch the lock.
2. **Reframe as a serialized-queue fix** — replace the flag with an actual \`Promise\` chain that funnels both operations through one queue. Larger change, would re-open with a clean branch.
3. **Keep just the smoke test** — the test additions in \`tests/session-cache.test.ts\` are still useful even if the lock comes off.

I'd lean toward **option 1**. Let me know your preference and I'll execute.

(For context — this branch was also entangled with #44; that one has been force-pushed with only the Gateway/Install fixes and is clean now.)

pmos69:
Closing this as stale / not the right fix.

On re-review, the `_cacheLock` boolean does not actually serialize a real race here: the relevant `syncSessionCache()` / `updateSessionTitle()` paths are synchronous inside the Electron main process, and the PR would simply drop a title update if it happened while the flag was set. If there is a future cross-process or async cache-write race, the right shape would be an actual serialized queue or file-level locking.

The branch is also now significantly stale and conflicting, and would overwrite later session-cache fixes (profile-aware paths, stale message-count refresh, O(N) merge behavior, safe writes, etc.). Existing current coverage is much stronger in `tests/session-cache-sync.test.ts`; the smoke test added here is basically low-value by comparison. Thanks for the investigation — let's close this one rather than merge the lock.

PR #35, opened 2026-04-30
ci: add Windows (winget) and Fedora (RPM) release artifacts

Aiacos:
## Summary

- Adds a `release_windows` job that builds an NSIS installer (`hermes-desktop-<version>-setup.exe`) on `windows-latest`.
- Adds a `generate_winget` job that fills YAML manifest templates (`build/winget/*.template.yaml`) with the installer SHA256 and uploads them as the `winget-manifests-<version>` CI artifact, ready for manual submission to [`microsoft/winget-pkgs`](https://github.com/microsoft/winget-pkgs).
- Extends `release_linux` to also build an `.rpm` for Fedora alongside the existing `.AppImage` and `.deb`.
- Adds explicit `oneClick: true` / `perMachine: false` to NSIS (matches electron-builder defaults; pinning prevents future drift) and Linux packaging metadata (`vendor`, `synopsis`, `description`).
- Adds a `dry_run` boolean input to `workflow_dispatch` (default `true`) so the workflow can be tested on a branch without creating tags or releases. Real releases (push to `release` branch) are unaffected.

The winget manifests are deliberately **not** included in the GitHub Release files (operator-facing, not user-facing). To publish to winget after a release, download the `winget-manifests-<version>` artifact from the release run and submit a PR to `microsoft/winget-pkgs`.

Spec: `docs/superpowers/specs/2026-04-30-windows-winget-fedora-rpm-release-design.md`
Plan: `docs/superpowers/plans/2026-04-30-windows-winget-fedora-rpm-release.md`

## Verification

- [x] `npm run lint` — clean for the two new files (`scripts/generate-winget-manifests.mjs`, `tests/winget-generator.test.ts`)
- [x] `npm run typecheck` — clean (node + web)
- [x] `npm run test` — 5 new tests for the manifest generator pass; 2 pre-existing failures in `tests/constants.test.ts` are unrelated (i18n drift from ca075db, out of scope)
- [x] `npm run build:rpm` produces a valid `.rpm` on Fedora (verified locally; metadata: `Name: hermes-desktop`, `Vendor: Nous Research`)
- [x] `workflow_dispatch` with `dry_run=true` on the contributor's fork: all 6 build jobs succeed, `publish` skipped as designed (run completed in ~4 min)
- [x] Generated winget manifests inspected: no leftover placeholders, valid uppercase SHA256, correct URLs pointing at `fathah/hermes-desktop` releases

## Notes for the maintainer

[...]

PR #36, opened 2026-04-30
fix(office): show Docker setup guidance when in Remote mode

Aiacos:
## Summary

The Office (Claw3D) tab silently fails when the user is in Remote connection mode (Settings → Connection → Remote). Root cause: Office runs `hermes-office` locally, and its in-process `hermes-adapter` is hardcoded to talk to a co-located Hermes Agent. In remote mode there is no local Hermes Agent, so the adapter has nothing to bridge.

This PR doesn't fix the underlying limitation (that would need upstream `hermes-office` adapter work + a "Remote Office URL" setting in the desktop app) — it surfaces it clearly to the user instead of failing silently:

- New `'remote-mode'` state in `Office.tsx`, entered at status-check time when `isRemoteMode()` is true.
- An explanatory screen with two options:
  - **Switch to Local mode** — single button that calls `setConnectionConfig('local', '', '')` and refreshes.
  - **Self-host Office on the Docker side** — links to a new step-by-step guide.
- en + zh-CN i18n strings for the new screen.
- New guide `docs/guides/office-on-docker.md` with a sample `docker-compose.yml`, caveats, and explicit notes about what does and doesn't work today.
- Bundled investigation document `docs/issues/2026-04-30-runtime-connection-issues.md` as the rationale (it also covers a separate Issue 1 about remote-Hermes connection diagnostics, which is intentionally not addressed in this PR — a focused follow-up).

## What this is NOT fixing

- The desktop app's Office tab still cannot connect to a remote Office instance (e.g., one running in Docker). The guide is honest about this; users open `http://<docker-host>:3000` in a regular browser. Adding remote-Office-URL support in the desktop app + (likely) a `HERMES_AGENT_URL` env in upstream `hermes-office`'s adapter would be a future PR.
- Issue 1 (remote-connection diagnostics — Settings UI doesn't distinguish auth-failed from unreachable, no URL validation) is documented but not implemented here. Each is small and orthogonal; happy to do them as separate PRs.

## Verification

- [x] `npm run typecheck` clean (node + web)
- [x] `npm run test` — 237 pass, 2 pre-existing constants failures unrelated to this change
- [x] `npx eslint` on the three modified TS/TSX files: no NEW errors or warnings introduced. Pre-existing react-hooks errors at `Office.tsx:50-52` and pre-existing prettier warnings on long string literals are out of scope.
- [x] Visually walked the new render branch in the source — types match `OfficeState`, all i18n keys exist in both locales, links resolve to fathah/hermes-desktop docs/.

## Manual test plan (for the maintainer)

1. Local installation of the build (or `npm run dev`).
2. Settings → Connection → switch to Remote, enter any URL, save.
3. Open Office tab. Expected: the new "Office requires Local connection mode" screen with two option blocks.
4. Click **Switch to Local mode**. Expected: connection mode flips to local, Office tab re-checks and shows the normal "Set Up Claw3D" install screen (or the ready state, if already installed).
5. Switch back to Remote. Open Office. Expected: same guidance screen as step 3. Click **Docker setup guide** — opens the GitHub blob view of `docs/guides/office-on-docker.md` in the user's browser.

## Files changed

- `src/renderer/src/screens/Office/Office.tsx` (+76 / -7) — new state + render branch
- `src/shared/i18n/locales/en/office.ts` (+14) — new strings
- `src/shared/i18n/locales/zh-CN/office.ts` (+13) — new strings (machine-assisted; native review welcome)
- `docs/guides/office-on-docker.md` (new) — Docker deployment walkthrough
- `docs/issues/2026-04-30-runtime-connection-issues.md` (new) — investigation + future fix proposals

## Related

- Companion PR #35 — release pipeline (Windows winget + Fedora RPM). Independent of this one; can land in either order.

### DIFF
--- a/docs/guides/office-on-docker.md
+++ b/docs/guides/office-on-docker.md
@@ -1,0 +1,256 @@
+# Running Hermes Office (Claw3D) alongside Hermes Agent in Docker
+
+This guide describes how to deploy the **Office** module (`fathah/hermes-office`) in Docker
+together with a Dockerized **Hermes Agent**, so that users connecting to the agent over the
+network can also access the 3D visualization.
+
+> **Important limitation:** the desktop app's Office tab currently spawns Office's
+> `npm run dev` and `hermes-adapter` **locally** on the machine running the desktop app
+> and does not point its in-app webview at a remote Office instance. Following this guide
+> gives you Office in a browser at `http://<docker-host>:3000`, **not** inside the desktop
+> app's Office tab. The Office tab in the desktop app shows a guidance screen when in
+> remote mode and will not start the local Office stack.
+
+This is a known gap. If/when the desktop app gains support for a configurable remote
+Office URL, the same Docker deployment described here will work behind that URL field.
+
+---
+
+## Architecture
+
+When everything runs locally:
+
+```
+[ Desktop app  (Electron renderer + webview) ]
+        │ webview → http://localhost:3000
+        ▼
+[ hermes-office (Next.js dev server, port 3000) ]
+        │ ws://localhost:18789
+        ▼
+[ hermes-adapter (Node WebSocket bridge, port 18789) ]
+        │ in-process / IPC
+        ▼
+[ Hermes Agent (Python, API on 127.0.0.1:8642) ]
+```
+
+When deploying to Docker, you reproduce the **bottom three boxes** as containers and
+expose the Office UI on port 3000:
+
+```
+[ Browser at http://<docker-host>:3000 ]
+        │ webview / page load
+        ▼
+┌─────────────────────────────────────────────┐
+│  Docker network "hermes-net"                │
+│                                             │
+│  hermes-office  ◀──── ws ────  hermes-agent │
+│   (port 3000,                  (port 8642)  │
+│    port 18789)                              │
+└─────────────────────────────────────────────┘
+        ▲
+        │ port-forwards
+        ▼
+[ Docker host: 192.168.1.177 ]
+   ports 3000, 8642 published
+```
+
+The desktop app on a different machine can be configured (Settings → Connection → Remote)
+to use the Hermes Agent at `http://192.168.1.177:8642`. **Chat works** through that path.
+The Office tab will show the guidance screen instead of trying to start a local stack.
+For the 3D UI, the user opens `http://192.168.1.177:3000` in any browser.
+
+---
+
+## Prerequisites
+
+- Docker Engine 24+ and Docker Compose v2 installed on the host
+- The host's IP reachable from the desktop machine (e.g., `192.168.1.177`)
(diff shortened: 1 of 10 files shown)


fathah:
conflict

pmos69:
Thanks for this, and sorry this sat long enough for the branch to drift. I took another pass against current `main`, and I think the core UX issue is still real, but the implementation needs to be narrowed/updated before it can land.

The remaining problem I see is specifically **pure HTTP Remote mode**: Desktop is connected to an already-running Hermes API endpoint, but it does not have filesystem/process access to that host. In that mode the Office stack cannot be installed/started from the desktop app, and the Models tab can still end up showing/editing the local `models.json` instead of the connected remote agent’s models.

A few important things have changed since this PR was opened:

- Current Desktop now distinguishes pure Remote mode from SSH tunnel mode. `isRemoteMode()` now means `mode === "remote" || mode === "ssh"`, while `isRemoteOnlyMode()` means only `mode === "remote"`.
- SSH mode should **not** be blocked by this fix. SSH has remote command/filesystem access, and current Office code already has SSH-aware remote Office detection via `claw3dStatus().remoteUrl`.
- The current `Layout` already uses `isRemoteOnlyMode()` for several local-only screens, but Office and Models are still not covered by that gate.
- The Models IPC path currently handles SSH via `sshListModels(...)`, but pure Remote still falls back to local `listModels()`, which is misleading for a user connected to a remote API server.

What I think needs to be done for this PR, or for a superseding focused PR:

1. Rebase onto current `main` and resolve the conflicts.
2. Gate **Office** only when `isRemoteOnlyMode()` is true. In practice, that probably means reusing the `remoteMode` state already present in `src/renderer/src/screens/Layout/Layout.tsx`, not adding a new check based on `isRemoteMode()`.
3. Preserve SSH behavior. In SSH mode the Office screen should still be reachable so it can use the SSH/remote Office detection path rather than showing a pure-Remote Docker/setup warning.
4. Gate **Models** in pure Remote mode, or make Models truly remote-aware. Until pure Remote can query/update the connected agent’s model catalog through a remote API, showing the local model list is probably worse than a clear notice.
5. Keep the user-facing notice short and specific: something like “This screen needs local or SSH access to the Hermes host; pure HTTP Remote mode cannot manage these files/processes.” A link to external docs is fine, but the bundled Docker guide / `getDocUrl` plumbing in this branch feels broader than this fix needs and may be harder to keep accurate.
6. Add or update tests for the mode split:
   - Local mode: Office and Models remain available.
   - SSH mode: Office and Models remain available.
   - Pure Remote mode: Office and Models show the remote-only notice or otherwise avoid local-only behavior.
7. Include any new copy in the i18n files rather than hardcoding English in the component.

So: I think the idea is still relevant, but this exact branch is stale and currently too broad. Would you like to pick it back up and rework it along those lines? If not, no worries; we can prepare a smaller replacement PR that preserves the core idea here and credits/supersedes this one.

PR #44, opened 2026-05-01
fix: cancel pending timeouts on gateway toggle and guard Install state updates

Ricardo-M-L:
## Summary

**Bug 1 - Gateway.tsx race condition:**
Users clicking start/stop rapidly caused stale `setTimeout` callbacks to overwrite correct gateway state. Fix: track timeout refs and clear them on each toggle.

**Bug 2 - Install.tsx unmounted state update:**
Installation promise chain could call setState after component unmount, causing React warnings. Fix: add `isMounted` flag.

## Test plan
- [ ] Rapid gateway toggle shows correct state
- [ ] Navigating away during install produces no React warnings

### DIFF
--- a/src/renderer/src/screens/Gateway/Gateway.tsx
+++ b/src/renderer/src/screens/Gateway/Gateway.tsx
@@ -1,4 +1,4 @@
-import { useState, useEffect, useCallback } from "react";
+import { useState, useEffect, useCallback, useRef } from "react";
 import { GATEWAY_SECTIONS, GATEWAY_PLATFORMS } from "../../constants";
 import { useI18n } from "../../components/useI18n";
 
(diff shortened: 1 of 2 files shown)


fathah:
conflict

Ricardo-M-L:
Sorry for the noise — the previous head accidentally included the unrelated session-cache lock commit from #29. Force-pushed the branch to cherry-pick **only** the Gateway/Install changes onto current main (head: 49a86fd). Re-resolved one tiny conflict in \`Install.tsx\` (kept the existing i18n \`t("install.installationFailedHint")\` call rather than my old hardcoded English).

Now the diff is exactly the 2 React fixes:

- **Gateway.tsx**: track \`statusCheckTimeoutRef\` and \`clearTimeout\` it on every toggle, so a stale callback from a previous start/stop click can't overwrite the current gateway state.
- **Install.tsx**: \`isMounted\` guard so the install promise chain can't \`setState\` after unmount (squelches the React "Can't perform state update on unmounted component" warning).

Local \`pnpm test\` passes (259/259). Ready for re-review when you have a moment 🙏

PR #54, opened 2026-05-03
feat(renderer): wire Schedules tab to /api/jobs in remote mode

MillsAirCode:
## Summary

Today, Schedules is hard-disabled in remote-mode connections — `Layout.tsx` renders a `<RemoteNotice feature="Schedules" />` placeholder instead of the actual Schedules screen. The local implementation in `src/main/cronjobs.ts` reads `~/.hermes/cron/jobs.json` directly and shells out to `hermes cron …`, neither of which is meaningful against a remote backend.

Good news: the Hermes API server (`gateway/platforms/api_server.py`) **already exposes the full cron CRUD surface** — `/api/jobs` GET/POST, `/api/jobs/:id` GET/PATCH/DELETE, `/api/jobs/:id/{pause,resume,run}` POST. So this is a desktop-only PR; no matching Hermes runtime PR is needed to ship it.

## What changed

**`src/main/hermes.ts`** — export the existing `getApiUrl()` and `getRemoteAuthHeader()` helpers (previously only used internally for `sendMessageViaApi`). One source of truth for the remote base URL + Bearer header avoids re-deriving from `getConnectionConfig()` in each consumer.

**`src/main/cronjobs.ts`** —
- Extract a shared `normalizeJob()` helper. The same shape mapping is now used for both local jsonl rows and remote `/api/jobs` responses.
- Add tiny `remoteFetch()` and `remoteJsonError()` shims so each handler is one expressive call.
- Each of `listCronJobs / createCronJob / removeCronJob / pauseCronJob / resumeCronJob / triggerCronJob` now starts with an `if (isRemoteMode())` branch. Local-mode code paths are unchanged byte-for-byte after the `normalizeJob` extraction.

**`src/renderer/src/screens/Layout/Layout.tsx`** — drop the `<RemoteNotice feature="Schedules">` branch; render `<Schedules>` directly in both modes. The screen's renderer code already calls `window.hermesAPI.listCronJobs(...)` etc., which now do the right thing transparently.

## Notes

- The `profile` arg is ignored in remote mode — the Hermes server is already scoped per-profile by its own config and the desktop has no way to override it remotely. Same convention as `sendMessageViaApi`.
- If the server returns 501 (`{"error": "Cron module not available"}`), `listCronJobs` logs and returns `[]`. Schedules renders cleanly as "no jobs" rather than a hard error — graceful degradation if a user is on a Hermes runtime build with cron disabled.
- A small `remoteJobAction()` helper handles pause/resume/run/delete to avoid 4× the same try/catch boilerplate.

## Rebased on current main (2026-05-03)

The conn-scope follow-up (PR #53) is no longer needed — fathah landed the same fix manually on main in 4926e16. After rebasing this branch onto current `main`, git automatically dropped the cherry-picked commit since the patch is already upstream. This PR now contains only the Schedules feature commit, cleanly applies on top of v0.3.2.

## Test plan

- [x] `npx tsc --noEmit -p tsconfig.web.json` exits 0
- [x] `npx tsc --noEmit -p tsconfig.node.json` shows only the pre-existing i18next TS2742 (unrelated)
- [x] `npx eslint src/main/cronjobs.ts src/main/hermes.ts src/renderer/src/App.tsx src/renderer/src/screens/Layout/Layout.tsx` clean (one pre-existing setState-in-effect warning at App.tsx:86 is baseline)
- [x] `npx vitest run` shows the same 2 pre-existing failures as upstream/main, no new failures
- [x] **Manual end-to-end against a live Hermes API server (api_server platform, port 18642).** All 6 handlers (list/create/pause/resume/trigger/delete) round-trip cleanly; error paths (missing job, bad auth) handled gracefully. Full output and harness details in the comment below.

### DIFF
--- a/src/main/cronjobs.ts
+++ b/src/main/cronjobs.ts
@@ -4,6 +4,7 @@ import { join } from "path";
 import { execFile } from "child_process";
 import { HERMES_HOME, HERMES_PYTHON, HERMES_SCRIPT } from "./installer";
 import { profileHome } from "./utils";
+import { isRemoteMode, getApiUrl, getRemoteAuthHeader } from "./hermes";
 
 export interface CronJob {
   id: string;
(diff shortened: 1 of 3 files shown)


MillsAirCode:
## Manual e2e test — passed against a live Hermes API server

I spun up an isolated Hermes runtime (`HERMES_HOME=/tmp/hermes-test`, `api_server` platform enabled on port 18642 with `key: e2e-test-key`), then invoked each of the 6 changed handlers in `cronjobs.ts` directly through `tsx` with stub modules supplying remote-mode connection config.

### Happy path — all 6 handlers ✓

```
step 1: list (expect empty)
  -> 0 jobs
step 2: create
  -> { success: true }
step 3: list (expect 1)
  -> 1 jobs: [ '97620494811f e2e-test-job state=active' ]
step 4: pause 97620494811f
  -> { success: true }
  -> state after pause: paused enabled: false
step 5: resume 97620494811f
  -> { success: true }
  -> state after resume: active enabled: true
step 6: trigger 97620494811f
  -> { success: true }
step 7: delete 97620494811f
  -> { success: true }
step 8: list (expect empty again)
  -> 0 jobs

ALL 6 HANDLERS PASSED ✓
```

### Error paths

**Missing job ID** — the server returns 400 with a structured error:
```
pause non-existent job:
  { success: false, error: 'Invalid job ID format' }
delete non-existent job:
  { success: false, error: 'Invalid job ID format' }
```

The `remoteJsonError()` helper unwraps the `{ "error": "..." }` body cleanly. Renderer gets a useful error string back.

**Bad auth** — `401 {"error": {"message": "Invalid API key", ...}}` — `Bearer` header from `getRemoteAuthHeader()` flows through correctly; flipping the key surfaces the expected failure.

### Schema match

`/api/jobs` response shape (`{"jobs": [job_dicts]}` with `id`, `name`, `schedule`, `prompt`, `enabled`, `state`, `next_run_at`, `last_run_at`, `last_status`, `last_error`, `repeat`, `deliver`, `skills`, `script`) maps cleanly onto the existing `CronJob` interface via the extracted `normalizeJob()` helper. No renderer-side changes needed.

This is direct invocation of the actual `cronjobs.ts` code in this PR, not a re-implementation. The harness mocks only the `./hermes` module (to force `isRemoteMode()` true and supply the test URL/key) — the fetch logic, normalization, error handling, and JSON parsing are all the merged code paths.

PR #66, opened 2026-05-04
fix: Security: Harden Electron sandbox, IPC communication and file operations

joathan:
Description:
  This PR introduces several critical security improvements based on an audit of the application's Electron configuration and IPC handling. The goal is to eliminate
  potential RCE (Remote Code Execution) vectors via XSS and prevent unauthorized file system access.

  Critical Fixes:
   - IPC Hardening (askpass.ts): Replaced nodeIntegration: true and contextIsolation: false with a secured sandbox: true and contextIsolation: true environment for the
     password dialog.
   - Secure Bridge: Implemented a dedicated preload script (askpass-preload.ts) using contextBridge. The IPC channel is now passed securely via URL hash fragments instead
     of being injected into the HTML, preventing XSS-to-RCE escalation.
   - Main Window Hardening: Enabled sandbox: true and disabled webviewTag on the main BrowserWindow to minimize the attack surface.

  High-Impact Improvements:
   - URL Validation (safeOpenExternal): Added a protocol allowlist (http:, https:, mailto:) for all shell.openExternal calls, blocking dangerous protocols like file:// or
     custom handlers.
   - Shell Injection Prevention: Replaced execSync with execFileSync in installer.ts to prevent command injection through shell metacharacters in path variables.
   - Path Traversal Protection: Added validation in src/main/skills.ts to ensure skillPath stays within the designated skills directory, preventing arbitrary file reads via
     IPC.

  Reliability & Data Integrity:
   - Atomic File Writes: Updated safeWriteFile in utils.ts to use a write-to-temp + rename pattern. This ensures file integrity and prevents data corruption during
     concurrent writes or system crashes.
   - Preload Compatibility: Refactored the main preload script to be compatible with Electron's sandbox mode by removing dependencies on Node.js-specific modules (like
     @electron-toolkit/preload).

**These changes ensure the application follows the latest Electron Security Best Practices while maintaining full functionality.**


fathah:
Thanks for the thorough security pass — the rationale is well-documented and the changes are well-scoped. A few items before this can land:

## Blocker — regression in the preload stub

Removing `@electron-toolkit/preload` is the right call for sandbox compatibility, but the replacement stub in `src/preload/index.ts` only exposes `versions` and `type`:

```ts
window.electron = { process: { versions: process.versions, type: process.type } }
```

The renderer reads `platform` here:

```ts
// src/renderer/src/App.tsx:21
const isMac = window.electron?.process?.platform === "darwin";
```

After this PR, `platform` is `undefined`, so `isMac` is always `false` on macOS — any macOS-specific behavior gated on it silently breaks. Please add `platform: process.platform` to the stub in both the `contextIsolated` and fallback branches.

## Nits (non-blocking)

- **`setWindowOpenHandler` is redundant.** Both branches of the `if` return `{ action: "deny" }`; the `if` can be collapsed:
  ```ts
  mainWindow.webContents.setWindowOpenHandler((details) => {
    safeOpenExternal(details.url);
    return { action: "deny" };
  });
  ```
- **`skills.ts` traversal check.** `resolve(rel) === rel` works but is non-obvious; `path.isAbsolute(rel)` reads more directly and is what most reviewers will expect.
- **`package-lock.json` version bump.** The lockfile flips `0.2.2 → 0.3.2`. Worth confirming that's an artifact of a `main` rebase and not introduced by this branch.
- **CSP `frame-src http://localhost:*`.** Loosens to any localhost port — fine if intentional for the Office webview, but worth a comment in the HTML so a future reader understands why.

## Things I checked and like

- Askpass channel via URL hash + `contextBridge` is a clean fix for the `require("electron")` injection.
- `will-attach-webview` forcing `sandbox`/`contextIsolation` and stripping `preload`/`preloadURL` is the right belt-and-suspenders.
- `execSync` → `execFileSync` in `installer.ts` and the `statSync` check on the import path close real shell-injection vectors.
- Atomic write via tmp + rename in `safeWriteFile` is a nice reliability bonus alongside the security work.

Once the `platform` stub is restored I'm happy to approve.

joathan:
@fathah  done in [5b97d5f](https://github.com/fathah/hermes-desktop/pull/66/commits/5b97d5f834470dbb5af457d88f811c62bc205092)

pmos69:
Thank you for the security pass here. A lot of the core hardening from this branch has since landed through #92, including the Electron navigation/window policy, sandbox-compatible preload work, askpass isolation, and the xecSync -> xecFileSync doctor path. Since this branch is now conflicted and mostly superseded by that merged work, I’m going to close this PR to keep the queue clear.\n\nThere are still a couple of useful small ideas from this review that are worth carrying forward separately, especially validating skill-content IPC paths carefully across installed, profile-specific, and bundled skills. I’ll follow up with a focused replacement PR for those remaining pieces rather than trying to merge this stale branch. Thanks again for surfacing the original hardening work.

PR #68, opened 2026-05-05
Add SSH tunnel connection mode

ComicBit:
## Overview

This PR adds a third connection mode to hermes-desktop: **SSH Tunnel**. Instead of requiring a publicly exposed HTTP server or API key, the app tunnels all traffic through SSH to the remote host — no open ports, no credentials to manage.

### SSH Tunnel mode
- A distinct **SSH Tunnel** option appears on both the Welcome (onboarding) screen and the Settings screen
- All fields start blank (no prefill) so it works out of the box for any user
- The app establishes a local SSH port-forward on connect, then routes all API traffic through `localhost`
- On startup, if SSH mode is saved, the tunnel is re-established automatically

### Full feature parity with local mode
All screens that work locally now work over SSH:

| Feature | How it works |
|---|---|
| **Chat** | API calls go through SSH tunnel; Bearer token is read from remote `~/.hermes/.env` automatically |
| **Sessions** | Queries remote `state.db` via SSH + Python |
| **Memory / Persona** | Reads/writes `MEMORY.md`, `USER.md`, `SOUL.md` on remote via SSH |
| **Skills** | Lists, installs, uninstalls skills on remote via SSH |
| **Tools / Toolsets** | Reads/writes remote `config.yaml` |
| **Providers / Env** | Reads/writes remote `~/.hermes/.env` |
| **Gateway** | Start/stop gateway on remote; platform status from live `gateway_state.json` |
| **Profiles** | Lists, creates, deletes profiles on remote |
| **Logs** | Tails remote log files via SSH |
| **Versions** | Runs `hermes --version` on remote |
| **Doctor** | Runs `hermes doctor` on remote |
| **Models** | Reads remote `models.json` |

### Files added
- `src/main/ssh-tunnel.ts` — SSH tunnel lifecycle (start, stop, test, status)
- `src/main/ssh-remote.ts` — All SSH-proxied feature implementations

### Files modified
- `src/main/config.ts` — `SshConnectionConfig` interface; mode extended to `"local" | "remote" | "ssh"`
- `src/main/hermes.ts` — `isRemoteOnlyMode()`, SSH auth header with cached remote API key
- `src/main/index.ts` — All IPC handlers updated to proxy through SSH when in SSH mode
- `src/preload/index.ts` + `index.d.ts` — New IPC methods exposed to renderer
- `src/renderer/src/screens/Welcome/Welcome.tsx` — SSH as third panel with blank fields
- `src/renderer/src/screens/Settings/Settings.tsx` — SSH button + form
- `src/renderer/src/screens/Layout/Layout.tsx` — Uses `isRemoteOnlyMode()` so SSH shows all screens
- `src/renderer/src/App.tsx` — Handles SSH mode on startup

## Implementation notes

- Uses the system `ssh` binary — no Node.js SSH library dependency
- `sshExec()` runs shell commands; `sshPython()` pipes Python scripts via stdin — avoids shell escaping issues for complex data operations
- `~` does not expand inside double-quoted bash strings; all paths use `$HOME/` instead
- Gateway platform state is read from `gateway_state.json` (live) rather than `config.yaml` (static)
- `HOMEASSISTANT_URL`/`HOMEASSISTANT_TOKEN` are aliased to `HA_URL`/`HA_TOKEN` so credentials display correctly regardless of which env var name the server uses

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "hermes-desktop",
-  "version": "0.2.2",
+  "version": "0.3.4",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "hermes-desktop",
-      "version": "0.2.2",
+      "version": "0.3.4",
       "hasInstallScript": true,
       "dependencies": {
         "@electron-toolkit/preload": "^3.0.2",
@@ -109,7 +109,6 @@
       "integrity": "sha512-CGOfOJqWjg2qW/Mb6zNsDm+u5vFQ8DxXfbM09z69p5Z6+mE1ikP2jUXw+j42Pf1XTYED2Rni5f95npYeuwMDQA==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.29.0",
         "@babel/generator": "^7.29.0",
@@ -474,7 +473,6 @@
         }
       ],
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=18"
       },
@@ -498,7 +496,6 @@
         }
       ],
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=18"
       }
@@ -974,6 +971,7 @@
       "dev": true,
       "license": "BSD-2-Clause",
       "optional": true,
+      "peer": true,
       "dependencies": {
         "cross-dirname": "^0.1.0",
         "debug": "^4.3.4",
@@ -995,6 +993,7 @@
       "dev": true,
       "license": "MIT",
       "optional": true,
+      "peer": true,
       "dependencies": {
         "graceful-fs": "^4.2.0",
         "jsonfile": "^6.0.1",
@@ -1011,6 +1010,7 @@
       "dev": true,
       "license": "MIT",
       "optional": true,
+      "peer": true,
       "dependencies": {
         "universalify": "^2.0.0"
       },
@@ -1025,6 +1025,7 @@
       "dev": true,
       "license": "MIT",
       "optional": true,
+      "peer": true,
       "engines": {
         "node": ">= 10.0.0"
       }
@@ -2801,7 +2802,8 @@
       "resolved": "https://registry.npmjs.org/@types/aria-query/-/aria-query-5.0.4.tgz",
       "integrity": "sha512-rfT93uj5s0PRL7EzccGMs3brplhcrghnDoV26NqKhCAS1hVo+WdNsPvE/yb6ilfr5hi2MEk6d5EWJTKdxg8jVw==",
       "dev": true,
-      "license": "MIT"
+      "license": "MIT",
+      "peer": true
     },
     "node_modules/@types/babel__core": {
       "version": "7.20.5",
@@ -2973,7 +2975,6 @@
       "resolved": "https://registry.npmjs.org/@types/node/-/node-22.19.15.tgz",
       "integrity": "sha512-F0R/h2+dsy5wJAUe3tAU6oqa2qbWY5TpNfL/RGmo1y38hiyO1w3x2jPtt76wmuaJI4DQnOBu21cNXQ2STIUUWg==",
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "undici-types": "~6.21.0"
       }
@@ -3001,7 +3002,6 @@
       "resolved": "https://registry.npmjs.org/@types/react/-/react-19.2.14.tgz",
       "integrity": "sha512-ilcTH/UniCkMdtexkoCN0bI7pMcJDvmQFPvuPvmEaYA/NSfFTAgdUSLAoVjaRJm7+6PvcM+q1zYOwS4wTYMF9w==",
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "csstype": "^3.2.2"
       }
@@ -3012,7 +3012,6 @@
       "integrity": "sha512-jp2L/eY6fn+KgVVQAOqYItbF0VY/YApe5Mz2F0aykSO8gx31bYCZyvSeYxCHKvzHG5eZjc+zyaS5BrBWya2+kQ==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "peerDependencies": {
         "@types/react": "^19.2.0"
       }
@@ -3104,7 +3103,6 @@
       "integrity": "sha512-rLoGZIf9afaRBYsPUMtvkDWykwXwUPL60HebR4JgTI8mxfFe2cQTu3AGitANp4b9B2QlVru6WzjgB2IzJKiCSA==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@typescript-eslint/scope-manager": "8.58.0",
         "@typescript-eslint/types": "8.58.0",
@@ -3489,7 +3487,6 @@
       "integrity": "sha512-UVJyE9MttOsBQIDKw1skb9nAwQuR5wuGD3+82K6JgJlm/Y+KI92oNsMNGZCYdDsVtRHSak0pcV5Dno5+4jh9sw==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "bin": {
         "acorn": "bin/acorn"
       },
@@ -3523,7 +3520,6 @@
       "integrity": "sha512-IWrosm/yrn43eiKqkfkHis7QioDleaXQHdDVPKg0FSwwd/DuvyX79TZnFOnYpB7dcsFAMmtFztZuXPDvSePkFw==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "fast-deep-equal": "^3.1.1",
         "fast-json-stable-stringify": "^2.0.0",
@@ -4118,7 +4114,6 @@
         }
       ],
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "baseline-browser-mapping": "^2.10.12",
         "caniuse-lite": "^1.0.30001782",
@@ -4746,7 +4741,8 @@
       "integrity": "sha512-+R08/oI0nl3vfPcqftZRpytksBXDzOUveBq/NBVx0sUp1axwzPQrKinNx5yd5sxPu8j1wIy8AfnVQ+5eFdha6Q==",
       "dev": true,
       "license": "MIT",
-      "optional": true
+      "optional": true,
+      "peer": true
     },
     "node_modules/cross-spawn": {
       "version": "7.0.6",
@@ -5115,7 +5111,6 @@
       "integrity": "sha512-glMJgnTreo8CFINujtAhCgN96QAqApDMZ8Vl1r8f0QT8QprvC1UCltV4CcWj20YoIyLZx6IUskaJZ0NV8fokcg==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "app-builder-lib": "26.8.1",
         "builder-util": "26.8.1",
@@ -5210,7 +5205,8 @@
       "resolved": "https://registry.npmjs.org/dom-accessibility-api/-/dom-accessibility-api-0.5.16.tgz",
       "integrity": "sha512-X7BJ2yElsnOJ30pZF4uIIDfBEVgF4XEBxL9Bxhy6dnrm5hkzqmsWHGTiHqRiITNhMyFLyAiWndIJP7Z1NTteDg==",
       "dev": true,
-      "license": "MIT"
+      "license": "MIT",
+      "peer": true
     },
     "node_modules/dotenv": {
       "version": "16.6.1",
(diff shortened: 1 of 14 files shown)


fathah:
This is a genuinely useful addition — SSH mode closes a real gap (no need to expose a port + manage an API key). Architecture is clean: `ssh-tunnel.ts` for lifecycle, `ssh-remote.ts` for per-feature shims. A few things I'd want addressed before merge.

## Blocking

### 1. Every IPC handler now blocks the main process

`sshExec` and `sshPython` use `spawnSync`, and with `BatchMode=yes` + no `ControlMaster`, every call opens a fresh SSH session (TCP connect + key exchange + auth) — easily 1–3s per call. The Electron main process is single-threaded, so each IPC call freezes the UI for the full roundtrip. With chat, sessions, memory, env, skills, etc. all going through this, the app is going to feel broken.

Fixes (either, ideally both):

- Switch to async `spawn` and return a Promise from `sshExec`/`sshPython`.
- Add SSH multiplexing in `buildExecArgs` / `buildSshArgs`:
  ```
  -o ControlMaster=auto
  -o ControlPath=~/.ssh/cm-hermes-%r@%h:%p
  -o ControlPersist=60s
  ```
  After the first connect, subsequent calls reuse the master socket — orders of magnitude faster.

### 2. Shell interpolation into remote commands

`replace(/"/g, '\\"')` doesn't make a string safe inside double-quoted shell — backticks, `$(...)`, and `\` are still active. Affected sites:

- `sshReadFile`: `cat "${p}"` — `p` includes paths returned from the remote (e.g. skill `path` from `sshListInstalledSkills` is fed back into `sshGetSkillContent`).
- `sshWriteFile`: `mkdir -p "${dir}" && cat > "${p}"`.
- Logs handler: `` tail -n ${lines} "${remotePath}" `` — `lines` is unquoted entirely; a non-numeric value is straight injection.
- Skills install/uninstall, profile create/delete — `"${safe}"` with the broken escape.

Threat model is mostly user-attacking-their-own-remote, so the impact is limited, but the pattern fails review and there are flows (remote-supplied skill paths, the `lines` argument) where the input isn't strictly user-controlled.

Recommend: pass dynamic values via stdin or as separate argv elements wherever possible, or use a real shell-quoter (`printf %q` on the remote, or build the command as `bash -c '…' -- arg1 arg2` and reference `"$1"`/`"$2"` inside).

Also: `tail -n ${lines}` should at minimum be coerced with `Number(lines) | 0` before interpolation.

### 3. Python script string interpolation

`sshGetSessionStats` builds the Python script with `db = os.path.expanduser("${dbPath}")`, where `dbPath` derives from `profile`. The `[a-zA-Z0-9_-]` filter that's used in `sshCreateProfile`/`sshDeleteProfile` isn't applied here — if `profile` reaches this without the same filter, it's a Python code-injection.

Same for `sshGetSessionMessages` — `sessionId.replace(/'/g, "")` only strips single quotes; a trailing backslash escapes out of the surrounding quote.

Cleaner pattern: never interpolate values into the Python source. Pass them as JSON on stdin and read them via `json.load(sys.stdin)`, then use them as variables (and as parameters in `conn.execute(sql, (id,))` for SQL).

## Non-blocking but worth a follow-up

- **Stderr in user-facing errors.** `throw new Error(result.stderr?.trim())` surfaces SSH banners, key paths, hostname mismatches into UI toasts. Sanitize before display.
- **Tunnel state tracking.** `tunnelProcess.killed` flips to true immediately on `SIGTERM` even if the child hasn't exited; `isSshTunnelActive()` can lie. Track an explicit `running` boolean toggled in the `exit` handler instead.
- **`getSshTunnelUrl()`** returns `http://127.0.0.1:18642` even before any tunnel is started — callers may use a stale URL. Return `null` or throw when `activeConfig` is null.
- **Tunnel cleanup on app quit.** Verify `app.on("before-quit")` (or equivalent) calls `stopSshTunnel()` so we don't leak `ssh` processes on crash/quit.
- **`spawnSync` 30s timeout** is fine for most ops but might be too short for skill installs that pull from network — consider a per-call override.
- **`StrictHostKeyChecking=accept-new` + `BatchMode=yes`.** Acceptable defaults, but if the user has no SSH key configured the failure mode should produce a clear "configure an SSH key first" error rather than a generic "SSH command failed".
- **Conflict with #66.** Both PRs touch `src/preload/index.ts`, `src/main/index.ts`, `src/renderer/src/App.tsx`. Whichever lands second needs a rebase. #66's `sandbox: true` shouldn't break this PR's renderer-side IPC, but worth re-verifying after merge order is decided.

## Things I checked and like

- `ssh-tunnel.ts` lifecycle is solid: free-port discovery, `ExitOnForwardFailure=yes`, server alive keepalives, a dedicated `testSshConnection` that polls then hits `/health`.
- Argv-mode `spawn("ssh", [...])` for the SSH invocation itself — no local shell injection.
- `~/` → `$HOME/` substitution is the right call (and the PR description acknowledges the gotcha).
- Profile-name allowlist on create/delete.
- Reading live `gateway_state.json` for platform status instead of static `config.yaml` — better correctness.
- Welcome flow with blank fields rather than auto-prefilling something user-specific.

Happy to re-review once #1 (async + multiplexing) and the interpolation cleanup in #2/#3 are in.

ComicBit:
Addressed the SSH review feedback in local commit `a6f0f25` (`Address SSH mode review feedback`).

What changed:
- Replaced synchronous SSH helpers with async `spawn` so SSH-backed IPC no longer blocks the Electron main process.
- Added SSH multiplexing to command execution and tunnel setup with `ControlMaster=auto`, `ControlPath`, and `ControlPersist=60s`.
- Cleaned up remote shell interpolation for file reads/writes, logs, skills, and profiles using quoted args/stdin patterns; log line counts are coerced and bounded.
- Moved session/profile-derived Python values to JSON stdin and SQL parameters instead of interpolating into Python source.
- Sanitized SSH errors before exposing them to the UI.
- Hardened tunnel lifecycle: explicit running state, null/stale URL protection, health checks against `/health`, reconnect on lost gateway/tunnel health, and gateway restart after remote update.
- Fixed SSH mode routing for logs, memory, models, gateway platform state, debug dump, update, model config, memory providers, and installed skills.
- Updated installed-skill discovery to match the remote server’s actual installed skill layout, including both direct and category-based skill directories.
- Guarded session rendering against null message content.

Verification:
- `npm run typecheck` passed.
- `npm test` passed on rerun: 259 tests.
- `npm run build` passed.
- Changed-file ESLint with `--quiet` passed.
- Full `npm run lint` still reports pre-existing unrelated repo lint issues in `.agents`, tests, and locale files.
- Manual SSH-mode smoke test was run by Leandro and reported working.

Mac artifacts were also produced locally in `dist/`:
- `hermes-desktop-0.3.4.dmg`
- `Hermes Agent-0.3.4-arm64-mac.zip`

ComicBit:
Rebased the SSH branch onto current upstream/main and resolved the conflict in src/main/index.ts by preserving both the upstream locale import and the SSH remote IPC helpers. I also mirrored the upstream skills uninstall CLI fix in the SSH remote path by removing the unsupported --yes flag from remote skill uninstall.\n\nVerification after the rebase:\n- npm run typecheck\n- npm test: 259 passed\n- npm run build\n- npx eslint on the SSH/conflict-touched files with --quiet\n\nGitHub now reports mergeStateStatus=CLEAN for this PR at head c7a3008.

ComicBit:
@fathah all done. Let me know if it's good. but for some reason @MillsAirCode opened a new PR to solve those issues. There were also other issues I've solved on here, the ssh was unreliable tbh on many parts of the app. Now seems stable.

MillsAirCode:
Hey @ComicBit, sorry about #76 — that was the wrong move on my part. Here are the fixes for fathah's 3 blocking review items if you want to apply them to your branch:

**1. Async SSH (spawnSync → spawn+Promise)**
- Replace `import { spawnSync }` with `import { spawn, ChildProcess }`
- New `collectOutput(proc)` helper that promisifies stdout/stderr/status from a ChildProcess
- `sshExec()` and `sshPython()` become async, using `spawn()` + `collectOutput()`
- Add ControlMaster multiplexing to `buildExecArgs()`: `-o ControlMaster=auto -o ControlPath=/tmp/hermes_ssh_<hash>.sock -o ControlPersist=300`
- Every exported function becomes async (returns `Promise<T>`)
- IPC handlers in `index.ts` that call these without returning need `await` added (set-env, set-config, start-gateway, stop-gateway, set-platform-enabled, and the `sshReadRemoteApiKey` call in start-ssh-tunnel)

**2. Shell injection**
- Replace the `replace(/"/g, '\\"')` pattern with base64 encoding:
  - `shellSafe(value)` encodes via `Buffer.from(value).toString("base64")`
  - Remote side decodes: `cat "$(echo 'BASE64' | base64 -d)"`
- Affected: `sshReadFile`, `sshWriteFile`, `sshInstallSkill`, `sshUninstallSkill`, `sshSearchSkills`

**3. Python injection**
- `sshPython()` accepts optional `paramsJson`, base64-encodes it, prepends a decoder line to the script
- Python scripts read `_P["dbPath"]` etc instead of interpolated values
- All SQL uses parameterized `?` placeholders
- Affected: `sshGetSessionStats`, `sshListSessions`, `sshGetSessionMessages`, `sshSearchSessions`

Full diff is on the `ssh-tunnel-fix` branch at MillsAirCode/hermes-desktop if you want to cherry-pick.

ComicBit:
I've already done the fixes. @MillsAirCode Bro take your ai model on the leash

fathah (changes requested):
Thanks for pushing through the review feedback — the lifecycle changes look great. A quick re-review against `c7a3008`:

## Nicely addressed ✓

- `isSshTunnelHealthy()` + health-check-then-restart on chat send and in `ensureSshTunnel()`.
- `getSshTunnelUrl()` returns `null` when inactive, with `getApiUrl()` failing loudly instead of using a stale URL.
- Async/await consistency across the IPC handlers (`set-env`, `set-config`, `set-platform-enabled`, `start-gateway`, `stop-gateway`, …).
- 120 s HTTP timeout in `sendMessageViaApi()` with a clear user-facing error.
- Provider/model change now restarts the gateway so the running process picks up the new config.
- `ControlMaster=auto` + `ControlPersist=60s` for SSH connection reuse — big perf win for multi-call screens.

## Still blocking ❌

### 1. YAML injection in `sshSetConfigValue` (must-fix)

`src/main/ssh-remote.ts:592`:

```ts
const updated = regex.test(content)
  ? content.replace(regex, `$1"${value}"`)
  : content;
```

`value` is interpolated raw. It reaches the UI through `sshSetModelConfig` → `set-model-config` IPC → the Settings screen, so any `"`, `\`, or newline in `provider`, `model`, or `base_url` corrupts `config.yaml` on the remote (e.g. `baseUrl = "x\napi_key: stolen"` injects an arbitrary YAML key). Self-inflicted only — not a privilege escalation — but it silently breaks the remote config, which is hard to debug.

Cheapest fix: reject or escape on write.

```ts
if (/["\\\n]/.test(value)) {
  throw new Error("config value contains illegal characters (\", \\, newline)");
}
const escaped = value.replace(/\\/g, "\\\\").replace(/"/g, '\\"');
const updated = regex.test(content)
  ? content.replace(regex, `$1"${escaped}"`)
  : content;
```

Better long-term: parse + serialize via a YAML lib (`yaml` is already in the dep tree if I'm reading right) so the regex-based mutation goes away entirely.

### 2. `StrictHostKeyChecking=accept-new` (please document or surface)

`src/main/ssh-tunnel.ts:111` and `src/main/ssh-remote.ts:27`. Trust-on-first-use is a defensible default for a developer-facing tool, but right now it's silent — the user has no way to verify the fingerprint of the host they just trusted, and no signal if the host key changes (subsequent connects fail closed, but the user only sees a generic SSH error).

Two cheap improvements, either is enough:

- Surface the host key fingerprint on first connect in the Settings UI ("Connected to `host` with key `SHA256:…` — verify and confirm"), or
- Add a short note in `README` / a tooltip on the SSH form explaining that the first connection trusts the host and stores its key in `~/.ssh/known_hosts`.

## Nice-to-have (non-blocking)

- **Master socket cleanup**: `stopSshTunnel()` kills the tunnel process but leaves the `ControlMaster` socket alive for `ControlPersist=60s`. An `ssh -O exit` on the same `ControlPath` during shutdown would clean it up immediately.
- **`bash -c '...' -- "$@"` pattern**: nicely done — shell-quoted, positional-arg-passed. Worth a short comment on top of the helper noting *why* (that `$1`/`$2` inside `bash -c` are not re-expanded), so it isn't accidentally "simplified" later.
- **Catch-blocks returning `[]`**: a few `try { … } catch { return []; }` cases swallow real errors (network blips look identical to "directory empty"). At minimum, `console.error` so they show up in main-process logs.
- **`ssh-remote.ts` size (1140 lines)**: every IPC handler in `index.ts` is now `if (mode === "ssh") return sshFoo(); return foo();`. Probably worth a follow-up issue to introduce a `Backend` interface (`LocalBackend`, `SshBackend`) so future features only need one implementation point — not blocking this PR.

Happy to merge once #1 is addressed and there's a short note for #2. Thanks again for the thorough work — SSH mode is going to be a really useful addition.

ComicBit:
@fathah happy to help!

Addressed the latest review follow-up in faa852c.

Changes
Hardened sshSetConfigValue so SSH model/config writes reject YAML-breaking values containing quotes, backslashes, or newlines before touching the remote config file.
Added a regression test covering the rejected values.
Added an SSH settings note explaining StrictHostKeyChecking=accept-new trust-on-first-use behavior, including that the host key is stored in ~/.ssh/known_hosts and future key changes fail closed.
Verification
npx vitest run tests/ssh-remote.test.ts ✅ 4 passed
npm run typecheck ✅
npm test ✅ 263 passed
npm run build ✅
npx eslint src/main/ssh-remote.ts tests/ssh-remote.test.ts --quiet ✅

Note: linting the full Settings.tsx file still hits the pre-existing react-hooks/set-state-in-effect warning at line 159, unrelated to this patch. Typecheck and build pass.

PR #71, opened 2026-05-05
feat(desktop): wire Sessions tab for remote mode

IvanZaro:
### Summary

Wire the Sessions tab to work in remote mode by adding an HTTP client that fetches from the backend API, modifying IPC handlers to branch on `isRemoteMode()`, and removing the `RemoteNotice` gate for Sessions.

### Changes

1. **`src/main/sessions-remote.ts`** (NEW) — HTTP client that mirrors `sessions.ts` types but uses `fetch()` against the remote API server
2. **`src/main/index.ts`** — IPC handlers (`list-sessions`, `get-session-messages`, `search-sessions`) now branch on `isRemoteMode()`
3. **`Layout.tsx`** — Removed `<RemoteNotice feature="Sessions" />` so Sessions renders in both modes

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -71,6 +71,11 @@ import {
 } from "./config";
 import { listSessions, getSessionMessages, searchSessions } from "./sessions";
 import {
+  listSessionsRemote,
+  getSessionMessagesRemote,
+  searchSessionsRemote,
+} from "./sessions-remote";
+import {
   syncSessionCache,
   listCachedSessions,
   updateSessionTitle,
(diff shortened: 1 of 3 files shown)


pmos69:
Thanks for tackling this. The underlying gap was real: Sessions should not be blocked forever in pure Remote mode, and this PR pointed at the right direction for that earlier codebase.

I’m going to close this as superseded rather than merge it, because current `main` has moved on enough that this branch is now stale/conflicting and incomplete against the current Sessions flow.

A few concrete reasons:

- The PR currently conflicts with `main`.
- Current `Sessions.tsx` primarily uses `syncSessionCache()` / `listCachedSessions()` for the main list, plus `searchSessions()` and delete handling. This PR only wires `list-sessions`, `get-session-messages`, and `search-sessions`, so simply removing the `RemoteNotice` would not fully enable the current Sessions screen in pure Remote mode.
- Current `main` already has SSH-mode session support through the SSH remote helpers, so the remaining gap is specifically pure HTTP Remote mode.
- #107 is the more complete successor approach: it introduces a shared remote API adapter and routes Sessions/cache/search plus other management views through Hermes Agent API endpoints.

So the idea here is still valid, but the implementation has been superseded by the broader #107 direction and would need to be rebuilt/rebased around current `main` rather than merged as-is.

Thanks again for catching the Remote-mode Sessions gap early.

PR #76, opened 2026-05-06
SSH tunnel: async, injection fixes, ControlMaster

MillsAirCode:
Addresses all 3 blocking items from review of #68:

1. **Async SSH** — all 35+ exported functions in ssh-remote.ts converted from spawnSync to async spawn+Promise. ControlMaster multiplexing added (ControlPersist=300) on both exec and tunnel paths to amortize connection setup.

2. **Shell injection** — replaced double-quote escaping with base64-encode pattern. User-controlled values (paths, identifiers, queries) are base64-encoded before shell interpolation. base64 output is [A-Za-z0-9+/=] so no metacharacters can escape.

3. **Python injection** — params delivered as base64-encoded JSON, decoded in Python via `json.loads(base64.b64decode(...))`. All SQL queries use parameterized `?` placeholders instead of string interpolation.

Also: IPC handlers in index.ts updated to `await` the now-async calls.

Both `typecheck:node` and `typecheck:web` pass clean.

Supersedes #68 (ComicBit's original SSH tunnel work is the base — this commit layers the review fixes on top).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "hermes-desktop",
-  "version": "0.2.2",
+  "version": "0.3.4",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "hermes-desktop",
-      "version": "0.2.2",
+      "version": "0.3.4",
       "hasInstallScript": true,
       "dependencies": {
         "@electron-toolkit/preload": "^3.0.2",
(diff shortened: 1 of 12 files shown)


leopiccione:
bro why would you work an open pr on another PR while I was working on it?

MillsAirCode:
Closing this — should have contributed the fixes to #68 directly instead of opening a competing PR. Sorry about that. Will post the changes as a review comment on #68 so you can pick them up.

PR #92, opened 2026-05-10
Security hardening: Electron boundaries, IPC validation, and Windows install support

ASTairov:
## Summary
- Restore a reproducible verification baseline for security-sensitive desktop behavior.
- Harden Electron navigation/window boundaries and keep renderer isolation intact.
- Isolate the askpass password dialog behind a minimal preload bridge.
- Validate profile names at IPC path boundaries before filesystem access.
- Keep sandboxed preload code self-contained.
- Fix Windows GUI installation by using the official PowerShell installer instead of invoking install.sh through bash/WSL.

## Verification
- npm run typecheck
- npm test
- npm run build
- GUI smoke-tested on Windows
- Windows GUI install tested locally after the PowerShell installer fix


fathah:
Plz resolve the conflicts

ASTairov:
Resolved the conflicts with current main.

I merged upstream/main into this branch, preserved the stricter profile IPC validation logic, reconciled the Windows installer/constants changes, and regenerated package-lock.json from the merged package.json.

Verification:
- npm test: 297 tests passed
- npm run build: passed

fathah:
@ASTairov 

Thanks for the thorough hardening pass — verified locally on the branch:

- `npm run typecheck` ✅
- `npm test` ✅ (297/297, 17 files)
- `npm run build` ✅

**One regression to fix before this can land**

The merge commit (`410f32a0`) silently drops `src/main/sudoCreds.ts` and its call site in `installer.ts`. That file was added to `main` in `8119b64e Sudo Setup` to precache sudo credentials upfront on Linux — specifically to keep `install.sh` from deadlocking when Playwright's `--with-deps` shells out to `apt-get` with no TTY. The merge message lists "SSH setup, i18n updates, Windows installer" but not sudo, so this looks like an inadvertent miss during conflict resolution rather than an intentional revert.

The remaining `setupAskpass` bridge isn't a full substitute — per the original commit's comments, nested `sudo apt-get` calls inside Playwright don't honor `SUDO_ASKPASS`, which is exactly why the precache was added. Net effect: Linux GUI installs that hit `playwright install --with-deps` (or any nested `sudo apt-get`) will hang. macOS/Windows are unaffected.

Could you re-merge `main` and restore:

- `src/main/sudoCreds.ts`
- `import { precacheSudoCredentials } from "./sudoCreds";` in `installer.ts`
- The `precacheSudoCredentials(...)` call + `sudoPrecache.stop()` cleanup in the install flow (around lines 115–146 of the pre-merge installer)

**A couple of smaller items worth a look while you're in there**

1. **Webview post-attachment navigation isn't covered.** `will-attach-webview` validates the initial `src`, but once attached the webview can navigate freely — there's no `will-navigate` handler on the webview's own `webContents`. Worth attaching policy listeners via `app.on("web-contents-created", ...)` so a loopback page can't follow a link to `http://attacker.com:3000`. The only current `<webview>` is the Claw3D viewer at `http://localhost:${port}`, so this is defense-in-depth rather than a concrete exploit today.

2. **`runHermesDoctor`** still uses `execSync` with a shell-built command string. Quoting is careful, but `execFileSync` would eliminate the shell entirely and match how the other helpers now invoke the CLI.

**What's great about this PR (for the record)**

- Main window flips to `sandbox: true` + `contextIsolation: true` + `nodeIntegration: false` + `webSecurity: true` + `allowRunningInsecureContent: false`.
- Askpass dialog moves off `nodeIntegration: true` to a dedicated sandboxed preload with strict CSP (`script-src 'none'`) and `event.sender === win.webContents` validation.
- Clean `security.ts` allowlist for external/navigation/webview URLs, with all `shell.openExternal` calls routed through one helper.
- Profile-name regex (`^[a-z0-9_][a-z0-9_-]{0,63}$`) applied to `profileHome`, create/delete/setActive, and `listProfiles` filtering — closes the path-traversal and option-flag-injection surface.
- Sandbox-compatible preload (drops `@electron-toolkit/preload` runtime import, inlines the minimal subset actually used).
- Windows GUI install now uses the official `install.ps1` with explicit `HermesHome`/`InstallDir` and `python -m hermes_cli.main` invocation — solves the missing-WSL failure.
- ~450 lines of new tests locking it all in.

Once the sudo precache is restored, happy to take another pass.

fathah:
As WSL users are stuck in sudo permissions, we cant remove those. Please fix that, and let's merge

ASTairov:
@fathah 
Thanks for catching this. You were right: the previous merge resolution accidentally dropped the Linux sudo precache path from `8119b64`.

Fixed in `411198b`:

- restored `src/main/sudoCreds.ts`
- restored `precacheSudoCredentials(...)` in the install flow, including `sudoPrecache.stop()` cleanup
- kept the restored sudo prompt within this PR's security model:
  - `nodeIntegration: false`
  - `contextIsolation: true`
  - `sandbox: true`
  - strict CSP
  - sender-bound IPC validation
  - navigation/popups/webview attachment denied
- changed `runHermesDoctor` from shell-built `execSync` to `execFileSync`
- added post-attachment webview navigation/redirect policy via `app.on("web-contents-created", ...)`
- added regression coverage for the restored sudo precache wiring and webview policy

Verified locally:

- `npm run typecheck` ✅
- `npm test` ✅ 304/304
- `npm run build` ✅

`npm run lint` still reports pre-existing unrelated repo-wide lint/prettier issues outside this patch, but targeted ESLint on the changed files has no errors.

PR #94, opened 2026-05-10
security: keep remote API keys in the main process

ASTairov:
Stop returning the stored remote API key from get-connection-config. The renderer now receives only the connection mode, remote URL, and a hasApiKey flag, which keeps the saved bearer token out of the renderer preload contract.

Keep remote startup and Settings checks working by letting the main process apply the stored key when testRemoteConnection targets the currently configured remote URL. Explicit keys still override this path for first-time connection tests.

Avoid a renderer-driven key forwarding path by preserving an existing key only when the remote URL is unchanged. If the URL changes without a newly supplied key, the stored key is cleared instead of being sent to a different host.

Mask saved remote keys in Settings and replace newly entered keys with the mask after save so React state does not retain the secret longer than needed.

Adds regression coverage for public connection config redaction, same-URL key preservation, different-URL key clearing, and main-process authenticated remote health checks.

Verified with: npm test -- connection-config-security.test.ts ipc-handlers.test.ts; npm run typecheck; npm run build; npx vitest run --testTimeout 60000. Note: plain npm test hits the existing Windows timeout in session-cache-sync.test.ts unless the timeout is raised.


fathah:
Please resolve conflicts

ASTairov:
Resolved conflicts with current main.

I merged upstream/main into this branch, kept remote API keys confined to the main process, reconciled the new SSH connection config with the public renderer config, and added coverage to ensure SSH settings can be exposed without leaking the stored remote API key.

Verification:
- targeted security/SSH tests: 26 passed
- npm test: 276 passed
- npm run build: passed

ASTairov:
Resolved the merge conflict against the latest `fathah/main` and pushed merge commit `75cca99`.

Conflict notes:
- The only manual conflict was in `package-lock.json`.
- I kept the lockfile from the updated `main` side because the merged `package.json` does not add dependency changes over `main`, so this avoids unnecessary npm lockfile churn.
- Rechecked the PR behavior after the merge: renderer receives `hasApiKey` instead of the stored remote API key, main process still applies the saved key for same-URL remote health checks, and URL changes without a new key clear the saved key.

Verification run after resolving:
- `npx vitest run --testTimeout 60000` — 20 files / 311 tests passed.
- `npm run build` — typecheck and Electron/Vite production build passed.

PR #96, opened 2026-05-10
feat: add local CLI provider

ASTairov:
Add a generic Local CLI provider instead of a Codex-specific provider in the renderer. The provider stores desktop-only CLI command settings separately from model config, exposes get/set IPC through preload, and hides API-key/Base URL credential UI that does not apply to local CLI auth.

Route local chat requests with provider=cli through a main-process CLI runner. The runner detects the codex command and uses a safer Codex exec adapter with stdin prompts, ephemeral read-only sandboxing, no color output, and output-last-message capture; other commands use a constrained stdin/stdout fallback. Command fields are normalized to reject shell fragments.

Also make install readiness work for this local provider path by treating cli as a no-API-key provider and resolving Windows virtualenv Python from venv\\Scripts\\python.exe. Add targeted tests for CLI argument building, command sanitization, provider constants, preload surface, and installer path resolution.


ASTairov:
Resolved the conflicts on this PR by merging the latest `upstream/main` into `codex/block-6-local-cli-provider`.

Kept the Local CLI provider changes while preserving the newer upstream SSH/remote/install/i18n work:
- Local CLI send path remains wired through `LOCAL_CLI_PROVIDER`.
- SSH tunnel imports and remote-mode handling from `main` are preserved.
- `cli` is still treated as a local provider during install/config checks.
- `auth.json` credential detection from `main` is preserved.
- Preload now exposes both Local CLI config IPC methods and SSH config IPC methods.
- Local CLI settings copy is preserved in `en`, `es`, and `pt-BR`.

Verified locally:
- `npm run typecheck` ✅
- `npx vitest run tests/local-cli.test.ts tests/installer-utils.test.ts tests/installer-paths.test.ts tests/preload-api-surface.test.ts tests/ssh-options.test.ts tests/ssh-remote.test.ts` ✅
- `npm run build` ✅

Note: on Windows, full `npm test` initially required `npm rebuild better-sqlite3` because the native module had been rebuilt for Electron ABI. After that, the only remaining issue was the known 5s timeout in `tests/session-cache-sync.test.ts`; the same file passes with `--testTimeout=30000`.

Issue #91, opened 2026-05-10
no working on Macos, not accecptin api key opencode, not comminication with remote oder local agent


PDCem:
no working on Macos, not accecptin api key opencode, not comminication with remote oder local agent

Issue #81, opened 2026-05-07
Bug: WSL Hermes installation not detected + install script domain unreachable


catowabisabi:
### Environment
- OS: Windows 11 + WSL2 (Ubuntu)
- Hermes Desktop version: latest
- Hermes Agent version: v0.9.0 (2026.4.13)
- Hermes installed in: WSL2 (`/home/enomars/.hermes/`)

### Problem 1: `hermes.computer` domain is unreachable (NXDOMAIN)

The built-in installer attempts to fetch the install script from `https://hermes.computer/install.sh`, but this domain does not resolve:

```
** server can't find hermes.computer: NXDOMAIN
```

Verified on both Windows (via DD-WRT router DNS) and WSL2. The domain appears to be dead or not yet registered, which causes the installation to fail with exit code 1 regardless of system state.

### Problem 2: Existing WSL2 Hermes installation is not detected

I already have Hermes Agent fully installed and running inside WSL2 (`hermes-gateway.service` active on `127.0.0.1:8642`). However, when selecting **Local** mode in the Desktop App, it ignores the existing installation and attempts to reinstall from scratch.

The installer tries to spawn a Windows-native Python process:

```
spawnSync C:\Users\enoma\.hermes\hermes-agent\venv\bin\python ENOENT
```

This fails because Hermes is installed inside WSL2, not natively on Windows. The Desktop App does not check for an existing WSL2 installation at `127.0.0.1:8642` before attempting a fresh install.

### Problem 3: Remote mode does not expose full Desktop App functionality

As a workaround I tried **Remote** mode pointing to `http://127.0.0.1:8642`, which successfully connects to the WSL2 gateway. However, Remote mode disables most of the Desktop App's features (settings, gateway management, memory, tools, etc.), making it an inadequate workaround for users who already have Hermes running in WSL2.

### Expected Behaviour

1. The installer should check if a Hermes gateway is already reachable at `127.0.0.1:8642` before attempting installation, and skip the install step if detected.
2. The install script URL should be updated or the app should handle DNS failures gracefully with a clear error message.
3. WSL2 installations should be treated as equivalent to a local installation when the gateway is reachable on localhost.

### Steps to Reproduce

1. Install Hermes Agent inside WSL2 and start `hermes-gateway` via systemd
2. Confirm gateway is running: `ss -tlnp | grep 8642` shows `LISTEN 127.0.0.1:8642`
3. Download and launch Hermes Desktop on Windows
4. Select **Local** mode
[...]

Issue #41, opened 2026-04-30
Unable to chat with Hermes on Desktop - get 'error: invalid API key'


serena-ven:
Hi I have downloaded the desktop app successfully, but every time I try to chat I get 'error: invalid API key'. I have inserted my Anthropic API key both under credentials pool and LLM providers. here are the details: 

Mac M3, macOS
Connection mode: Local
hermes dashboard running (API server connected)
Error: "Invalid API key" in chat
Tried: hermes-local, Anthropic key, both in Credentials Pool and LLM Providers

It would be great to know how to sort this out.

Many thanks. 
Serena

yshao0711:
same issue here with latest release

turnipy:
Same on linux with .appimage (Cachy OS)

serena-ven:
Update: I actually got Hermes itself to sort it out. It took a while but it did identify the issues in the end. You may want to try the same.

turnipy:
**I did the same in Hermes and it was fixed - see fix output below:**

My prompt: 
_When I use https://github.com/fathah/hermes-desktop installed via appimage on this pc, all works except chat which
says Error: Invalid API key, but I am using chatgpt 5.5 via oauth so it should work_

────────────────────────────────────────

⚕ Hermes 
    Found it. This is not your ChatGPT / OpenAI Codex OAuth failing.

    Root cause:
    hermes-desktop detects the local Hermes API server at:

    http://127.0.0.1:8642

    The /health endpoint returns OK without auth, so the desktop app decides “API server is available” and uses it for chat.

    But your actual chat endpoint:

    /v1/chat/completions

    requires API_SERVER_KEY from ~/.hermes/.env.

    In hermes-desktop local mode, the app does not send that API_SERVER_KEY as a Bearer token. So Hermes API server returns:

    Error: Invalid API key

    That error is about the local Hermes API server auth, not about ChatGPT 5.5 OAuth.

    I verified:
    - Hermes CLI is logged into OpenAI Codex OAuth.
    - Provider is openai-codex.
    - Model is gpt-5.5.
    - API call without API_SERVER_KEY gets 401 Invalid API key.
    - API call with API_SERVER_KEY succeeds and gpt-5.5 replies correctly.

    I applied the local workaround by creating:

    /home/[home]/.hermes/desktop.json

    with the desktop connection set to remote mode pointing at localhost:

    connectionMode: remote
    remoteUrl: http://127.0.0.1:8642
[...]

Issue #30, opened 2026-04-29
local connecting to Hermes Agent API: Error: Invalid API key


Sekubamba:
Hi since I started using Hermes Desktop I'm not able to connect it to hermes-agent API: http://127.0.0.1:8642/v1. Please let me know what I have to do. Thank you.

pmos69:
This is the bug fixed by [PR #241](https://github.com/fathah/hermes-desktop/pull/241) (`fix: send api server auth for desktop chat`). When you have `API_SERVER_KEY` set in `config.yaml` (which is the default after recent installs), the desktop's chat must include it as `Authorization: Bearer …` on every request to the local gateway — without that header the gateway returns `Invalid API key`. The PR adds that header on the local-mode chat path. It's currently open and mergeable; this issue should auto-close when it lands.

PR #102, opened 2026-05-11
feat: support OpenAI Codex provider setup

leejamesss:
## Summary

Adds first-class setup support for the `openai-codex` provider in Hermes Desktop.

`openai-codex` is different from most providers currently shown in Desktop: it is backed by the OpenAI Codex CLI / OAuth flow, so Desktop should not require the user to enter or validate a static API key before considering the local Hermes install usable. This PR teaches both the main-process install checks and the renderer setup UI about that no-key provider path.

## What changed

- Treat `openai-codex` as a provider that does **not** require an API key during local install validation.
- Add OpenAI Codex CLI to the provider setup / model-selection constants used by the renderer.
- Add a setup card for OpenAI Codex CLI with no-key instructions instead of an API-key input flow.
- Add localized copy for the new provider setup path:
  - `en`
  - `zh-CN`
  - `es`
  - `pt-BR`
- Add tests covering:
  - provider metadata / constants for `openai-codex`
  - no-key provider detection in the installer/provider helper logic

## Motivation / related reports

Several existing reports show users getting stuck around provider credentials, local-vs-remote setup, or Desktop treating a working Hermes Agent setup as unusable because the setup flow expects the wrong kind of auth state:

- Related: #30 reports local connection failing with `Error: Invalid API key`.
- Related: #41 reports Desktop chat failing with `error: invalid API key` even after adding provider credentials.
- Related: #91 mentions macOS provider/API-key acceptance and local/remote communication failures.
- Related context: #81 notes that remote mode is an inadequate workaround when users already have Hermes reachable locally but Desktop gating/install checks get in the way.

This PR is intentionally narrower than those broader connection/auth issues: it only fixes the Desktop setup path for providers like `openai-codex` where the correct auth flow is CLI/OAuth-based and no API key should be collected by Desktop.

## Behavior before

When configuring Hermes Desktop locally, provider setup assumed providers needed a static API key. For `openai-codex`, that creates a bad UX: the provider can be valid from Hermes Agent's perspective, but Desktop's setup/install validation still expects API-key-style configuration.

## Behavior after

Users can select OpenAI Codex CLI during setup, see instructions for authenticating through the CLI/OAuth flow, and continue without entering an API key. Local install checks recognize `openai-codex` as a no-key provider.

## Test plan

- [x] `npm run typecheck`
- [x] `npm test`
- [x] `npx eslint <touched files>`
- [x] Manually built the Electron app and verified the packaged app contains the `openai-codex` provider metadata in both main and renderer bundles.
- [x] Verified local Desktop config can remain in local mode without a remote-mode workaround.

## Notes

Full `npm run lint` still reports pre-existing lint errors in unrelated files, including existing `.agents` skill reference files and existing screen/test files. The files touched in this PR pass targeted ESLint.


fathah:
conflict

leejamesss:
@fathah Conflict Resolved

pmos69:
Heads-up: my PR #262 expands the provider list in `src/renderer/src/constants.ts` and the i18n locale `constants.ts` files. Looks like your PR is already conflicting with main, so when you rebase you'll likely also pick up overlap with #262 — flagging so you can choose merge ordering with the maintainer if useful.

leejamesss:
Thanks for the heads-up @pmos69. I updated this branch against the latest main and resolved the overlap in the provider constants/i18n area. Typecheck passes locally.

PR #107, opened 2026-05-11
[codex] Use remote API for management views

maxtrigify:
## Summary
- Adds a remote API adapter for pure HTTP remote mode.
- Routes sessions, session cache/search, profiles, memory, SOUL.md, tools, skills, and gateway status IPC handlers through Hermes Agent API endpoints when connection mode is remote.
- Keeps existing local and SSH behavior unchanged; skill install/uninstall still report unsupported until the Agent API exposes those write endpoints.

## Companion PR
Requires the Hermes Agent API surface from NousResearch/hermes-agent#23742.

## Validation
- npm run typecheck
- npm test
- npm run build
- npm run build:unpack

### DIFF
--- a/src/main/hermes.ts
+++ b/src/main/hermes.ts
@@ -195,6 +195,9 @@ function sendMessageViaApi(
     "Content-Type": "application/json",
     ...getRemoteAuthHeader(),
   };
+  if (_resumeSessionId) {
+    headers["X-Hermes-Session-Id"] = _resumeSessionId;
+  }
 
   let sessionId = _resumeSessionId || "";
   let hasContent = false;
@@ -226,10 +229,7 @@ function sendMessageViaApi(
       probeUrl,
       {
         method: "POST",
-        headers: {
-          "Content-Type": "application/json",
-          ...getRemoteAuthHeader(),
-        },
+        headers,
       },
       (res) => {
         let raw = "";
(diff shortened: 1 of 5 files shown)


pmos69:
Thanks for putting this together. I think the remote-management goal is still the right direction, but this PR is now tied to an API foundation that has gone stale.

The Desktop changes here depend on NousResearch/hermes-agent#23742, which is still draft, currently conflicting, and has not moved since May 11. Since then, the agent side has merged narrower pieces that look like the better foundation for Desktop remote-mode work:

- NousResearch/hermes-agent#33134 merged the `/api/sessions` session control API.
- NousResearch/hermes-agent#33016 merged read-only `/v1/skills` and `/v1/toolsets`.

That means the next Desktop implementation should probably not build against the broad #23742 endpoint shape. A more mergeable replacement would be split around the merged surfaces first, for example:

1. wire the Desktop Sessions screen in pure remote mode to `/api/sessions` from agent#33134;
2. add a limited/read-only Skills or Toolsets view from `/v1/skills` and `/v1/toolsets` from agent#33016;
3. keep Profiles, Memory, Persona, full Skills management, Gateway status/control, etc. gated until the corresponding smaller agent APIs land.

I am going to close this as superseded/stale rather than leave it blocked on #23742. The intent is worth preserving, but the implementation target should be updated to the APIs that actually merged upstream.

Issue #109, opened 2026-05-11
Windows 11 install stuck on switching to root user (see below)


LizardKing2212:
> Running official Hermes install script...
> 
> 
> ┌─────────────────────────────────────────────────────────┐
> │             ⚕ Hermes Agent Installer                    │
> ├─────────────────────────────────────────────────────────┤
> │  An open source AI agent by Nous Research.              │
> └─────────────────────────────────────────────────────────┘
> 
> ✓ Detected: linux (ubuntu)
> → Checking for uv package manager...
> ✓ uv found at ~/.local/bin (uv 0.11.2 (x86_64-unknown-linux-gnu))
> → Checking Python 3.11...
> ✓ Python found: Python 3.11.15
> → Checking Git...
> ✓ Git 2.43.0 found
> → Checking Node.js (for browser tools)...
> ✓ Node.js v22.22.2 found (Hermes-managed)
> → Checking internet connectivity for package install and web tools...
> ✓ Internet connectivity looks good
> → Checking ripgrep (fast file search)...
> ✓ ripgrep 14.1.0 found
> → Checking ffmpeg (TTS voice messages)...
> ✓ ffmpeg 6.1.1-3ubuntu5 found
> → Installing to /home/kingm/.hermes/hermes-agent...
> → Existing installation found, updating...
> From https://github.com/NousResearch/hermes-agent
>    1d4a4997b..3e7145e0b  main       -> origin/main
> Already on 'main'
> Your branch is behind 'origin/main' by 1 commit, and can be fast-forwarded.
>   (use "git pull" to update your local branch)
> From https://github.com/NousResearch/hermes-agent
>  * branch                main       -> FETCH_HEAD
> Updating 1d4a4997b..3e7145e0b
> Fast-forward
>  cli.py                                  |  105 +--
>  gateway/run.py                          |  101 +--
>  hermes_cli/commands.py                  |    2 -
>  hermes_cli/goals.py                     | 1440 ++-----------------------------
>  run_agent.py                            |   19 -
>  tests/cli/test_cli_goal_interrupt.py    |   17 +-
>  tests/gateway/test_goal_verdict_send.py |   14 +-
>  tests/hermes_cli/test_goals.py          |  899 +------------------
>  tui_gateway/server.py                   |    1 -
>  9 files changed, 112 insertions(+), 2486 deletions(-)
> ✓ Repository ready
> → Creating virtual environment with Python 3.11...
> → Virtual environment already exists, recreating...
[...]

PR #111, opened 2026-05-11
docs: add Japanese documents

eltociear:
I created Japanese translated documents.


pmos69 (changes requested):
Thank you for adding the Japanese documentation. This is a very welcome contribution, and having `README.ja-JP.md` / `CONTRIBUTING.ja-JP.md` would be useful for the project.

I do think this needs a refresh before we merge it, because the English source docs have changed since this PR was opened. The Japanese README appears to be based on an older version of `README.md`, so a few details no longer match current `main`.

Could you please update the Japanese translations against the current versions of:

- `README.md`
- `CONTRIBUTING.md`

A few concrete things I noticed:

- Current `README.md` uses the local `previews/...` assets and the `hermesagents.cc` download link/button, while this PR still uses older GitHub user-attachment images and older release/install wording.
- The current README has a Windows/WSL installer note that is missing from the Japanese translation.
- Some links in the Japanese docs point to `NousResearch/hermes-desktop`, but the current repository/issues are under `fathah/hermes-desktop`.
- `CONTRIBUTING.md` now includes a “Keep Pull Requests Small” section that is missing from `CONTRIBUTING.ja-JP.md`.

Once the Japanese docs are synchronized with the current English docs, this should be much easier to review and merge. Thanks again for contributing this translation.

eltociear:
Hi @pmos69, thank you so much for the thorough and kind review! 🙏

You were absolutely right — my fork was based on an older `main`, which is why the Japanese docs had drifted from the current English source. I've rebased the branch onto the latest upstream `main` and re-synchronized both translations from scratch. The branch now diffs cleanly against `main`: only the new `日本語` language links plus the two new Japanese files.

Here is how I addressed each point you raised:

- **Assets & download link** — `README.ja-JP.md` now uses the local `previews/...` assets and the `hermesagents.cc` download button/link, matching the current `README.md` (replaced the older GitHub user-attachment images and release wording). The Discord badge was also updated to the current Telegram badge.
- **Install section** — Restructured to match current `main` (download button + Windows / Fedora), including the previously missing **Windows/WSL installer note** (the passwordless-sudo workaround, referencing #109).
- **Preview table** — Updated to the current screenshot lineup (Chat, Profiles, Models, Providers, Tools, Skills, Schedules, Gateway, Persona, Kanban, Office, Settings).
- **"Keep Pull Requests Small"** — Added the missing section to `CONTRIBUTING.ja-JP.md` (translated as「プルリクエストは小さく保つ」).
- **Repo/issue links** — The Japanese docs now mirror the current English source exactly, so the links are consistent with `README.md` / `CONTRIBUTING.md` on `main`.

Please take another look whenever you have a moment, and let me know if anything still looks off. Thanks again for helping get this in good shape! 🙇

pmos69:
Thank you for the careful refresh. I rechecked the current diff and this addresses the drift I called out earlier: the README now follows current main, the preview/download assets are updated, the WSL note is included, and CONTRIBUTING has the small-PR guidance.

The remaining repo/issue links match the current English source, so I would not treat those as a blocker for this PR. Thanks again for taking the time to rebase and resync both Japanese docs.

PR #140, opened 2026-05-12
Implement automated manifest generation

krataratha:
Added automated template processing and directory creation for Winget manifests.

### DIFF
--- a/scripts/generate-winget-manifests.mjs
+++ b/scripts/generate-winget-manifests.mjs
@@ -1,18 +1,8 @@
 // scripts/generate-winget-manifests.mjs
-//
-// Fills the YAML templates in build/winget/ with the current version,
-// installer URL, and SHA256 of the NSIS installer in dist/, and writes
-// the result under dist/winget/manifests/n/NousResearch/HermesDesktop/<version>/.
-//
-// Run from CLI: VERSION=0.2.3 PUBLISH_OWNER=fathah node scripts/generate-winget-manifests.mjs
-// Or import as ESM and call generateWingetManifests({ rootDir, version, name, publishOwner }).
-
 import { createHash } from "node:crypto";
-import { readFileSync, writeFileSync, mkdirSync, existsSync } from "node:fs";
+import { readFileSync, writeFileSync, mkdirSync, existsSync, readdirSync } from "node:fs";
 import { join } from "node:path";
-import { fileURLToPath } from "node:url";
 
-// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
 export function generateWingetManifests({
   rootDir,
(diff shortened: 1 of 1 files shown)


fathah:
Thanks for the contribution! Before this can be merged, there are a few issues with the diff that need to be addressed.

## Blocking: the script no longer parses

The new closing `}` is added right after `console.log(...)`, but the original function body continues below it:

```js
  console.log(`✅ Manifests generated in: ${manifestDir}`);
}

  const replacements = {
    VERSION: version,
    ...
```

That leaves the rest of `generateWingetManifests()` orphaned outside the function. The script will fail to run (top-level code referencing `version`, which is only in function scope). It looks like the new block was pasted in without removing the previous implementation — could you delete the old `replacements` / template-writing code that now sits below the closing brace?

## Placeholder format mismatch

The new code replaces mustache-style tokens (`{{version}}`, `{{sha256}}`, `{{installerUrl}}`, ...), but the existing templates in `build/winget/` use uppercase tokens (`VERSION`, `INSTALLER_URL`, ...) — that's what the original `replacements` map targeted.

Only `scripts/generate-winget-manifests.mjs` is changed in this PR, so unless the templates are updated in the same change, none of the placeholders will actually be substituted and the generated manifests will contain literal `{{version}}` strings.

Two options:
1. Update the templates in `build/winget/` to use the new `{{...}}` style in this PR, or
2. Keep the existing uppercase token format and just add the directory-creation / template-loop logic.

## Smaller things worth keeping

- The CI artifact hint in the error message was useful — please keep `"...or download the windows-artifacts CI artifact into dist/."`.
- The header docstring documented CLI usage (`VERSION=... PUBLISH_OWNER=... node scripts/generate-winget-manifests.mjs`); it'd be good to preserve it.
- Worth double-checking that `fileURLToPath` is genuinely unused before removing the import.

## Scope note

The PR description says "Added automated template processing and directory creation," but the existing script already did both of those things. Could you update the description to describe what specifically changes (e.g. switching token format, restructuring the template loop, etc.)? That'll make review easier.

Once the function is closed cleanly and the placeholder format matches the templates, happy to take another look.

pmos69:
Thanks again for the contribution. Since this has not been updated after the maintainer feedback, I am going to close it for now.

The PR as written still has the blocking issues noted above: the script's control flow is broken by the early closing brace, and the placeholder handling/output generation no longer matches the current Winget template flow. Current `main` also already has the useful manifest-generation pieces in a working form: output directory creation, template processing, the CLI entrypoint, the CI artifact hint, and explicit template-to-output filename mapping.

If there is a concrete remaining improvement to the Winget generator, a fresh focused PR would be welcome.

PR #152, opened 2026-05-13
feat: add OpenCode Go/Zen provider support (fix #148)

### LINKED ISSUE #148: Bug: OpenCode Go provider missing from Model Provider dropdown
Hi,

I found a bug in hermes-desktop where the "OpenCode Go" provider does not appear in the "Model Provider" dropdown on the Providers settings page, even though it is fully supported by the backend.

What I observed:

The LLM Providers block correctly shows "OpenCode Go" and accepts API keys for it

The CLI (hermes model) lists OpenCode Go as an available provider

The backend includes opencode-go in HERMES_OVERLAYS and CANONICAL_PROVIDERS

i18n strings for opencode-go exist (opencodeGoApiKey, opencodeGoHint are defined in the English locale)

But the dropdown on the Providers page does NOT include "OpenCode Go" as an option

Root cause:
The dropdown is populated from a hardcoded PROVIDERS.options array in the bundled JS. OpenCode Go was added to the i18n labels and backend but was never added to the options array.

The following entry is missing from PROVIDERS.options:
{ value: "opencode-go", label: "OpenCode Go" }

It also needs corresponding entries in PROVIDERS.labels and PROVIDERS.setup to be fully functional.

Steps to reproduce:

Open hermes-desktop

Go to Settings > Providers

Click the "Model Provider" dropdown

"OpenCode Go" is not listed, even though it is a supported provider

Impact:
Users who configure OpenCode Go must use the CLI to set the provider, or work around it by selecting "Local / Custom". The provider works fine once configured, but the UI does not offer it as a selection.

Expected behavior:
"OpenCode Go" should appear in the Model Provider dropdown alongside other supported providers like MiniMax, Anthropic, OpenRouter, etc.

Thank you for a great product!

chen05321:
## Fixes #148

Adds OpenCode Go and OpenCode Zen as first-class providers in the desktop UI.

Changes:
- PROVIDERS options/labels/setup in constants.ts
- Default models in default-models.ts
- Env key injection + URL mapping in hermes.ts
- i18n for all 5 locales (en, zh-CN, es, id, pt-BR)


fathah:
Thanks for picking this up — #148 is a real bug and the dropdown wiring here is solid. I checked the fix against the `hermes-agent` backend before reviewing and there's one issue that needs addressing before this can land.

## What's correct

- **Both providers are real.** `hermes-agent` registers both `opencode-zen` and `opencode-go` as full provider profiles in `plugins/model-providers/opencode-zen/__init__.py` and `hermes_cli/providers.py`. Adding both to the dropdown is the right call.
- **Base URLs match the backend exactly:** `https://opencode.ai/zen/v1` and `https://opencode.ai/zen/go/v1`.
- **Env keys match:** `OPENCODE_ZEN_API_KEY` and `OPENCODE_GO_API_KEY`.
- **`KNOWN_API_KEYS` injection** in `src/main/hermes.ts` is the right place for these — that path correctly forwards both keys to the CLI subprocess.
- **i18n** covers all 5 locales the project supports.

## What needs fixing

### 1. Default model IDs don't exist on the backend

The backend's canonical allowed-models list lives in `hermes_cli/models.py:331-381`. None of the four model IDs in this PR appear there:

| PR adds | In backend allow-list? |
|---|---|
| `deepseek-v4-flash` (go) | ❌ — DeepSeek isn't carried by Go at all |
| `deepseek-v4-lite` (go) | ❌ — same |
| `gemini-2.5-flash` (zen) | ❌ — closest is `gemini-3-flash` |
| `gemini-2.5-pro` (zen) | ❌ — closest is `gemini-3-pro` / `gemini-3.1-pro` |

Anyone who picks one of these out of the box will get a 4xx from the OpenCode API.

The backend plugin already suggests sensible defaults via its `default_aux_model` settings — `gemini-3-flash` for Zen and `glm-5` for Go. Suggested replacements pulled from the actual allow-lists:

- **opencode-go:** `glm-5`, `kimi-k2.5` (or `qwen3.6-plus`)
- **opencode-zen:** `gemini-3-flash`, `claude-sonnet-4-6` (or `gpt-5.4`)

Full lists for reference:
- Go: `kimi-k2.6`, `kimi-k2.5`, `glm-5.1`, `glm-5`, `mimo-v2.5-pro`, `mimo-v2.5`, `mimo-v2-pro`, `mimo-v2-omni`, `minimax-m2.7`, `minimax-m2.5`, `qwen3.6-plus`, `qwen3.5-plus`
- Zen: `kimi-k2.5`, `gpt-5.4-pro`, `gpt-5.4`, `gpt-5.3-codex`, `gpt-5.2`, `gpt-5.1`, `gpt-5`, `claude-opus-4-6`, `claude-sonnet-4-6`, `claude-haiku-4-5`, `gemini-3.1-pro`, `gemini-3-pro`, `gemini-3-flash`, `minimax-m2.7`, `glm-5`, `qwen3-coder`, …

### 2. `URL_KEY_MAP` entry is either dead code or buggy

```ts
{ pattern: /opencode\.ai/i, envKey: "OPENCODE_GO_API_KEY" },
```

This branch in `src/main/hermes.ts` is gated by `LOCAL_PROVIDERS.has(mc.provider)` (line 502), which only contains `custom`, `lmstudio`, `ollama`, `vllm`, `llamacpp`. Since the new providers aren't in that set, `URL_KEY_MAP` is **never consulted** for proper `opencode-go` / `opencode-zen` selections — the keys reach the CLI via `KNOWN_API_KEYS` instead, which works for both.

If someone *does* reach that branch (Local/Custom path with an OpenCode URL), the regex matches **both** Zen and Go URLs and unconditionally picks `OPENCODE_GO_API_KEY` — so a Zen URL would silently send the Go key.

Two options:
- **Drop the entry** (recommended — it's dead code for the proper provider path).
- Or **fix the ordering** with a more specific Zen pattern ahead of the generic one:
  ```ts
  { pattern: /opencode\.ai\/zen\/v1/i, envKey: "OPENCODE_ZEN_API_KEY" },
  { pattern: /opencode\.ai/i,          envKey: "OPENCODE_GO_API_KEY" },
  ```

## Nits (not blocking)

- `opencodeGoTag: ""` is defined in every locale but never rendered — matches the pattern used by other providers though, so fine to leave.

## Summary

Core fix is correct and worth landing. Two changes needed:
1. Swap the four default model IDs for ones in the backend's allow-list.
2. Drop (or reorder) the `URL_KEY_MAP` entry.

Happy to take this once those are in.

pmos69:
Heads-up: my PR #262 also expands the provider list in `src/renderer/src/constants.ts` (plus the per-locale `constants.ts` i18n files). Both PRs are currently mergeable independently, but whichever lands second will need a small textual rebase — adjacent array entries. Just flagging so it's not a surprise. Happy to coordinate ordering if useful.

pmos69 (changes requested):
Thanks for taking this on. The core issue is still valid: OpenCode Go/Zen should be selectable in the Desktop provider UI, and #148 is not fully addressed on current `main`.

I think this PR needs to be refreshed before it can land, mostly because the provider code has changed a lot since it was opened and this branch now conflicts with current `main`.

One important correction to the earlier discussion: we should not treat the Hermes Agent curated OpenCode model lists as the source of truth for Desktop defaults. Those curated lists can lag behind the actual provider. OpenCode's own Go/Zen model lineups have also moved since this PR was opened, and Desktop already follows a better pattern for providers that support model listing: poll the provider's `/models` surface where possible, rather than baking a broad static catalog into Desktop.

Could you please narrow and refresh this PR around the provider wiring itself?

Suggested scope:

1. Add OpenCode Go and OpenCode Zen to the provider dropdown / labels on current `main`.
2. Add or preserve the setup entries for the documented endpoints:
   - OpenCode Zen: `https://opencode.ai/zen/v1`
   - OpenCode Go: `https://opencode.ai/zen/go/v1`
3. Ensure the relevant env vars are forwarded to the Hermes subprocess:
   - `OPENCODE_ZEN_API_KEY`
   - `OPENCODE_GO_API_KEY`
4. Make sure these providers work with Desktop's live model discovery path, in line with other providers that expose a model list.
5. Avoid adding a broad static default-model catalog in Desktop. If a fallback is needed, keep it very small and documented; otherwise let live discovery populate models.

A few things to avoid in the refresh:

- Please do not add a generic `/opencode\.ai/ -> OPENCODE_GO_API_KEY` URL mapping. Zen and Go share the same host family, so a broad host regex can route Zen through the Go key. If URL-derived key mapping is needed, it should distinguish the Zen and Go paths explicitly, or be omitted if the first-class provider path already forwards the right env vars.
- Please do not copy the current Hermes Agent curated model lists into Desktop. They can become stale, and Desktop's UX should be based on provider discovery where the provider supports it.
- Please update all current locales on `main`; the project now has more locale files than the five touched in this branch.

With those changes, this would be a much smaller and more durable fix for #148. Thanks again for the contribution.

PR #169, opened 2026-05-13
feat: Remote server management, SSH tunnel setup wizard, Cloudflare tunnel, and auto-connect

guberm:
Summary

- Remote Server Setup Wizard: 4-step guided wizard (deploy management server over SSH → verify connection → configure tunnel → success). Handles password-based SSH deploy of the Python management server, Cloudflare tunnel configuration, SSH key generation/installation for tunnel-free key-based access, and manual URL entry.

- SSH remote management: Full `ssh-remote.ts` backend that executes all management operations (profiles, config, sessions, cron, toolsets, skills, memory, soul, gateway) via SSH exec without requiring a locally-exposed port.

- HTTP remote management proxy: `manage-proxy.ts` + `hermes-manage-server.py` for `remote` mode — POST `/manage` API server that dispatches to the local hermes-agent and is embedded in the app bundle via a Rollup `.py → string` transform plugin (no separate resource file).

- SSH key tunnel setup: Wizard generates an ed25519 keypair, installs the public key on the remote server via SFTP (using the deploy password), saves the private key to `userData/hermes-remote.key`, then switches the connection mode to `ssh` for key-based auto-reconnect on every app launch.

- Cloudflare Tunnel integration: New Tunnel section in Gateway settings — start/stop cloudflared tunnel, quick (TryCloudflare) and named tunnel modes, copy public URL to clipboard. Wizard Step 3 also supports configuring Cloudflare from the remote host.

- Auto-connect: Toggle in Gateway settings that starts the gateway automatically on app launch.

- Gateway status fixes for remote/SSH modes: `gateway-status` IPC now checks `gateway.pid` on remote with a fallback to `/health` probe via the SSH tunnel; `get-tunnel-status` and `get-tunnel-config` return SSH-tunnel-aware responses; `loadConfig` in the Gateway UI wraps each call independently so one failure doesn't silence the rest.

- Model forwarding fix: Selected model is threaded through the chat send chain and not forwarded to the remote gateway, preventing 400 errors from remote deployments that don't support the client-chosen model.


fathah:
conflict

guberm:
Conflicts resolved — merged upstream/release into mg branch. Took upstream's Chat.tsx hook refactoring (useChatActions, useModelConfig, useFastMode, useChatScroll, useLocalCommands) and re-applied our model-forwarding fix so the selected model is correctly passed through to sendMessage in all code paths. Branch updated and pushed.

pmos69:
Thanks for this @guberm — there's a lot of good thinking here, and several of the ideas are things the project still wants. Unfortunately the PR as a whole isn't mergeable anymore, so I'm going to close it as superseded and capture the salvageable parts as focused follow-ups.

## Why close

- **Conflicting + stale.** The branch is in a `CONFLICTING` state and hasn't been updated since mid-May; main has moved substantially since.
- **SSH remote/tunnel already shipped, independently.** The core contribution here — SSH-exec management without a locally-exposed port — now lives in `src/main/ssh-remote.ts` and `src/main/ssh-tunnel.ts`, which landed via later PRs (e.g. #304 for the systemd gateway path, plus follow-up fixes). So that direction is already in `main` through a different implementation.
- **The bundled management server is an abstraction the project moved away from.** `hermes-manage-server.py` + `manage-proxy.ts` (the embedded Python HTTP proxy) never merged, and the current approach is to talk to the merged hermes-agent APIs directly rather than proxy every local operation through a custom server.
- **Remote-mode design diverged.** Current code deliberately *gates* unsupported screens behind `RemoteNotice` (see `Layout.tsx`) until real agent APIs exist, rather than removing the remote-mode blockers wholesale. This PR pushes the opposite direction.
- **Scope.** It also bundles unrelated release-workflow/package edits and chat-model forwarding, which makes the security-sensitive SSH/key/deploy parts hard to review in one surface.

## Worth salvaging as separate, focused PRs

These are genuinely missing from `main` today and would be welcome on their own:

1. **Cloudflare Tunnel support** — no `cloudflared` integration exists yet. Good as its own small PR with a security review of the exposure model.
2. **Auto-connect** (start gateway on launch) — not currently implemented.
3. **A guided remote/SSH setup wizard** — there's a manual remote/SSH config form in Settings today, but no wizard. A wizard aligned with the existing `ssh-remote.ts` / `ssh-tunnel.ts` would be a nice UX layer on top.

One security note for whoever picks up the wizard: this PR generated an ed25519 keypair and stored the **private key in plaintext at `userData/hermes-remote.key`**. A revived wizard should store that key through the encrypted vault rather than plaintext at rest.

Closing now — thanks again for the contribution and the ideas, several of which are on the roadmap in smaller pieces.

PR #172, opened 2026-05-14
feat(chat): add command history navigation with arrow keys

marcgloor:
## Summary

- Pressing `ArrowUp` in the chat input scrolls back through previously sent messages (like a terminal shell history)
- Pressing `ArrowDown` scrolls forward; going past the newest entry restores whatever was typed before navigating
- Only active on single-line input (so multi-line messages still get normal cursor movement); also disabled while the slash-command menu is open
- Works for both regular sends and Quick Ask (`/btw`) sends

## Implementation

Three new state variables in `Chat.tsx`:
- `inputHistory` — array of sent message strings, grows on each send
- `historyIndex` — current position while navigating (-1 = not navigating)
- `historyDraft` — saves the current draft so it can be restored on `ArrowDown` past the newest entry

The textarea is resized and the cursor placed at end-of-text after each navigation step.

## Test plan

- [x] Send a few messages, then press `ArrowUp` to recall them in order
- [x] Press `ArrowDown` to move forward; pressing past the newest restores the original draft
- [x] Confirm arrow keys still move the cursor normally inside multi-line (Shift+Enter) input
- [x] Confirm slash-command menu arrow navigation is unaffected

### DIFF
--- a/src/renderer/src/screens/Chat/Chat.tsx
+++ b/src/renderer/src/screens/Chat/Chat.tsx
@@ -248,6 +248,9 @@ function Chat({
   const [slashMenuOpen, setSlashMenuOpen] = useState(false);
   const [slashFilter, setSlashFilter] = useState("");
   const [slashSelectedIndex, setSlashSelectedIndex] = useState(0);
+  const [inputHistory, setInputHistory] = useState<string[]>([]);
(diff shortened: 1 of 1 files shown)


marcgloor:
Freaking awesome mate! Keep up the good work!

PR #179, opened 2026-05-14
fix: hide Windows subprocess consoles

ASTairov:
## Summary

This PR fixes Windows command prompt windows briefly flashing and stealing focus while Hermes Desktop is working.

The issue happens when the Electron main process launches child processes without `windowsHide: true`. On Windows, those subprocesses can briefly create visible console windows, interrupting typing or focus in other applications.

## What Changed

- Added a shared Desktop subprocess helper in `src/main/process-options.ts`.
- The helper forces `windowsHide: true` while preserving all existing child process options.
- Applied the hidden subprocess option to Desktop-launched processes for:
  - Hermes Python CLI calls
  - Hermes gateway startup
  - installer / update / doctor / version checks
  - backup / import / dump commands
  - cron commands
  - profile commands
  - skill install/search/uninstall commands
  - SSH tunnel and SSH remote subprocesses
- Added unit tests for the helper to verify:
  - it always enables hidden Windows subprocess windows
  - it does not mutate the caller's original options object

## Scope

This PR is intentionally scoped to Hermes Desktop only.

It does not modify Hermes Agent's internal Python subprocess behavior. The reported issue was reproduced and fixed at the Desktop subprocess-launch layer.

## Manual Verification

Tested on Windows with Hermes Desktop running normal agent work.

Before this fix:
- command prompt windows flashed repeatedly
- the flashing windows stole focus
- typing in other apps was interrupted

After this fix:
- subprocess windows no longer flash
- focus is not stolen
- Hermes continues working normally

## Test Plan

- [x] `npx vitest run tests/process-options.test.ts`
- [x] `npm run typecheck`
- [x] `git diff --check`
- [x] `npx eslint --no-cache src/main/process-options.ts tests/process-options.test.ts src/main/cronjobs.ts src/main/hermes.ts src/main/installer.ts src/main/profiles.ts src/main/skills.ts src/main/ssh-remote.ts src/main/ssh-tunnel.ts`
  - 0 errors
  - existing Prettier warnings remain
- [x] `npm test`
- [x] `npm run build`
- [x] Manual Windows runtime verification

### DIFF
--- a/src/main/cronjobs.ts
+++ b/src/main/cronjobs.ts
@@ -5,6 +5,7 @@ import { execFile } from "child_process";
 import { HERMES_HOME, HERMES_PYTHON, hermesCliArgs } from "./installer";
 import { profileHome } from "./utils";
 import { isRemoteMode, getApiUrl, getRemoteAuthHeader } from "./hermes";
+import { HIDDEN_SUBPROCESS_OPTIONS } from "./process-options";
 
 export interface CronJob {
   id: string;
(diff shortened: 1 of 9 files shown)


hiltonmccarthy21-wq:
Thank u very much, u are the best,fast and very much relayebel

PR #194, opened 2026-05-15
docs: hierarchical AGENTS.md tree + upstream runtime-kernel proposal

dislovelhl:
Adds AI-readable AGENTS.md documentation across the hermes-desktop codebase (20 files), plus a design proposal for hardening the upstream hermes-agent core runtime with smolagents-inspired structure.

AGENTS.md tree:
  - Root + src/{main,preload,renderer,shared}/
  - src/renderer/src/{assets,components,screens}/ + Chat/hooks/
  - src/shared/i18n/locales/
  - tests/, build/, scripts/, docs/superpowers/ Each file: purpose, key files, AI-agent guidance, dependencies. Every non-root file has a `<!-- Parent: ../AGENTS.md -->` reference; all 19 parent references resolve.

Runtime-kernel proposal (`docs/proposals/2026-05-15-smolagents-runtime-upgrade.md`):
  Three-tier rollout plan for the upstream NousResearch/hermes-agent
  runtime — adopt typed step machine + governance gate first (no
  smolagents dep), add CodeAgent as an optional dep second, defer full
  migration. The concrete kernel implementation lives in the companion
  PR on the hermes-agent fork
  (dislovelhl/hermes-agent#runtime/multistep-loop-kernel).

No source changes to hermes-desktop. SSE contract, IPC, and Electron process boundaries are unaffected.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,98 @@
+<!-- Generated: 2026-05-15 | Updated: 2026-05-15 -->
+
+# hermes-desktop
+
+## Purpose
+
+Hermes Desktop is a native cross-platform Electron application that installs, configures, and runs the [Hermes Agent](https://github.com/NousResearch/hermes-agent) self-improving AI assistant. It provides a GUI for chat, sessions, profiles, memory, skills, tools, scheduling, messaging gateways, and Claw3d Office. It supports both **local** (spawns Hermes at `127.0.0.1:8642`) and **remote** (connects to a remote Hermes API server) modes with SSE streaming.
+
+## Key Files
+
+| File | Description |
+|------|-------------|
+| `package.json` | npm scripts, dependencies. Main entry: `out/main/index.js` |
+| `electron.vite.config.ts` | Electron-Vite config — main externalizes `better-sqlite3`, preload has 2 entry points (index + askpass), renderer aliases `@renderer` |
+| `electron-builder.yml` | Cross-platform packaging config (mac dmg, win nsis, linux AppImage/deb/rpm/snap); GitHub release publisher |
+| `tsconfig.json` / `tsconfig.node.json` / `tsconfig.web.json` | Split TS configs — `node` covers main+preload, `web` covers renderer |
+| `eslint.config.mjs` | Flat ESLint config using Electron Toolkit presets |
+| `vitest.config.ts` | Vitest test runner configuration |
+| `dev-app-update.yml` | electron-updater dev override |
+| `skills-lock.json` | Locked versions of bundled Hermes skills |
+| `README.md` / `README.zh-CN.md` | Project overview (English + Simplified Chinese) |
+| `CONTRIBUTING.md` / `CONTRIBUTING.zh-CN.md` | Contribution guidelines |
+| `LICENSE` | MIT license |
+
+## Subdirectories
+
+| Directory | Purpose |
+|-----------|---------|
+| `src/` | Application source — main, preload, renderer, shared (see `src/AGENTS.md`) |
+| `tests/` | Vitest test suites for main-process and preload logic (see `tests/AGENTS.md`) |
+| `build/` | Packaging assets — icons, mac entitlements, winget templates (see `build/AGENTS.md`) |
+| `scripts/` | Build-time scripts — winget manifest generator (see `scripts/AGENTS.md`) |
+| `resources/` | Runtime resources bundled into asar (icon.png) |
+| `docs/` | Long-form planning docs and design specs (see `docs/AGENTS.md`) |
+| `.github/` | CI/CD — release workflow that builds all platforms |
+| `.agents/` / `.claude/` / `.omc/` | Local AI-agent tooling skills + session state (NOT project code) |
+
+## For AI Agents
+
+### Working In This Directory
+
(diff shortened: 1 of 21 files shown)


pmos69 (changes requested):
Thanks for putting this together. I think the general direction is useful: Hermes Desktop would benefit from AI/contributor guidance that explains the Electron boundaries, the Desktop/Hermes Agent split, and the places where contributors tend to make mistakes.

That said, I do not think this PR should merge in its current form. I am requesting changes, but not because the idea is bad. I think it should be narrowed and rewritten into a smaller, current, Desktop-specific AGENTS guidance set.

## Main reasons to rework this PR

### 1. The Desktop guidance currently names the wrong renderer API

Several AGENTS files tell future agents to use `window.api.*`, for example in the root IPC guidance and in the renderer/Chat guidance. The actual Desktop preload surface is `window.hermesAPI`, and the codebase uses `window.hermesAPI` throughout.

This is not just wording. AGENTS.md files are executable guidance for future coding agents, so this would actively steer future changes toward broken renderer code.

Examples from the PR:

- `AGENTS.md` says IPC is consumed as `window.api.x()`
- `src/renderer/AGENTS.md` says privileged operations go through `window.api`
- `src/renderer/src/AGENTS.md` says use `window.api.*`
- `src/renderer/src/screens/AGENTS.md` says every screen calls `window.api.*`
- `src/renderer/src/screens/Chat/AGENTS.md` and `hooks/AGENTS.md` describe `window.api.chat.*` / `window.api.models.*`

Please rewrite these to match the real `window.hermesAPI` API and current preload shape.

### 2. The upstream Hermes Agent runtime proposal should not live in this Desktop PR

`docs/proposals/2026-05-15-smolagents-runtime-upgrade.md` is explicitly about `NousResearch/hermes-agent`, not this Desktop repo. I think that material should be moved to the upstream Hermes Agent repo, an issue/discussion there, or omitted from this Desktop PR.

There are two reasons:

- It is not Desktop contributor guidance. It describes a Python runtime architecture plan for another repository.
- It is already stale against current upstream. For example, it describes `run_agent.py` as a 15,883-line / 821 KB file, but current upstream `NousResearch/hermes-agent` main is much smaller now. That makes the proposal misleading as checked-in Desktop documentation.

A Desktop AGENTS.md can and should have a small bridge section explaining the boundary with Hermes Agent, but it should not carry a full upstream runtime migration proposal.

### 3. The docs are too detailed in ways that will rot quickly

The new AGENTS files include generated timestamps, exact file sizes, version claims, implementation details, and very specific file descriptions. Those are useful for a one-time map, but expensive to keep correct as living agent guidance.

For maintainability, I would strongly prefer shorter, more durable guidance:

- process boundaries and invariants
- where code for each area lives
- what not to do
- what tests/checks usually matter
- how Desktop relates to Hermes Agent
- links to existing docs where details already exist

Avoid file sizes, “largest file” claims, exact dependency versions, and generated timestamps unless there is automation to keep them current.

### 4. Some guidance invents conventions the app does not consistently follow

For example, `src/main/AGENTS.md` says IPC handlers either return `{ ok: true, data }` or `{ ok: false, error }`. Existing handlers return many shapes: booleans, strings, arrays, `{ success }`, feature-specific objects, etc.

It is fine to recommend following nearby handler patterns or adding typed result objects for new isolated modules, but the docs should not state a repo-wide convention that does not currently exist.

## What I would like this PR to become

I would be happy with a narrower PR that only adds Desktop-specific AGENTS guidance, rebuilt against current `main`, with a smaller tree. Root + the major boundary folders may be enough:

- `AGENTS.md`
- `src/AGENTS.md`
- `src/main/AGENTS.md`
- `src/preload/AGENTS.md`
- `src/renderer/AGENTS.md`
- `tests/AGENTS.md`
- maybe `docs/AGENTS.md` / `scripts/AGENTS.md` if they contain durable guidance

The most important content for Desktop would be:

- Electron boundary rules: main/preload/renderer separation
- use `window.hermesAPI`, never direct `ipcRenderer` in renderer
- keep preload implementation and typings in sync
- main-process guards for local-only IPC in remote/SSH modes
- config/profile/session persistence cautions
- i18n expectations for visible UI text
- testing expectations, including live Desktop testing where relevant
- packaging/release caution where appropriate

I would also include a small bridge section in the root file, along these lines:

```md
## Hermes Agent Boundary

Hermes Desktop installs, configures, launches, and talks to Hermes Agent, but it does not own the upstream agent runtime.

Desktop changes belong here when they affect:
- Electron main/preload/renderer code
- `window.hermesAPI` IPC
- installer/update flows
- config/profile UI
- local/SSH/remote connection handling
- session rendering and Desktop persistence
- packaging and live Desktop testing

Upstream Hermes Agent changes belong in `NousResearch/hermes-agent` when they affect:
- Python agent runtime behavior
- provider implementations
- tools and tool execution
- gateway protocol semantics
- memory/compression/session internals
- CLI behavior exposed independently of Desktop

When a bug crosses the boundary, first identify whether Desktop is misconfiguring, misrendering, or misrouting data, or whether Hermes Agent itself is returning/storing the wrong behavior.
```

That gives future agents the routing rule we need without importing a stale upstream architecture proposal into the Desktop repo.

## Suggested path

Please narrow this PR to maintainable Hermes Desktop AGENTS guidance, remove the smolagents/upstream runtime proposal from this branch, correct the API names and IPC guidance, and avoid brittle generated/file-size/version details. With those changes, this could be a useful contribution.

PR #197, opened 2026-05-16
fix(sessions): Repair session cache sync and search

leedusty91-prog:
The session cache synchronization logic was flawed, leading to stale data in the UI and causing search and session loading to fail.

1.  The sync query was based on `started_at`, which missed updates to existing sessions (e.g., new messages). This is now changed to `updated_at` to correctly fetch all modified sessions.

2.  The merging logic could lead to duplicate entries and failed to properly update existing cached sessions. The implementation now uses a Map-based approach to robustly upsert session data, preventing duplicates and ensuring data consistency.

This fixes the user-reported issues of session search returning no results and being unable to open historical sessions.

### DIFF
--- a/src/main/session-cache.ts
+++ b/src/main/session-cache.ts
@@ -88,11 +88,11 @@ export function syncSessionCache(): CachedSession[] {
     const rows = db
       .prepare(
         `SELECT s.id, s.started_at, s.source, s.message_count, s.model, s.title
-         FROM sessions s
-         WHERE s.started_at > ?
-         ORDER BY s.started_at DESC`,
+         FROM sessions s 
+         WHERE s.updated_at > ?
+         ORDER BY s.updated_at DESC`,
       )
-      .all(cache.lastSync > 0 ? cache.lastSync - 300 : 0) as Array<{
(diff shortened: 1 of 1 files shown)


fathah:
conflict

pmos69:
Thanks for the fix and for calling out the session-cache sync problem. The underlying issue was real, but this branch has since been superseded by later work and should not merge as-is.

The useful pieces are now covered elsewhere:

- Duplicate prevention / Map-based merge is already present on current `main` in `src/main/session-cache.ts`.
- Existing-session `messageCount` refresh for old sessions was fixed in #252.
- Existing-session `title` and `model` refresh was fixed in #378.
- The remaining search/clear stale-UI and empty-stale-cache recovery cases are covered by #439, with focused tests plus live/manual verification.

There is also one important implementation problem here: this PR changes the incremental query from `started_at` to `updated_at`, but the current Desktop session-cache code/tests do not have a `sessions.updated_at` column. So this branch would likely fail at runtime with `no such column: s.updated_at` against the schema Desktop is using.

Closing as superseded by #252, #378, and #439. Appreciate the contribution — it pointed at a real class of bugs, just no longer needs this branch to land.

PR #202, opened 2026-05-16
Fix Sessions loading hang on Intel macOS

bobbyunknown:
# Fix Sessions loading hang on Intel macOS

## Summary

Fixes a Sessions screen hang on Intel macOS when the packaged `better-sqlite3` native module has the wrong architecture.

## Problem

On an Intel macOS machine, the app bundle was `x86_64`, but the packaged native module was `arm64`:

```txt
Hermes Agent.app: Mach-O 64-bit executable x86_64
better_sqlite3.node: Mach-O 64-bit bundle arm64
```

Electron failed to load the module:

```txt
mach-o file, but is an incompatible architecture (have 'arm64', need 'x86_64')
```

That made `syncSessionCache()` fail. The renderer did not catch the rejected IPC call, so the Sessions page stayed in the loading state forever.

## Changes

- Catch `syncSessionCache()` failures in the main process and fall back to cached sessions.
- Catch session loading errors in the renderer and always clear the loading state in `finally`.
- Rebuild macOS native dependencies for the target release architecture in CI.
- Verify the packaged `better_sqlite3.node` architecture during macOS release builds.

## Testing

Installed dependencies successfully:

```bash
npm ci
```

Ran project checks:

```bash
npm run lint && npm run typecheck
```

Current result: the command fails due to existing lint/typecheck issues elsewhere in the repo, not from this patch. The edited code follows the existing patterns and the runtime fix was verified locally by replacing the wrong-arch `better_sqlite3.node` with an x86_64 build and confirming no native load errors.

image


fathah:
conflict

bobbyunknown:
> conflict

Conflict resolved. Ready for review.

pmos69:
Adding downstream evidence: a fresh user report came in today on #348 — Intel macOS, v0.5.0 packaged build, Sessions tab stuck on permanent loading spinner. Independent diagnosis from the reporter (after some back-and-forth) landed on exactly the failure mode this PR documents: arm64 `better_sqlite3.node` inside the x86_64 bundle → native module fails to load → IPC rejects → renderer has no `try/finally` so the loading flag never clears.

So the bug is now confirmed in the wild on a real v0.5.0 install, not just bobbyunknown's original repro. The fix shape in this PR (CI rebuild per target arch + arch-verification step + renderer-side `try/finally`) is the correct end-to-end remediation.

Nothing else to add — just wanted to put the corroboration on the record so it's easier to triage when this gets a review. Thanks @bobbyunknown for the thorough writeup back on May 16.

bionicbutterfly13:
That issue was what brought me here just now.

PR #231, opened 2026-05-17
fix: persist setup model selection reliably

hege009:
## Summary

Fixes desktop setup persistence so model configuration and locale survive app restarts.

This PR addresses cases where Hermes Desktop could repeatedly show setup/onboarding or lose selected settings because setup choices were not reliably written to the persistent config files.

Changes include:

- Persist desktop locale to `${HERMES_HOME}/desktop.json`
- Restore saved locale on app startup
- Ensure setup model config creates/updates `${HERMES_HOME}/config.yaml`
- Read and write model settings specifically under the `model:` YAML section
- Persist:
  - `model.provider`
  - `model.default`
  - `model.base_url`
- Avoid overwriting an existing default model with an empty value
- Keep writing `config.yaml` even when the live `/api/model/set` call succeeds
- Allow setup to persist an explicit model name for hosted providers
- Add regression tests for locale and model config persistence

## Why

Previously, `setModelConfig()` returned early if `config.yaml` did not exist. That meant first-run setup could save an API key but fail to persist the selected provider/model, causing the app to behave like setup was incomplete on later launches.

Locale selection was also only kept in memory, so language changes were lost after restart.

In addition, live model updates through the Hermes API could make the running gateway update successfully while leaving `config.yaml` unchanged, so the selection was lost after restart.

## Verification

Tested locally with `HERMES_HOME=D:\Programs\hermes`.

Ran:

```powershell
npm run typecheck:node
npm run typecheck:web
npx vitest run tests/model-config.test.ts tests/locale-persistence.test.ts tests/ipc-handlers.test.ts src/shared/i18n/index.test.ts src/renderer/src/components/I18nProvider.test.tsx

### DIFF
--- a/src/main/config.ts
+++ b/src/main/config.ts
@@ -44,6 +44,116 @@ function writeDesktopConfig(data: Record<string, unknown>): void {
   writeFileSync(desktopConfigFile(), JSON.stringify(data, null, 2), "utf-8");
 }
 
+function getConfigSection(content: string, section: string): string {
+  const lines = content.split("\n");
+  const start = lines.findIndex((line) =>
+    new RegExp(`^${escapeRegex(section)}:\\s*$`).test(line),
+  );
+  if (start === -1) return "";
+
+  const body: string[] = [];
+  for (let i = start + 1; i < lines.length; i++) {
+    const line = lines[i];
+    if (/^[A-Za-z_][\w-]*:/.test(line)) break;
+    body.push(line);
+  }
+  return body.join("\n");
+}
+
+function readYamlValue(content: string, key: string, section?: string): string | null {
+  const source = section ? getConfigSection(content, section) : content;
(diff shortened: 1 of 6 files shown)


fathah:
Conflict

pmos69:
Thanks for putting this together. I’m going to close this as superseded now: the two main fixes here have since landed in narrower follow-up PRs.

- Locale persistence is covered by #211, with current persistence in src/main/locale.ts and regression coverage in 	ests/locale-persistence.test.ts.
- Setup/model config persistence for missing config.yaml is covered by #246, with stronger scoped model: block handling in src/main/config.ts and regression coverage in 	ests/config-model-block.test.ts.

So the original problem is addressed on current main, and this branch is now conflicting. Really appreciate the contribution — this was pointing at a real issue, it just got resolved through later smaller PRs.

PR #239, opened 2026-05-18
fix(gateway): treat EPERM as alive and verify python image on Windows

pmos69:
## Summary

`process.kill(pid, 0)` is the POSIX-idiomatic existence check, but on Windows libuv requests `PROCESS_TERMINATE` access to issue the call. The gateway runs detached — whether the desktop spawned it (`spawn` with `detached: true`) or the Python CLI started it from a separate console (the CLI's daemon mode re-execs itself as `pythonw.exe` so the gateway runs windowless in the background). Either way, the detachment is enough that `OpenProcess` from an unrelated Electron main process refuses the `PROCESS_TERMINATE` handle — even though both run as the same user — and `process.kill` surfaces the failure as `EPERM`. The previous catch-all `try { ... } catch { return false }` conflated `EPERM` with `ESRCH`, so the desktop reported the gateway as **Stopped** while it was actually running.

## Changes

- Distinguish error codes: `ESRCH` means dead; anything else (notably `EPERM` on Windows) means alive. New `pidIsAlive(pid)` helper in `utils.ts`.
- Add a defensive image-name verifier `pidIsAliveAs(pid, prefixes)` that calls `tasklist` to confirm the live PID is in fact a python process — `python.exe` when the desktop spawned the gateway, `pythonw.exe` when it was started via the CLI's daemon mode. Guards against:
  - PID-reuse false positives on native Windows (PID recycled to an unrelated process after gateway crashed).
  - WSL coexistence false positives: when `HERMES_HOME` points into WSL via UNC mount and the PID file contains a Linux PID that happens to collide with a Windows PID.
- If `tasklist` is unavailable or times out, fall back to trusting the liveness check rather than blocking a healthy gateway on a flaky verifier.
- Fix `profiles.ts:isGatewayRunning` parsing only bare-integer PID files. The current Python CLI writes JSON (`{"pid": N, "kind": ..., "argv": [...]}`); the old `parseInt` silently failed on that, so per-profile gateway status was always **Stopped** regardless of actual state.

## Files

- `src/main/utils.ts` — new `pidIsAlive`, `pidIsAliveAs`, `getProcessImageNameWin` helpers.
- `src/main/hermes.ts` — `isGatewayRunning()` uses `pidIsAliveAs` with `[python, pythonw]` prefixes.
- `src/main/profiles.ts` — per-profile `isGatewayRunning()` handles JSON PID format and uses the shared verifier.

## Test plan

- [x] `npm run typecheck` passes (node + web).
- [x] `npm test` — all 317 tests pass across 22 files.
- [x] Verified against a real running gateway (PID owned by `pythonw.exe` started by `hermes gateway run`): old code reported Stopped, new code reports Running.
- [ ] macOS/Linux: `pidIsAliveAs` short-circuits to `pidIsAlive` (image check is Windows-only), so POSIX behavior is unchanged.
- [ ] Manual: kill the gateway and confirm UI flips to Stopped within the next 10s poll.

### DIFF
--- a/src/main/hermes.ts
+++ b/src/main/hermes.ts
@@ -18,7 +18,7 @@ import {
   isSshTunnelHealthy,
   startSshTunnel,
 } from "./ssh-tunnel";
-import { stripAnsi } from "./utils";
+import { pidIsAliveAs, stripAnsi } from "./utils";
 import { readModels } from "./models";
 import { HIDDEN_SUBPROCESS_OPTIONS } from "./process-options";
 import {
@@ -952,16 +952,16 @@ export function stopGateway(force = false): void {
   apiServerAvailable = false;
 }
 
+// Python image prefixes covering both native Windows (pythonw.exe / python.exe)
+// and POSIX (python, python3, pythonw). Used to verify the PID we read from
+// gateway.pid actually belongs to a python process before reporting alive.
+const GATEWAY_IMAGE_PREFIXES = ["python", "pythonw"];
+
 export function isGatewayRunning(): boolean {
   if (gatewayProcess && !gatewayProcess.killed) return true;
   const pid = readPidFile();
   if (!pid) return false;
-  try {
-    process.kill(pid, 0);
-    return true;
-  } catch {
-    return false;
-  }
+  return pidIsAliveAs(pid, GATEWAY_IMAGE_PREFIXES);
 }
 
 export function isApiReady(): boolean {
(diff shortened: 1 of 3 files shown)


fathah:
conflict

pmos69:
Rebased on main — conflict was a trivial import-formatting clash on `hermes.ts`. Now MERGEABLE / CLEAN. Typecheck and full test suite (403/403) pass.

PR #241, opened 2026-05-18
fix: recognize oauth profiles and authenticate desktop chat to local api server

### LINKED ISSUE #30: local connecting to Hermes Agent API: Error: Invalid API key

### LINKED ISSUE #41: Unable to chat with Hermes on Desktop - get 'error: invalid API key'

### LINKED ISSUE #67: Hermes Desktop: “Not logged into Nous Portal” error even though `hermes model` + CLI chat work
Hey folks, I’m having a persistent auth/session issue with Hermes Desktop (Electron wrapper) vs the Hermes Agent CLI.

Setup:
Hermes Agent v0.12.0
macOS (Apple Silicon)
Provider: Nous Portal (https://inference-api.nousresearch.com/v1)
Profile: scribe
Models tested: arcee-ai/trinity-large-thinking and llama-3.3-70b-versatile
What works
CLI works reliably:
HERMES_PROFILE=scribe hermes model → “Login successful! Auth state: ~/.hermes/profiles/scribe/auth.json”
HERMES_PROFILE=scribe hermes chat → responds normally
What fails
Hermes Desktop sometimes returns:
Error: Internal server error: Hermes is not logged into Nous Portal. Run hermes model to re-authenticate.
Seems like Desktop is reading different state than the CLI.
I also see Desktop state directory at:
~/Library/Application Support/hermes-desktop
Question
What’s the recommended way to ensure Hermes Desktop uses the same auth/home directory as the CLI (~/.hermes) and doesn’t get stuck in a “not logged in” loop?
Happy to provide logs — tell me which file(s) you want (e.g., errors.log, etc.).

### LINKED ISSUE #126: Local API auth broken + Office fails to start — basic functionality is unusable
I've been trying to use Hermes Desktop for a few days and keep hitting blocking bugs that make the app essentially unusable. Three issues, all already reported by others but still open/unfixed:

## 1. "Error: Invalid API key" when chatting in Local mode

**Setup:**
- macOS, Hermes Agent running locally on `http://127.0.0.1:8642`
- `API_SERVER_KEY` is set in Hermes config
- Connection mode: Local

**What happens:**
Every chat request fails with `Error: Invalid API key`. The desktop app never sends the `Authorization: Bearer <API_SERVER_KEY>` header to the local API server.

**Expected:** Desktop should read `API_SERVER_KEY` from the active Hermes profile and authenticate local requests.

**Related:** #30, #41. PR #10 already has a fix for this but has not been merged for months.

## 2. Chat session spins forever on API error

When the above auth error (or any API failure) happens, the chat UI shows an infinite loading spinner instead of surfacing the error. There's no timeout, no retry, no error message. Users have to kill the tab or restart the app.

This is a basic error-handling gap that should not exist in a released desktop app.

## 3. Office (Claw3D) tab is blank / fails to start

**Setup:** Same as above, Local mode.

**What happens:**
Clicking the Office tab opens a blank/white page. The embedded `hermes-office` service never becomes ready.

**Related:** PR #7 fixes startup readiness checks so the webview waits for `/office` to be reachable before loading. PR #36 adds a user-visible message for Remote mode. Both PRs are unmerged.

## Overall impression

These are not edge cases — they are the core login → chat → Office workflows. With multiple PRs already written and sitting unreviewed, it feels like the project is not being actively maintained.

As a user, I shouldn't need to:
- Disable `API_SERVER_KEY` to make local chat work
- Manually refresh Office to get it to render
- Dig through open PRs to find patches that fix obvious bugs

Please consider prioritizing review and merge of #10, #7, and #36. Otherwise the Desktop client is not a viable way to use Hermes Agent.

### LINKED ISSUE #185: [Error]: Invalid API key - Model works
It shows invalid API key, but in hermes terminal chat, it works normally. Does anyone know how to deal with it? Thanks

Image

pmos69:
## Summary

Supersedes #10 (original PR has been inactive for >30 days, in CONFLICTING state and not mergeable as-is). Two independent fixes from that PR, rebased onto current `main` with the reviewer's outstanding feedback addressed. Original work by `Hermes Agent <hermes@local.invalid>` (credited via `Co-authored-by`).

Both commits left intact as separate units so they can land or be revisited independently.

## Commit 1 — `fix: detect oauth-backed active profiles in install gate`

Profiles whose only credential is an OAuth grant in `auth.json` (e.g. `openai-codex` under a named profile) were reported by `checkInstallStatus` as un-configured, so the desktop showed the install/setup screen instead of letting the user proceed. The check was also default-profile only — named profiles never had their own `auth.json` or `.env` consulted.

- `checkInstallStatus` now reads the active profile from `~/.hermes/active_profile` and probes that profile's `.env` and `auth.json`.
- New `hasOAuthCredentials()` in `config.ts` recognizes a credential as configured only when the entry contains at least one of `access_token` / `refresh_token` / `api_key` (under either `providers[<name>]` or `credential_pool[<name>]`). The previous looser installer-side check accepted empty `providers: { x: {} }` or a bare `active_provider`, which masked real onboarding errors.

## Commit 2 — `fix: send api server auth for desktop chat`

The local hermes gateway can require a shared secret (`api_server.token` in `config.yaml` / `API_SERVER_KEY` in `.env`). When set, the desktop's chat must include it as `Authorization: Bearer …` on every chat request — without it the gateway returns *Invalid API key* and chat fails for users who otherwise have a working local setup.

- New `getApiServerKey()` resolves the value per profile (profile `config.yaml` → root `config.yaml` → profile `.env` → root `.env`).
- `sendMessageViaApi` attaches the Bearer header **only in local mode** — in remote/SSH mode the existing `getRemoteAuthHeader()` already populates `Authorization` and must not be overwritten.
- The error-probe request now reuses the same headers object so it carries identical auth (previously the probe built a separate header dict that omitted the api server key).

## Reviewer feedback from #10 addressed

1. **Shared `getActiveProfileNameSync` helper.** Lives once in `utils.ts`; `profiles.ts`'s async wrapper delegates to it. No more three near-identical copies across `installer.ts` / `config.ts` / `profiles.ts`.
2. **`InstallStatus.activeProfile` on the interface itself.** Both `src/main/installer.ts` and `src/preload/index.d.ts` carry the optional field, so renderers see it through TypeScript instead of via an `& { activeProfile? }` intersection at the call site.
3. **Credential-pool IPC threads `profile`.** `getCredentialPool` / `setCredentialPool` in the preload, main IPC handlers, and `HermesAPI` type all accept an optional `profile`. Renderers can target an explicit profile; omitting it falls back to the currently active profile (consistent with the renderer's existing default behavior).
4. **`getApiServerKey` cached via the existing 5s TTL.** Routed through the same `_cache` mechanism `readEnv()` uses, with invalidation hooks in `setEnvValue` / `setConfigValue` when the key being written is `API_SERVER_KEY`.

Test file `tests/installer-utils.test.ts` updated to assert the new (stricter) `hasOAuthCredentials` semantics: empty `providers: { x: {} }` and bare `active_provider` no longer count as configured; `access_token`, `refresh_token`, or `api_key` all qualify.

## Test plan

- [x] `npm test` — all 317 tests across 22 files pass.
- [x] Verified locally on a native Windows install: OAuth-only profile (`openai-codex`) no longer triggers the install gate, and chat against the local API server with `API_SERVER_KEY` set succeeds (returns the expected response instead of "Invalid API key").
- [ ] Toggle into remote mode and confirm the local Bearer header doesn't overwrite the remote auth header (covered by the `!isRemoteMode()` gate, but worth a manual sanity check).
- [ ] Settings UI credential pool on a named profile — confirm it reads/writes that profile's `auth.json` rather than the root one.

Closes #30.

Closes #41.

Closes #67.

Closes #126.

Closes #185.

### DIFF
--- a/src/main/config.ts
+++ b/src/main/config.ts
@@ -177,6 +177,7 @@ export function setEnvValue(
 
   const { envFile } = profilePaths(profile);
   invalidateCache(`env:${profile || "default"}`);
+  if (key === "API_SERVER_KEY") invalidateCache("apiServerKey:");
 
   if (!existsSync(envFile)) {
     safeWriteFile(envFile, `${key}=${value}\n`);
@@ -233,6 +234,7 @@ export function setConfigValue(
   value: string,
   profile?: string,
 ): void {
+  if (key === "API_SERVER_KEY") invalidateCache("apiServerKey:");
   const { configFile } = profilePaths(profile);
   if (!existsSync(configFile)) return;
 
@@ -486,6 +488,45 @@ export function setModelConfig(
 
 export function getHermesHome(profile?: string): string {
   return profilePaths(profile).home;
+}
+
+/**
+ * Resolve the API server's shared secret. Honoured by the local hermes
+ * gateway (api_server.token in config.yaml / API_SERVER_KEY in .env) when
+ * present; the desktop must include it as `Authorization: Bearer …` on
+ * every chat request, otherwise the gateway responds with "Invalid API
+ * key".
+ *
+ * Search order: profile's config.yaml → default config.yaml → profile's
+ * .env → default .env. Returns "" when none configured.
+ *
+ * Hot path: called per chat message and per error-probe. Reuse the same
+ * 5s TTL cache as readEnv() so we don't re-parse config.yaml + .env
+ * every call. Invalidated by setEnvValue / setConfigValue when the key
+ * being written is API_SERVER_KEY.
+ */
+export function getApiServerKey(profile?: string): string {
+  const cacheKey = `apiServerKey:${profile || "default"}`;
+  const cached = getCached<string>(cacheKey);
+  if (cached !== undefined) return cached;
+
+  const candidates = [
+    getConfigValue("API_SERVER_KEY", profile),
+    profile && profile !== "default" ? getConfigValue("API_SERVER_KEY") : null,
+    readEnv(profile).API_SERVER_KEY || null,
+    profile && profile !== "default" ? readEnv().API_SERVER_KEY || null : null,
+  ];
+
+  let value = "";
+  for (const candidate of candidates) {
+    const trimmed = String(candidate || "").trim();
+    if (trimmed) {
+      value = trimmed;
+      break;
+    }
+  }
+  setCache(cacheKey, value);
+  return value;
 }
 
 // ── Platform enabled/disabled ─────────────────────────────
--- a/src/main/config.ts
+++ b/src/main/config.ts
@@ -1,7 +1,13 @@
 import { readFileSync, writeFileSync, existsSync, mkdirSync } from "fs";
 import { join } from "path";
 import { HERMES_HOME } from "./installer";
-import { profilePaths, escapeRegex, safeWriteFile } from "./utils";
+import {
+  escapeRegex,
+  getActiveProfileNameSync,
+  profileHome,
+  profilePaths,
+  safeWriteFile,
+} from "./utils";
 import { getYamlPath } from "./yaml-path";
 
 // ── Connection Config (local / remote / ssh) ─────────────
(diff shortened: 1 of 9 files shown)


fathah:
Conflict

pmos69:
Rebased — conflict was against the recently-merged #239 (PID liveness helpers in utils.ts), #246 (model: block scoping in config.ts), and #243 (nested-config-yaml reader). Resolved by keeping both sides of all three additions; the OAuth install-gate check, the API_SERVER_KEY forwarding, and the new helpers now coexist cleanly. Typecheck and 435/435 tests pass.

macdesire:
I believe this may be related to PR #241.
The /office HTML loads fully over HTTP, including all _next/static/... references, but Electron/WebView later fails with TLS errors while loading JS/CSS/font assets and the runtime stays disconnected.

Direct API access to http://127.0.0.1:8642/v1/chat/completions works correctly via curl, so the Hermes gateway itself appears healthy.

This looks more like a frontend runtime / secure-context / websocket / Electron networking issue than a backend failure.

PR #244, opened 2026-05-18
fix: persist locale selection across app restarts

493Arceus:
## Summary
Fixes the issue where users had to re-select their preferred language after every app launch.

## Problem
The app's language setting (`locale`) was only stored in memory. On restart, it always reset to the default (`\"en\"`), forcing users to reconfigure their language preference every time.

## Changes
- `src/main/config.ts`: Export `readDesktopConfig` / `writeDesktopConfig` so they can be used by the locale module
- `src/main/locale.ts`: `getAppLocale()` now reads saved locale from `~/.hermes/desktop.json`; `setAppLocale()` persists the locale to `desktop.json`
- `src/main/index.ts`: Restore the saved locale from `~/.hermes/desktop.json` on app startup

## Verification
- `npm run typecheck` passes
- `npm run test` passes (317 tests)
- `npm run build` succeeds

### DIFF
--- a/src/main/config.ts
+++ b/src/main/config.ts
@@ -27,7 +27,7 @@ function desktopConfigFile(): string {
   return join(HERMES_HOME, "desktop.json");
 }
 
-function readDesktopConfig(): Record<string, unknown> {
+export function readDesktopConfig(): Record<string, unknown> {
   try {
     const f = desktopConfigFile();
     if (!existsSync(f)) return {};
@@ -37,7 +37,7 @@ function readDesktopConfig(): Record<string, unknown> {
   }
 }
 
-function writeDesktopConfig(data: Record<string, unknown>): void {
+export function writeDesktopConfig(data: Record<string, unknown>): void {
   if (!existsSync(HERMES_HOME)) {
     mkdirSync(HERMES_HOME, { recursive: true });
   }
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -146,6 +146,8 @@ import {
   isAllowedWebviewUrl,
 } from "./security";
 import type { AppLocale } from "../shared/i18n/types";
+import { setLocale } from "../shared/i18n";
+import { readDesktopConfig } from "./config";
 import {
   sshListInstalledSkills,
   sshGetSkillContent,
@@ -1299,6 +1301,13 @@ function setupUpdater(): void {
 app.whenReady().then(() => {
   app.name = "Hermes";
   electronApp.setAppUserModelId("com.nousresearch.hermes");
+
+  // Restore directly via the shared setter; going through setAppLocale
+  // would needlessly re-write desktop.json with the value we just read.
+  const savedLocale = readDesktopConfig().locale as AppLocale | undefined;
+  if (savedLocale) {
+    setLocale(savedLocale);
+  }
 
   app.on("browser-window-created", (_, window) => {
     optimizer.watchWindowShortcuts(window);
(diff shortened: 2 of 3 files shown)


493Arceus:
#234

pmos69:
Nice fix — clean approach to making main self-sufficient for locale. Two small suggestions, neither blocking:

**1. `src/main/index.ts` boot-restore — explain why it goes around the wrapper.**

The restore at `app.whenReady` imports `setLocale` from `../shared/i18n` directly rather than going through `setAppLocale` in `locale.ts`. That's actually the right call — `setAppLocale` would re-write `desktop.json` with the value you just read — but the asymmetry is non-obvious. A one-line comment would help a future reader:

```ts
// Restore directly via the shared setter; going through setAppLocale
// would needlessly re-write desktop.json with the value we just read.
const savedLocale = readDesktopConfig().locale as AppLocale | undefined;
if (savedLocale) {
  setLocale(savedLocale);
}
```

**2. `src/main/locale.ts` — validate the cast.**

`readDesktopConfig().locale as AppLocale | undefined` lets through any value in the JSON. If `desktop.json` ever carries a stale or hand-edited value outside `APP_LOCALES`, the bad value propagates into the i18n system. A small guard hardens it without adding ceremony:

```ts
import { APP_LOCALES /* … */ } from "../shared/i18n";

export function getAppLocale(): AppLocale {
  const saved = readDesktopConfig().locale;
  if (
    typeof saved === "string" &&
    (APP_LOCALES as readonly string[]).includes(saved)
  ) {
    return saved as AppLocale;
  }
  return getSharedLocale() || DEFAULT_ACTIVE_LOCALE;
}
```

Otherwise looks correct — closes #234 cleanly.

493Arceus:
Thanks for the thorough review, @pmos69! Both excellent points — I've applied both suggestions in the latest push:

1. Added the comment explaining why we restore via `setLocale` directly instead of going through `setAppLocale`.
2. Replaced the `as AppLocale | undefined` cast with a proper `APP_LOCALES.includes()` guard in `getAppLocale()`.

Much cleaner and safer this way.

fathah:
Conflict

pmos69:
Heads-up @493Arceus: the conflict source is the recently-merged [#243](https://github.com/fathah/hermes-desktop/pull/243) (andreab67's nested-config-yaml-reader), which already exported `readDesktopConfig` / `writeDesktopConfig` from `config.ts` — same two lines this PR also exports. The auto-merge sees "both sides modified the same line" even though both are adding the identical `export` keyword.

Quick rebase:

```
git fetch origin main
git rebase origin/main
# resolve config.ts — accept either side; they're literally the same
git add src/main/config.ts
git rebase --continue
git push --force-with-lease
```

Worth checking after rebase: `src/main/index.ts` may also have minor import-order drift from the other PRs merged today (#237 / #238 / #239 / #246). Same pattern — both sides are valid additions, just need to keep both. `npm run typecheck` will confirm the resolution compiles.

493Arceus:
Closing this PR as the locale persistence fix has already been superseded by a more comprehensive implementation recently merged into main (see the updated `src/main/locale.ts` with `readSavedLocale`/`writeSavedLocale`/`isAppLocale` and module-level auto-restore).

Thanks @pmos69 for the thorough review — the suggestions around validation guards and avoiding redundant desktop.json writes were incorporated into the final merged approach.

pmos69:
Thanks @493Arceus — and to be precise, the patterns that ended up in `main` (`isAppLocale` type guard, `readSavedLocale` for restore-without-rewrite, module-level auto-restore) came in via [#211](https://github.com/fathah/hermes-desktop/pull/211) from @ASTairov, which merged first. Nice convergence on the same shape from independent work — credit goes to them for the version that landed.

Issue #242, opened 2026-05-18
tries to pick personality as model


drishal:
<img width="1633" height="495" alt="Image" src="https://github.com/user-attachments/assets/80b11e85-2a18-492c-803d-21b0443c8d82" />

in this screenshot for example here you can see thats picking up config from personalities instead of actually picking up the right model by default 
and not only that it does not show any of the latest gpt models 
my config snippet 
```
 personalities:
    coding: You are an expert coding assistant. Write clean, efficient, well-commented
      code. Follow modern best practices, suggest optimizations, explain architecture
      decisions, and help with debugging or refactoring. Support multiple languages
      and frameworks.
    concise: You are a concise professional assistant. Keep responses brief, direct,
      and to the point while remaining complete and accurate. Eliminate redundant
      words.
    default: You give clear, accurate, and actionable responses. Prioritize precision, logical reasoning,
      and practical solutions. Use professional language, avoid slang, emojis, or
      unnecessary flair. Structure your answers for readability when appropriate (e.g.,
      bullet points, numbered steps, or sections).   <---it tries to set this as default model instead
```

PR #248, opened 2026-05-18
fix(config): support dotted YAML paths in getConfig / setConfig

### LINKED ISSUE #247: getConfig/setConfig silently fail on dotted paths and leak across YAML blocks
## Summary

Two bugs in `getConfigValue` / `setConfigValue` ([src/main/config.ts:188-220](https://github.com/fathah/hermes-desktop/blob/main/src/main/config.ts#L188)) — same root cause as #242, in a different code path:

1. **Dotted-path keys are silently no-ops.** The helpers treat the key string as a literal regex fragment, so `getConfig('agent.service_tier')` looks for a *literal* `agent.service_tier:` line — which never exists in real YAML, where the value lives nested as `agent:` → `service_tier:`. The renderer relies on dotted paths in several places; all of them silently fail.
2. **Flat keys leak across blocks.** A flat key like `service_tier` matches the first `service_tier:` line at *any* indentation, so if the file has both `agent.service_tier` and `telegram.service_tier`, reads pick whichever comes first in document order and writes overwrite the wrong one.

Renderer call sites that are currently broken because they pass dotted paths:
- `src/renderer/src/screens/Chat/hooks/useFastMode.ts:18,26` — `agent.service_tier` (Fast Mode toggle)
- `src/renderer/src/screens/Memory/Memory.tsx:115,561,581` — `memory.provider` (Memory provider selection)
- `src/renderer/src/screens/Settings/Settings.tsx:130,133,792,818` — `network.force_ipv4`, `network.proxy`
- `src/renderer/src/screens/Chat/hooks/useLocalCommands.ts:139`

The same issue affects the SSH-mode mirrors `sshGetConfigValue` / `sshSetConfigValue` ([src/main/ssh-remote.ts:577-605](https://github.com/fathah/hermes-desktop/blob/main/src/main/ssh-remote.ts#L577)) — identical regex pattern.

## Reproducer

Tested against pristine `main` (commit `3b757dd`) with a temporary vitest file:

```ts
writeFileSync(configFile, [
  "agent:",
  "  service_tier: fast",
].join("\n"));

const v = getConfigValue("agent.service_tier");  // → null
setConfigValue("agent.service_tier", "PROBE");   // → file unchanged
```

```ts
writeFileSync(configFile, [
  "telegram:",
  "  service_tier: 'oops-not-agent'",
  "agent:",
  "  service_tier: fast",
].join("\n"));

const v = getConfigValue("service_tier");        // → "oops-not-agent"
setConfigValue("service_tier", "PROBE");
// → telegram.service_tier becomes "PROBE", agent.service_tier untouched
```

## Impact

- **Fast Mode toggle (Chat header):** clicks have no persistent effect; the next session reverts.
- **Memory provider selector:** reads return `null` so the UI can't show the current provider; writes silently fail so the user thinks they switched but the gateway still uses the old one.
- **Settings → network options (`force_ipv4`, `proxy`):** same — invisible to the UI, invisible to writes.
- All four call sites work *by accident* when the user happens to have the key uniquely present at a level the regex stumbles into, but they don't work in general.

## Related

- #242 — same regex-scoping bug pattern in `getModelConfig` / `setModelConfig`. PR #246 fixes the model path with a block-aware reader (`readTopLevelBlock`). This issue is the symmetric problem in the generic config helpers; the fix is a small extension of the same approach.

## Fix plan

Replace the loose regex with a real YAML-path navigator that:

1. Splits the key on `.` into segments.
2. For each segment, finds the matching `key:` line at strictly-greater indentation than its parent.
3. Returns the value (or, for writes, the substring range to splice).

Handles arbitrary nesting depth, treats flat keys as 1-segment paths (so a renderer call with a single-segment key like `provider` only matches top-level `provider:`, not `model.provider`). I'll propose this in a follow-up PR with regression tests covering each of the call sites above.

I have a candidate fix and will open a PR shortly.

pmos69:
## Problem

Same shape of bug as #242, in `getConfigValue` / `setConfigValue` ([src/main/config.ts:181-213](https://github.com/fathah/hermes-desktop/blob/main/src/main/config.ts#L181)) and their SSH mirrors ([ssh-remote.ts:577-605](https://github.com/fathah/hermes-desktop/blob/main/src/main/ssh-remote.ts#L577)).

The helpers ran a loose `^\s*<key>:` regex against the whole config.yaml. Two consequences:

1. **Dotted paths silently no-op.** The key string was used as a literal regex fragment, so `getConfig('agent.service_tier')` looked for a flat line spelled exactly `agent.service_tier:` — which never exists in real YAML. Reads returned `null` and writes wrote nothing.
2. **Flat keys leaked across blocks.** A flat `service_tier` matched the first occurrence at any indent, so a value under `telegram:` could shadow `agent.service_tier`, and `setConfig('service_tier', 'X')` would overwrite whichever sibling block happened to appear first.

Renderer call sites currently silently broken (all use dotted paths):
- `src/renderer/src/screens/Chat/hooks/useFastMode.ts` — `agent.service_tier` (Fast Mode toggle)
- `src/renderer/src/screens/Memory/Memory.tsx` — `memory.provider` (Memory provider selector)
- `src/renderer/src/screens/Settings/Settings.tsx` — `network.force_ipv4`, `network.proxy`
- `src/renderer/src/screens/Chat/hooks/useLocalCommands.ts`

Issue #247 has the full repro.

## Fix

Replace the loose regex with a small dotted-path navigator:

- Split the key on `.` into segments.
- For each segment, find the next matching line at strictly-greater indent than its parent's. The first non-blank line inside a block defines the "direct child" indent; grandchildren are walked past without being treated as siblings of the segment.
- Single-segment keys are pinned to column 0 — `getConfig('provider')` only matches a top-level `provider:`, not `model.provider` or `auxiliary.vision.provider` by accident.
- Writes splice in place, preserving the line's surrounding whitespace and any trailing comment.
- Top-level missing keys are appended; missing nested paths are a no-op rather than guessing where to materialize a parent (avoids corrupting the file).

## Tests

New file `tests/config-value-paths.test.ts` covers 17 cases:

**`getConfigValue` — dotted paths:**
- ✅ 2-segment path (`agent.service_tier`)
- ✅ 2-segment with quoted value (`memory.provider`)
- ✅ 3-segment path (`agent.personalities.helpful`) — arbitrary nesting
- ✅ Missing parent block returns `null`
- ✅ Missing leaf key under existing block returns `null`
- ✅ Top-level key with same name as the leaf doesn't satisfy a dotted path
- ✅ Grandchildren ignored — `agent.service_tier` matches only direct child

**`getConfigValue` — flat keys pinned to top level:**
- ✅ Reads a true top-level key
- ✅ Does NOT match a nested occurrence when called flat
- ✅ Does NOT pick first nested sibling across blocks

**`setConfigValue` — dotted paths:**
- ✅ Updates in place, leaves siblings/grandparents untouched
- ✅ Does NOT write a sibling block's same-named key
- ✅ Missing nested path is a no-op (don't corrupt the file)
- ✅ Cache invalidation across write→read

**`setConfigValue` — flat keys:**
- ✅ Updates top-level key in place
- ✅ Appends new top-level key when missing
- ✅ Flat key does NOT overwrite a same-named nested key (top-level append instead)

## Behavior changes to be aware of

- **Flat keys now require column 0.** Any existing caller passing a single-segment key like `service_tier` and *relying* on it accidentally matching a nested occurrence will now get `null` / no-op. Searched the renderer for flat calls — none exist; all renderer calls use dotted paths. Internal callers (`setConfigValue('API_SERVER_KEY', …)`) already target top-level keys. So this is a strict bug fix in practice.
- **`setConfigValue` for missing nested paths is a no-op** rather than wedging in a malformed line. The renderer screens that drive these calls all read first to render the current state, so a no-op on a missing path simply means the UI shows "not set" — the same outcome as before the fix, just without writing corrupt YAML.

## Test plan

- [x] `npm test` — 23 files / **334 tests** pass (317 original + 17 new).
- [ ] Manual: toggle Fast Mode in the Chat header; restart the app and confirm the setting persists. Same for Memory provider and Settings → network options.

## Follow-up not in this PR

- The SSH-mode mirrors `sshGetConfigValue` / `sshSetConfigValue` in `ssh-remote.ts` carry the identical regex bug. Same fix mechanically applies, but it'd be a separate PR scoped to the remote path; this one is local-mode only.
- This is the third regex-over-YAML bug in `config.ts` (#242 / model fields, #247 / generic helpers, and the SSH mirror). At some point the project may want to drop in `yaml` or `js-yaml`. Not in scope here.

Closes #247. Related: #242, PR #246.

### DIFF
--- a/src/main/config.ts
+++ b/src/main/config.ts
@@ -209,6 +209,187 @@ export function validateEnvEntry(key: string, value: string): void {
   }
 }
 
+function stripYamlQuotes(raw: string): string {
+  const trimmed = raw.trim();
+  if (trimmed.length >= 2) {
+    const first = trimmed[0];
+    const last = trimmed[trimmed.length - 1];
+    if ((first === '"' && last === '"') || (first === "'" && last === "'")) {
+      return trimmed.slice(1, -1);
+    }
+  }
+  return trimmed;
+}
+
+/**
+ * Locate a dotted YAML path in `content` (e.g. "agent.service_tier" finds
+ * the `service_tier` field nested under top-level `agent:`). Returns the
+ * value plus the substring offsets a writer can splice over, or null
+ * when any segment of the path is missing.
+ *
+ * Why this exists: the renderer passes dotted paths like
+ * `agent.service_tier`, `memory.provider`, `network.force_ipv4` through
+ * `getConfig`/`setConfig`. The old implementation used the key string as
+ * a literal regex fragment, so it looked for a flat line spelled exactly
+ * `agent.service_tier:` — which never exists in real YAML and silently
+ * returned null. Flat keys also leaked across blocks (a `service_tier`
+ * under `telegram:` could shadow `agent.service_tier`). See issue #247.
+ *
+ * Each segment must appear at strictly-greater indent than its parent's
+ * line. Segments without dots are treated as 1-segment paths and pinned
+ * to the top level (column-0 keys only) — so a flat `provider` no longer
+ * matches `model.provider` or `auxiliary.vision.provider` by accident.
+ *
+ * Returns the first match in document order at each level; later
+ * duplicates at the same level are ignored, matching YAML semantics for
+ * mappings.
+ */
+interface YamlPathHit {
+  value: string;
+  /** Absolute offset where the writer should splice the new value. */
+  valueStart: number;
+  /** Absolute offset just past the substring the writer should replace.
+   *  Excludes any trailing comment so we don't clobber `# notes`. */
+  valueEnd: number;
+}
+
+function findYamlPath(content: string, dottedPath: string): YamlPathHit | null {
+  const segments = dottedPath.split(".").filter(Boolean);
+  if (segments.length === 0) return null;
+
+  let cursor = 0;
+  let parentIndent = -1;
+
+  for (let i = 0; i < segments.length; i++) {
+    const segment = segments[i];
+    const isLast = i === segments.length - 1;
+    const found = findSegmentInBlock(content, cursor, parentIndent, segment);
+    if (!found) return null;
+
+    if (isLast) {
+      return {
+        value: stripYamlQuotes(found.rawValue),
+        valueStart: found.valueStart,
+        valueEnd: found.valueEnd,
+      };
+    }
+
+    // Descend: subsequent search continues after the segment's header
+    // line, bounded by indent > parentIndent.
+    cursor = found.afterLine;
+    parentIndent = found.indent;
+  }
+
+  return null;
+}
+
+interface SegmentMatch {
+  /** Indent length of the matched line. */
+  indent: number;
+  /** Raw value substring (between the colon's gap and any trailing comment). */
+  rawValue: string;
+  valueStart: number;
+  valueEnd: number;
+  /** Absolute offset of the byte just past the matched line's newline. */
+  afterLine: number;
+}
+
+function findSegmentInBlock(
+  content: string,
+  startAt: number,
+  parentIndent: number,
+  segment: string,
+): SegmentMatch | null {
+  // Walk lines from startAt until we leave the parent's block (a line
+  // with indent <= parentIndent). Within the block, return the first
+  // line whose key matches `segment` at the *minimum* indent > parent's
+  // — which is the depth of direct children.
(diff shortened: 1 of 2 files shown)


fathah:
Conflict

PR #254, opened 2026-05-18
feat(chat): show reasoning_content from thinking-mode models

### LINKED ISSUE #223: The "think" mode is missing the display of "think" content.
When utilizing the "thinking mode" model, the content of `reasoning_content` within the response is not displayed on the UI interface; however, the Hermes agent itself is capable of displaying this `reasoning_content`. We hope you will consider adding this information to the display.

使用思考模式的模型，响应中的resoning_content的内容不会在ui界面显示，Hermes agent本身是可以显示resoning_content的。希望考虑添加这个信息。

pmos69:
## Summary

Thinking-mode models (DeepSeek-R1, OpenAI o-series via proxies, Gemini thinking, etc.) stream a `delta.reasoning_content` field alongside `delta.content`. The Hermes CLI displays this; the Desktop's chat UI silently drops it. Reported in #223 — *"使用思考模式的模型，响应中的 resoning_content 的内容不会在 ui 界面显示，Hermes agent 本身是可以显示 resoning_content 的"* (when using thinking-mode models, the response's reasoning_content isn't shown in the UI; the Hermes agent itself can show it).

## What it looks like

Three rendering states (visual sanity check via Preview):

1. **Streaming** — reasoning is arriving, no answer yet → auto-expanded, label reads *"Thinking…"* with a subtle pulse animation.
2. **Answered, collapsed** — once content starts arriving, collapses to *"Reasoning ▸"*. The reasoning stays available; the user can re-expand any time.
3. **Answered, expanded** — body in monospace, gray, scrollable, gated to 320px max-height.

Layout uses existing chat color tokens / theme variables, so dark mode and skin variants inherit automatically.

## Implementation — three thin layers

| Layer | File | What |
|---|---|---|
| Parser | `src/main/sse-parser.ts`, `src/main/hermes.ts` | Read `delta.reasoning_content` (most providers) or `delta.reasoning` (OpenRouter shape) — either string is forwarded via a new optional `cb.onReasoning(text)`. Empty / non-string values are skipped. Reasoning chunks don't count as visible content, so the SSE `hasContent` state for the existing probe-on-empty-response error path stays accurate. |
| IPC + preload | `src/main/index.ts`, `src/preload/index.ts`, `src/preload/index.d.ts` | Main forwards `cb.onReasoning` to renderer via a new `chat-reasoning` channel; preload exposes `onChatReasoning(cb)` on the `HermesAPI` type. |
| Renderer | `src/renderer/src/screens/Chat/types.ts`, `…/hooks/useChatIPC.ts`, `…/MessageRow.tsx`, `…/assets/main.css` | `ChatMessage.reasoning?: string` field; `useChatIPC` appends incoming reasoning chunks to the current agent bubble (or creates the bubble if reasoning arrives before any answer, which it usually does); `MessageRow` renders the collapsible block with state per-row. |

Block-typed extended-thinking from Anthropic Claude is **not** covered in this PR — it streams as separate `thinking`-typed blocks rather than a string field on the chat-completion delta, which needs a different parser path. Flagged as a follow-up in the code comments.

## Tests — 13 new, suite goes 317 → 330

**`tests/sse-parser.test.ts` — 6 new cases:**
- ✅ `delta.reasoning_content` (most providers' shape) → `onReasoning`, not `onChunk`
- ✅ `delta.reasoning` (OpenRouter shape) → `onReasoning`
- ✅ Both reasoning and content in the same delta → both callbacks fire
- ✅ Empty reasoning fields are skipped
- ✅ Non-string (`null`) reasoning is skipped, content still routed
- ✅ Missing `onReasoning` callback → no throw, content unaffected

**`src/renderer/src/screens/Chat/MessageRow.test.tsx` (new file) — 7 cases:**
- ✅ No reasoning field → no block rendered
- ✅ Reasoning on a user message → ignored (defensive — reasoning is agent-only)
- ✅ Reasoning + content → collapsed by default, `aria-expanded="false"`, body not in DOM
- ✅ Click toggle → body appears, `aria-expanded="true"`
- ✅ Streaming state (`isLast && isLoading && !content`) → auto-expanded with "Thinking…" label
- ✅ Same state once content starts arriving → flips to collapsed "Reasoning" label
- ✅ Whitespace-only reasoning → no block rendered (treated as no reasoning)

## i18n

Added `chat.thinking` and `chat.reasoning` keys to **en** and **zh-CN** locales (since the issue reporter is Chinese-speaking). Other locales fall back to English via the existing fallback chain and are translatable in a follow-up.

## Test plan

- [x] `npm test` — 23 files / **330 tests** pass (317 origin + 13 new).
- [x] Visual sanity check on all three rendering states via standalone HTML preview using the actual CSS classes verbatim.
- [ ] End-to-end against a real thinking-mode endpoint (DeepSeek-R1 via OpenRouter or similar). Couldn't do this on the dev box without a thinking-model account — manual smoke check is recommended before merge.

Closes #223.

### DIFF
--- a/src/main/hermes.ts
+++ b/src/main/hermes.ts
@@ -152,6 +152,12 @@ platforms:
 
 export interface ChatCallbacks {
   onChunk: (text: string) => void;
+  /**
+   * Streaming "thinking" content from reasoning models — sourced from
+   * `delta.reasoning_content` (most providers) or `delta.reasoning`
+   * (OpenRouter and some others). Issue #223.
+   */
+  onReasoning?: (text: string) => void;
   onDone: (sessionId?: string) => void;
   onError: (error: string) => void;
   onToolProgress?: (tool: string) => void;
@@ -314,6 +320,20 @@ function sendMessageViaApi(
           rateLimitRemaining: parsed.usage.rate_limit_remaining,
           rateLimitReset: parsed.usage.rate_limit_reset,
         });
+      }
+
+      // Reasoning / thinking-model chain-of-thought streams alongside
+      // regular content. Different upstreams use different field names;
+      // accept both common ones (issue #223). The block-typed
+      // extended-thinking format from Anthropic is not covered here.
+      const reasoning =
+        typeof delta?.reasoning_content === "string"
+          ? delta.reasoning_content
+          : typeof delta?.reasoning === "string"
+            ? delta.reasoning
+            : "";
+      if (reasoning && cb.onReasoning) {
+        cb.onReasoning(reasoning);
       }
 
       if (delta?.content) {
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -608,6 +608,9 @@ function setupIPC(): void {
             fullResponse += chunk;
             event.sender.send("chat-chunk", chunk);
           },
+          onReasoning: (text) => {
+            event.sender.send("chat-reasoning", text);
+          },
           onDone: (sessionId) => {
             currentChatAbort = null;
             event.sender.send("chat-done", sessionId || "");
--- a/src/main/sse-parser.ts
+++ b/src/main/sse-parser.ts
@@ -13,6 +13,16 @@ export interface ParsedUsage {
 
 export interface SseCallbacks {
   onChunk: (text: string) => void;
+  /**
+   * Streaming "thinking" / chain-of-thought content from reasoning models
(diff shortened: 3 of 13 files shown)


pmos69:
Closing for now — discovered the upstream gateway (hermes-agent) doesn't forward delta.reasoning_content over /v1/chat/completions, so this desktop-side change has no visible effect with the bundled gateway. Will revisit after the gateway side is addressed.

PR #259, opened 2026-05-19
feat(models): add "Fetch Available Models" feature

Clare31:
Add ability to fetch model list directly from OpenAI-compatible API endpoints:
- New provider-models.ts with fetchAvailableModels IPC handler
- New modelCatalog.ts helper for provider→env key resolution
- Updated Models and Providers UI with fetch/select UX
- i18n strings in en and zh-CN
- Unit tests for provider-models module

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -87,6 +87,7 @@ import {
   updateSessionTitle,
 } from "./session-cache";
 import { listModels, addModel, removeModel, updateModel } from "./models";
+import { fetchAvailableModels } from "./provider-models";
 import {
   listProfiles,
   createProfile,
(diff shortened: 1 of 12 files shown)


pmos69:
Hi @Clare31 — heads-up, I just opened [#267](https://github.com/fathah/hermes-desktop/pull/267) implementing the same feature (live model discovery from `/v1/models` for autocomplete in the Providers + Models UI). I noticed your PR was opened then immediately closed earlier today, so I'm assuming you self-withdrew it — but if you'd planned to revise and re-submit, please flag and I'll happily yield or merge ideas.

A quick comparison of the two approaches is in [the comment I just left on #267](https://github.com/fathah/hermes-desktop/pull/267); the implementations are sufficiently different that mixing them is reasonable. No strong opinion on whose lands — just want to make sure we don't step on each other.

PR #261, opened 2026-05-19
fix(config): auto-populate model.api_key for known-host custom providers (#260)

### LINKED ISSUE #260: 添加deepseek大模型，使用一直报错
自定义添加deepseek大模型，用的时候一直提示Error: Error code: 401 - {'error': {'message': 'Authentication Fails, Your api key: ****ired is invalid', 'type': 'authentication_error', 'param': None, 'code': 'invalid_request_error'}}

pmos69:
Closes #260 (workaround — see \"Why this is a workaround\" below).

## Symptom

User configures DeepSeek (or Groq, Mistral, etc.) as a custom provider through the Providers UI, sends a chat, gets:

\`\`\`
Error: Error code: 401 - {'error': {'message': 'Authentication Fails, Your api key: ****ired is invalid', ...}}
\`\`\`

The masked tail (\`****ired\`, or whatever the last 4 chars of the user's \`OPENAI_API_KEY\` happen to be) gives it away: an OpenAI key is being sent to api.deepseek.com.

## Root cause (upstream, not fixable here)

Lives in \`hermes-agent/hermes_cli/runtime_provider.py::_resolve_openrouter_runtime\`. When \`model.provider == \"custom\"\` and the credential pool for the resolved \`custom:<name>\` is empty, the api_key fallback chain reads:

\`\`\`python
api_key_candidates = [
    explicit_api_key,
    (cfg_api_key if use_config_base_url else \"\"),  # model.api_key
    (os.getenv(\"OLLAMA_API_KEY\") if _is_ollama_url else \"\"),
    os.getenv(\"OPENAI_API_KEY\"),       # ← leaks here
    os.getenv(\"OPENROUTER_API_KEY\"),   # ← or here
]
\`\`\`

There's no provision for \"if base_url points at DeepSeek, prefer \`DEEPSEEK_API_KEY\`\" — the matching env var is never consulted. So even though the user typed \`DEEPSEEK_API_KEY=...\` into the Providers UI's env section, the gateway routes around it.

The desktop UI's \"Credential Pool\" section doesn't help either: the dropdown only offers literal \"custom\", so keys land under \`credential_pool[\"custom\"]\`, while the gateway looks up \`credential_pool[\"custom:deepseek\"]\`.

## The workaround in this PR

\`cfg_api_key\` (inline \`model.api_key\` in config.yaml) **does** win the fallback chain before the leak. So \`setModelConfig\` now auto-copies the matching env var to \`model.api_key\` when:

- Provider is bare \`\"custom\"\`, **and**
- The configured \`base_url\` matches a known commercial host (\`expectedEnvKeyForModel\` from #250 already has the mapping table), **and**
- The matching \`<NAME>_API_KEY\` is set in the active profile's \`.env\`

If the env var is missing, or the provider switches, or the URL stops matching a known host, the stale \`api_key\` line is stripped on the next save — so it never lingers past relevance.

Scope is deliberately narrow:
- Local LLMs (Ollama, vLLM, LM Studio, …) → **untouched** (no URL match → no auto-key)
- Built-in providers (anthropic, openai, …) → **untouched** (only fires for bare \"custom\")
- Custom hosts with no \`<NAME>_API_KEY\` env var → **untouched**

## Why this is a workaround

Right fix is upstream: \`_resolve_openrouter_runtime\` should drop the \`OPENAI_API_KEY\`/\`OPENROUTER_API_KEY\` candidates when the resolved base_url isn't openai.com / openrouter.ai. I'll file that on \`NousResearch/hermes-agent\` separately. Once it lands, this PR can be reverted with no user-visible change.

## Tests

\`tests/custom-provider-auto-key.test.ts\` — 8 cases:
- DeepSeek + env set → writes api_key ✓
- Quoted env value → quotes stripped ✓
- Groq + env set → writes api_key ✓
- Env var missing → no api_key written ✓
- Localhost / unknown host → no api_key written ✓
- Non-custom provider → no api_key written ✓
- Stale api_key cleared on provider switch ✓
- Existing api_key updated in place when env value changes ✓

411/411 total passing. Typecheck clean.

### DIFF
--- a/src/main/config.ts
+++ b/src/main/config.ts
@@ -671,7 +671,29 @@ function pickAutoApiKeyForCustomProvider(
   return trimmed || null;
 }
 
-const API_KEY_LINE_REGEX = /^[ \t]*api_key:\s*.*\n?/m;
+/**
+ * Locate the `model:` block in a YAML document and return the offsets that
+ * bracket its body (children lines, not counting the `model:` header line).
+ * Returns null when there's no `model:` block at all.
+ *
+ * The boundaries are needed to scope `api_key` add/update/remove operations
+ * to the model block — every `auxiliary.*` subsection has its own
+ * `api_key:` line, and a naive `/^api_key:/m` replace would clobber those
+ * instead.
+ */
+function findModelBlockBody(
+  content: string,
+): { start: number; end: number } | null {
+  const headerMatch = content.match(/^model:[^\S\r\n]*\r?\n/m);
+  if (!headerMatch) return null;
+  const start = headerMatch.index! + headerMatch[0].length;
+  // The body runs until the next line that starts at column 0 (next
+  // top-level key) or end of file.  Blank lines stay inside the block.
+  const after = content.slice(start);
+  const nextTopMatch = after.match(/^\S/m);
+  const end = nextTopMatch ? start + nextTopMatch.index! : content.length;
+  return { start, end };
+}
 
 export function setModelConfig(
   provider: string,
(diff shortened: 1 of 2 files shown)


pmos69:
Upstream tracking issue filed: [NousResearch/hermes-agent#28660](https://github.com/NousResearch/hermes-agent/issues/28660). Once that lands, this workaround can be reverted.

pmos69:
Pushed [`14044ad`](https://github.com/pmos69/hermes-desktop/commit/14044ad) — fixes a scoping bug in the original commit: the `api_key` write regex matched the *first* `api_key:` line in the file, which on real configs is `auxiliary.vision.api_key` (the `auxiliary.*` subsections each carry an `api_key: ''` line). End-to-end testing caught it leaking the user's DeepSeek key into the wrong field while leaving the model block untouched. The fix scopes all add/update/remove operations to the `model:` block boundaries. New regression test asserts exactly one non-empty `api_key` lands in the file with a realistic auxiliary-heavy config.

E2E test result: with the fix in place, the workflow that closes #260 is:
1. **Models** page → **+ Add Model** → set provider to *Local / Custom*, fill in base URL (e.g. `https://api.deepseek.com/v1`) and paste the API key into the *API Key* field — this writes `<NAME>_API_KEY` to `.env` via the existing flow.
2. Pick that model in the chat's model picker — fires `setModelConfig`, the workaround reads `.env` and writes `model.api_key` into config.yaml.
3. Chat works.

That's the path that exercises both the existing Models-page key field and this PR's workaround. The **Providers** page's model section is a less ergonomic alternative (no key field there for custom providers) but is the same code path under the hood.

Worth noting (separate from this PR): the Providers page's model save also calls `addModel`, which seems to create duplicate entries when the inferred provider tag differs across saves (e.g. the same `deepseek-reasoner` showed up as both `LOCAL/CUSTOM` and `OPENAI` after running through the test flow). Probably worth a small follow-up to dedupe by `(model, baseUrl)` before insert.

fathah:
Conflict

fathah:
Please rebase to main

PR #262, opened 2026-05-19
feat(providers): expand provider dropdown with OpenAI-compatible endpoints (#34)

### LINKED ISSUE #34: No Open AI Compatible models?
How come no Open AI Compatible models? Its pretty standard now and cant use any coding plans either.

pmos69:
Closes #34.

## What's wrong

The Provider dropdown ([constants.ts:17](https://github.com/fathah/hermes-desktop/blob/main/src/renderer/src/constants.ts)) offers only ten options. A user looking to connect a commercial OpenAI-compatible endpoint — DeepSeek, Groq, Mistral, Together, Fireworks, Cerebras, Perplexity, Hugging Face, Z.ai — or a subscription "coding plan" — ChatGPT Codex, xAI Grok OAuth, Qwen OAuth, Kimi for Coding — sees no matching entry. The `custom` value is the technically-correct path, but its `Local / Custom` label suggests local-only LLMs (Ollama, vLLM, LM Studio), not commercial APIs.

Reporter on #34: *"Its pretty standard now and cant use any coding plans either."*

## What this changes

- **PROVIDERS.options** now lists every provider name `hermes-agent`'s `resolve_provider` already accepts (see `hermes_cli/auth.py::resolve_provider`). 21 entries, ordered: aggregators → first-party APIs → OAuth / subscription plans → catch-all `custom`. Each dropdown value maps to a provider name the gateway routes natively — no further plumbing needed; existing runtime resolution code does the work.
- **PROVIDERS.labels** mirrors the same set so the chat bubble and credential-pool UI render canonical names ("DeepSeek" / "Groq" / "ChatGPT (Codex Plan)") instead of falling back to the raw value.
- **`custom` label** renamed from `Local / Custom` to `OpenAI Compatible / Local` so anyone searching for "OpenAI compatible" finds it. New i18n key `customOpenAICompatibleName` added across all 6 locales (en, es, pt-BR, id, ja, zh-CN). Chinese already used a near-identical phrasing — kept the existing tone.

## Why this is small

This is purely a constants-level change. Every new entry already works in `hermes-agent` — the gateway has been able to route to all of them via `provider:` in `config.yaml`; the desktop just didn't surface the names in its dropdown. No new IPC, no runtime logic, no provider-resolution work.

Pair behaviour with #261 (the workaround for the OPENAI_API_KEY leak): together they make the end-to-end flow "pick DeepSeek (or any commercial endpoint) → set the env var → it works".

## Stats

- 7 files changed, +45 / -4
- Typecheck clean
- 403/403 tests passing
- Lint baseline unchanged


pmos69:
Local end-to-end testing surfaced three downstream bugs that needed to land alongside the dropdown expansion — pushed as follow-up commits:

| Commit | What | Why |
|---|---|---|
| [`8de578e`](https://github.com/fathah/hermes-desktop/pull/262/commits/8de578e) | Clear stale `base_url` when changing the Provider dropdown | Switching from a named provider (e.g. openai-codex with base_url `chatgpt.com/backend-api/codex`) to a different named provider (e.g. deepseek) left the prior base_url in React state. Save wrote a nonsensical `(provider, base_url)` pair to `config.yaml` |
| [`f1da020`](https://github.com/fathah/hermes-desktop/pull/262/commits/f1da020) | Strip `baseUrl` in `selectModel` for non-custom entries | Same bug at the chat model-picker boundary. The Models list accumulates entries from prior saves; clicking one with a stale `baseUrl` propagated it back to `config.yaml`, routing requests to the wrong host |
| [`4bf616f`](https://github.com/fathah/hermes-desktop/pull/262/commits/4bf616f) | Reload Models page entries on tab visibility | Models component mounts once and never refreshes — entries added via `addModel` from elsewhere (Providers save, chat picker first-touch) didn't appear until app restart |
| [`525c883`](https://github.com/fathah/hermes-desktop/pull/262/commits/525c883) | Decouple Models-library write debounce (2 s) from active-config debounce (500 ms) | Each 500 ms keystroke pause was creating a Models-library entry. Typing `deepseek-reasoner` slowly left ghost entries `deepseek-reaso`, `deepseek-reason`, `deepseek-reasone`, etc. — each a distinct model id that `addModel`'s `(model, provider)` dedup couldn't collapse |

All four are user-visible issues that the expanded dropdown surfaces more often (more provider options = more provider switches = more chances for stale state to leak). They could each be their own PR but they're collectively the cohesive "provider-switching paths now feel right" change.

End-to-end validated locally against a real DeepSeek key: pick DeepSeek from the new dropdown entry → set model → chat works on first try. Same for round-trips (switch provider, switch back). 403/403 tests pass, typecheck clean.

PR #267, opened 2026-05-19
feat(models): live model-discovery autocomplete for OpenAI-compatible providers (#265)

### LINKED ISSUE #265: feat(models): live model-discovery autocomplete for OpenAI-compatible providers
### Idea

When a user picks a provider in the **Providers** page (or in the **Add Model** dialog), the Model field should offer the provider's actual catalogue as autocomplete suggestions instead of being a freeform text input. Same for the Models page modal.

### Why

Currently users have to know and type the exact model id (`deepseek-chat`, `llama-3.1-70b-versatile`, `mistral-large-latest`, etc.). Typos create either a 404 on send or — worse — pollute the Models library with dead entries (e.g. `deepseek-reaso` from a typing pause; see #262 follow-ups). A live-discovery dropdown:
- removes the typing burden
- exposes models the user didn't know about
- automatically reflects new releases without a desktop update

### Upstream already has the helper

`hermes_cli/models.py::fetch_api_models(api_key, api_url)` in `hermes-agent` already does this. The CLI's `/model` picker uses it for live discovery (see `hermes_cli/model_switch.py:1529`). The desktop has no equivalent path.

### Provider coverage

`/v1/models` returns sensible results for OpenAI, DeepSeek, Groq, Mistral, Together, Fireworks, Cerebras, Perplexity, Hugging Face, OpenRouter, Z.ai (~12 of the 21 dropdown options after [#262](https://github.com/fathah/hermes-desktop/pull/262)). Anthropic returns a curated subset; Google / xAI / openai-codex / OAuth-only providers either don't expose this endpoint or use a different shape. Providers without coverage fall back to today's free-text Model field — no regression.

### Implementation sketch (~250 LOC)

1. **New IPC** `discoverProviderModels(provider, baseUrl?, apiKey?) → string[]`
   - Main resolves the canonical base URL from the provider name when missing (reuse the `PROVIDER_ENV_KEYS` / `URL_TO_ENV_KEY` mappings already in `installer.ts`)
   - Reads the matching `<NAME>_API_KEY` from `.env` if `apiKey` not provided
   - HTTP `GET <base_url>/models` with `Authorization: Bearer <key>`, parses OpenAI-format `{ data: [{id, ...}, ...] }`, returns sorted unique ids, or `[]` on any failure
   - Caches per `(provider, baseUrl)` for ~5 min in the existing config-cache pattern

2. **UI in Providers page**
   - When provider changes to a discovery-capable one **and** the matching env var is set: trigger discovery in background
   - Model field becomes a `<datalist>`-backed input — native HTML, no new library, free text still allowed
   - Right of the field: a small "↻" button + status text (`6 models found` / `Set <NAME>_API_KEY to load list`)

3. **Same datalist in Models page Add/Edit dialog**

### Out of scope

- OAuth providers (openai-codex, xai-oauth, qwen-oauth) — would need a different resolution path through the agent's auth.json. Could be a follow-up.
- Caching beyond per-session.
- Visual provider-icon decoration on the autocomplete list.

### Acceptance

- [ ] Picking DeepSeek in Providers with `DEEPSEEK_API_KEY` set shows `deepseek-chat`, `deepseek-reasoner`, etc. as Model field suggestions
- [ ] Picking a provider without an API key set shows the hint and no error
- [ ] Network failure / 5xx upstream → list silently empty, free-text input still works
- [ ] Picking a provider without `/v1/models` support → no API call made, free-text input still works

pmos69:
Closes #265.

## What this does

The Model field in **Providers** and the Model ID field in **Models → Add/Edit** now offer an autocomplete dropdown sourced from the provider's actual `/v1/models` catalogue. No more typing model ids from memory — pick from the live list, or keep typing free-form if you want a model the catalogue doesn't list.

Mirrors what `hermes-agent`'s CLI `/model` picker already does via `hermes_cli/models.py::fetch_api_models`; this brings the same affordance to the desktop UI.

## Providers covered

Out of the box, twelve providers' `/v1/models` endpoints are recognised — every entry here has a canonical inference URL hardcoded mirroring `hermes-agent`'s `PROVIDER_REGISTRY`:

| Provider | Endpoint shape | Auth |
|---|---|---|
| OpenAI, OpenRouter, DeepSeek, Groq, Mistral, Together, Fireworks, Cerebras, Perplexity, Hugging Face, Z.ai | OpenAI-format `{ data: [{ id, … }, …] }` | `Authorization: Bearer …` |
| Anthropic | Same shape | `x-api-key: …` + `anthropic-version: 2023-06-01` |

Providers without a usable endpoint (`auto`, `nous`, `google`, `xai`, `openai-codex`, `xai-oauth`, `qwen-oauth`, `qwen`, `minimax`, `kimi-coding`, `custom` w/o base_url) short-circuit with `status="unsupported"` so the UI silently falls back to free-text input — no regression.

## How it works

1. **`src/main/model-discovery.ts`** — new module. HTTP GET to `<base_url>/models`, parses the response, returns sorted unique model ids. Network failures, malformed JSON, and 4xx/5xx responses all return an empty list and never throw. Results cached per `(provider, baseUrl)` for 5 minutes.
2. **IPC** `discover-provider-models` exposed via preload. Takes `(provider, baseUrl?, apiKey?, profile?)`. When `apiKey` is omitted, the main process reads the matching `<NAME>_API_KEY` from the profile's `.env` using the existing `expectedEnvKeyForModel` helper.
3. **`src/renderer/src/hooks/useDiscoveredModels.ts`** — small React hook with a 400 ms debounce on provider/baseUrl change. Cancels in-flight requests on rapid input changes.
4. **Providers + Models UI** — Model fields get an HTML `<datalist>` populated from the hook's result, plus a status hint under the input (loading / N available / no-key / error / hidden when unsupported) and a manual "↻" refresh button.

The native `<datalist>` is a non-blocking suggestion list — the user can still type any model name the provider doesn't advertise (catches edge cases like DeepSeek's `deepseek-chat` alias, which is gateway-side and not exposed by their `/v1/models`).

## Dependency awareness (related PRs)

Strictly speaking this PR is independent — it'll merge cleanly on top of `main` and the discovery feature works on its own. But for the **full end-to-end flow** (pick a named provider in the dropdown → see its model list → send a chat that succeeds), it pairs naturally with:

- **#262** (provider dropdown expansion) — without it, users can only reach the named-provider discovery code path via the `custom` provider + base_url. With it, picking DeepSeek/Groq/Mistral/etc. directly from the dropdown triggers discovery.
- **#241** (API_SERVER_KEY forwarding on local chat) — without it, sending a chat to the discovered model returns 401 from the local gateway. Affects chat success, not discovery itself.
- **#261** (`OPENAI_API_KEY` leak workaround for custom providers) — without it, picking a custom-provider model and sending a chat may 401 with someone else's key. Again, chat success only.

None of those are hard blockers for this PR's logic, but they collectively define "discovery actually feels useful when you click through it". I tested all four stacked locally end-to-end against real DeepSeek, Groq, Mistral, and OpenRouter accounts before opening this.

## Tests

`tests/model-discovery.test.ts` — 11 tests against a real loopback HTTP server (no network):

- Parses standard OpenAI shape, sorted and deduplicated ✓
- `status=no-key` when caller passes nothing and `.env` has nothing ✓
- `status=unsupported` for the closed no-discovery provider set ✓
- Bearer header forwarded correctly ✓
- Anthropic's `x-api-key` + `anthropic-version` headers ✓
- Malformed JSON → empty list, no throw ✓
- 4xx / 5xx → empty list, no throw ✓
- Duplicate model ids in response → deduped ✓
- Cache hit on second call (1 actual HTTP call for 2 lookups) ✓
- `status=unknown-host` when neither caller nor mapping resolves a URL ✓
- `.env` fallback when caller passes no `apiKey` ✓

414/414 total passing, typecheck clean.

## Stats

10 source files + 1 test file, 5 locale files. +833 / -15.

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -13,6 +13,7 @@ import type { AppUpdater } from "electron-updater";
 import icon from "../../resources/icon.png?asset";
 import type { Attachment } from "../shared/attachments";
 import { stageAttachment, clearStagedAttachments } from "./attachment-staging";
+import { discoverProviderModels } from "./model-discovery";
 import {
   checkInstallStatus,
   verifyInstall,
@@ -724,6 +725,20 @@ function setupIPC(): void {
   ipcMain.handle("clear-staged-attachments", (_event, sessionId: string) => {
     clearStagedAttachments(sessionId);
   });
+
+  // Model discovery — fetch the provider's /v1/models for autocomplete.
+  ipcMain.handle(
+    "discover-provider-models",
+    (
+      _event,
+      provider: string,
+      baseUrl: string | undefined,
+      apiKey: string | undefined,
+      profile?: string,
+    ) => {
+      return discoverProviderModels(provider, baseUrl, apiKey, profile);
+    },
+  );
 
   // Gateway
   ipcMain.handle("start-gateway", async () => {
--- a/src/main/model-discovery.ts
+++ b/src/main/model-discovery.ts
@@ -1,0 +1,254 @@
+/**
+ * Provider model discovery.
+ *
+ * Hits the provider's OpenAI-compatible `/models` endpoint and returns the
+ * list of model ids it advertises.  Used to power autocomplete in the
+ * Providers UI Model field and the Models page Add/Edit dialog so users
+ * don't have to type the exact id from memory.
+ *
+ * Upstream `hermes-agent` has an equivalent helper at
+ * ``hermes_cli/models.py::fetch_api_models`` used by the TUI's /model
+ * picker; this mirrors that flow on the desktop side without going
+ * through the Python CLI.
+ */
+import http from "http";
+import https from "https";
+import { URL } from "url";
+import { readEnv } from "./config";
+import { expectedEnvKeyForModel } from "./installer";
+
+/** Canonical inference base URL for each named provider, mirroring
+ *  hermes-agent's PROVIDER_REGISTRY defaults.  Only the providers whose
+ *  `/models` endpoint we know responds usefully are listed; everything
+ *  else falls through to caller-supplied baseUrl or no discovery. */
+const PROVIDER_BASE_URLS: Record<string, string> = {
+  openai: "https://api.openai.com/v1",
+  openrouter: "https://openrouter.ai/api/v1",
+  deepseek: "https://api.deepseek.com/v1",
+  groq: "https://api.groq.com/openai/v1",
+  mistral: "https://api.mistral.ai/v1",
+  together: "https://api.together.xyz/v1",
+  fireworks: "https://api.fireworks.ai/inference/v1",
+  cerebras: "https://api.cerebras.ai/v1",
+  perplexity: "https://api.perplexity.ai",
+  huggingface: "https://router.huggingface.co/v1",
+  zai: "https://api.z.ai/api/paas/v4",
+  anthropic: "https://api.anthropic.com/v1",
+};
+
+/** Providers whose `/models` we never call — either they don't expose it,
+ *  use a different protocol, or rely on OAuth credentials we can't
+ *  reproduce from a static env var. */
+const NON_DISCOVERABLE_PROVIDERS = new Set<string>([
+  "auto",
+  "custom",
+  "nous",
+  "google",
+  "xai",
+  "openai-codex",
+  "xai-oauth",
+  "qwen-oauth",
+  "qwen",
+  "minimax",
+  "kimi-coding",
+]);
+
+// In-memory result cache to avoid hammering the provider on every keystroke.
+interface CacheEntry {
+  models: string[];
+  ts: number;
+}
+const CACHE_TTL_MS = 5 * 60 * 1000; // 5 minutes
+const _cache = new Map<string, CacheEntry>();
+
+function cacheKey(provider: string, baseUrl: string): string {
+  return `${provider.toLowerCase()}|${baseUrl.replace(/\/+$/, "").toLowerCase()}`;
+}
+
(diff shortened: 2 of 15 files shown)


pmos69:
Worth flagging: [PR #259](https://github.com/fathah/hermes-desktop/pull/259) by @Clare31 (opened earlier the same day, self-closed within ~30 seconds, never reviewed) implements a similar feature. The two approaches differ enough that picking one isn't an obvious choice:

| Aspect | #259 (Clare31) | #267 (this PR) |
|---|---|---|
| HTTP client | `fetch()` (Node 18+) | Node `http` / `https` |
| Failure mode | Throws | Returns `status` enum + empty list, never throws |
| Provider→base URL | Caller supplies baseUrl | Hardcoded canonical URLs for 12 named providers + caller override for `custom` |
| Trigger | Button press ("Fetch Available Models") | Auto on provider/baseUrl change, debounced 400 ms; manual ↻ refresh button |
| Caching | None | Per-`(provider, baseUrl)`, 5 min TTL |
| Coverage | Generic OpenAI-compatible endpoints | Generic + Anthropic-specific headers (x-api-key + anthropic-version) |
| i18n | en, zh-CN | en, es, id, ja, pt-BR, zh-CN |
| Tests | One file | 11 cases against a real loopback server |

Posted independently before I noticed #259 existed — the implementations were arrived at separately. Happy to fold ideas in either direction if a reviewer wants a merged form (e.g. Clare31's explicit "Fetch" button is more conservative on rate-limit-sensitive APIs; mine's autocomplete-on-change is friendlier for typo-prone fields). Just say what to combine.

PR #269, opened 2026-05-19
Add Russian localization for Hermes Desktop

Ovladimirovich:


Ovladimirovich:
Added complete Russian localization for Hermes Desktop UI.

All screens, navigation, settings, chat, models, tools, providers, memory, schedules, skills, office, gateway, sessions, agents, kanban, welcome, soul and setup are now available in Russian.

pmos69:
Heads-up — this PR looks like it won't build as-is. `src/shared/i18n/index.ts` adds:

```ts
import kanbanEn from "./locales/en/kanban";
```

and references `kanban: kanbanEn` in the resources, but `src/shared/i18n/locales/en/kanban.ts` isn't in the PR (only `locales/ru/kanban.ts` is). The EN source file for the new `kanban` namespace is missing — `npm run typecheck` / the build will fail on the unresolved import.

Looks like a forgotten `git add` for the new `en/kanban.ts` file (the `ru/*` files all made it in). Adding that one file should fix it.

Also a smaller note: `Kanban.tsx` wires `kanban.title/subtitle/refresh/dispatch`, but the `kanban` namespace defines ~20 keys — the column labels (Triage/To-do/Ready/Running/Blocked) and other strings are still hardcoded English. Not blocking, just flagging that Kanban isn't fully localized yet.

pmos69:
Thanks for adding Russian localization. This needs a rebase/update before we can merge.

The PR currently fails web typecheck because `AppLocale` now includes `"ru"`, but `LANGUAGE_NATIVE_NAMES: Record<AppLocale, string>` in `Settings.tsx` does not include a `ru: "Русский"` entry.

Also, this branch is now quite far behind `main`, and the Russian resource set is missing newer visible namespaces/keys from current `main`, including `discover`, `diagnose`, newer navigation keys, expanded Kanban strings, settings privacy/analytics/font/rounded-corners strings, and model discovery strings. With Russian selected those areas would fall back to English.

Could you please rebase on current `main`, add the missing `ru` language name, and fill in the newly added locale keys? The bot comments about `navigation.kanban` and the truncated memory string should be fixed as part of the same update.

PR #280, opened 2026-05-20
feat(providers): in-app OAuth sign-in for subscription providers

pmos69:
## Summary

Hermes Desktop can *select* OAuth/subscription providers (Codex, Gemini CLI, etc.) but had no way to *authenticate* them — users had to drop to a terminal and run `hermes auth add <provider>` by hand. This adds in-app sign-in.

## What it does

- New **"Subscription / OAuth Plans"** section on the Providers screen, with a **Sign in** card per provider: ChatGPT (Codex Plan), xAI Grok, Qwen, Gemini (CLI OAuth), MiniMax.
- Clicking **Sign in** opens a modal that spawns `hermes auth add <provider> --type oauth`, streams the CLI output live, and shows a prominent green ✓ / red ✗ outcome banner. Closing mid-flow kills the subprocess so its loopback redirect server doesn't linger.
- **Codex device-code helper:** Codex (unlike the browser-loopback providers) prints a URL + short code instead of opening a browser. The desktop detects that prompt, **opens the page automatically** and **copies the code to the clipboard**, adding a line to the log so the user knows. Detection is a pure, tested function keyed on the Codex prompt shape — it won't misfire for the browser-loopback providers or non-`https` URLs.

## Implementation

- `src/main/hermes-auth.ts` — spawn + stream + single-flight guard + cancel + `detectDeviceCode`.
- `oauth-login` / `oauth-login-cancel` IPC; preload surface.
- `OAuthLoginModal.tsx` on the Providers screen.
- i18n across all 8 locales.
- `tests/hermes-auth.test.ts` — 14 cases: provider allowlist, arg construction, profile flag, success/failure/cancel exit mapping, concurrent-login rejection, cancel, and device-code parsing.

## Manual testing

- **Codex** ✅ — device-code flow: page auto-opens, code pre-copied, green banner on completion.
- **Gemini (CLI OAuth)** ✅ — browser OAuth, credential added (`hermes auth list` confirms).
- **MiniMax** ✅ — browser OAuth, credential added.
- **xAI Grok** — flow launches the browser correctly; couldn't complete (no xAI subscription to test against).
- **Qwen** — surfaces a clear error: `hermes auth add qwen-oauth` requires the external Qwen CLI to be authenticated first (a hermes-agent requirement). The red banner shows it; left in the list since it works for users who *do* have the Qwen CLI.

`npm run typecheck` clean, `npm test` 502 passing.

## Known pre-existing item (not changed here)

`openai-codex` is in `PROVIDERS_WITHOUT_API_KEYS` (added by #102), so the install gate treats it as "configured" even with **no credential** — the only OAuth provider that bypasses the gate unauthenticated. The other four correctly gate on `hasOAuthCredentials`. Making this consistent (dropping `openai-codex` from that set so all five gate on real credentials) is a behaviour change to a merged PR and out of scope here — flagging for a maintainer call.

## Test plan

- [ ] Providers → "Subscription / OAuth Plans" → Sign in on each provider
- [ ] Codex: page opens, code is in clipboard, green banner on success
- [ ] Browser-loopback providers (Gemini/MiniMax/xAI): browser opens, green banner on success
- [ ] Close the modal mid-flow → no orphan `hermes`/python process
- [ ] Switch language → section + modal strings translate

### DIFF
--- a/src/main/hermes-auth.ts
+++ b/src/main/hermes-auth.ts
@@ -1,0 +1,156 @@
+import { spawn, type ChildProcess } from "child_process";
+import { homedir } from "os";
+import {
+  HERMES_PYTHON,
+  HERMES_REPO,
+  HERMES_HOME,
+  hermesCliArgs,
+  getEnhancedPath,
+} from "./installer";
+import { HIDDEN_SUBPROCESS_OPTIONS } from "./process-options";
+import { stripAnsi } from "./utils";
+
+/**
+ * Provider identifiers that authenticate via an interactive OAuth flow
+ * (`hermes auth add <provider> --type oauth`) rather than a static API
+ * key. Mirrors hermes-agent's `_OAUTH_CAPABLE_PROVIDERS` set, minus the
+ * API-key-capable `anthropic`/`nous` which the desktop sets up through
+ * the normal key flow.
+ */
+export const OAUTH_LOGIN_PROVIDERS = [
+  "openai-codex",
+  "xai-oauth",
+  "qwen-oauth",
+  "google-gemini-cli",
+  "minimax-oauth",
+] as const;
+
+export type OAuthLoginProvider = (typeof OAUTH_LOGIN_PROVIDERS)[number];
+
+export function isOAuthLoginProvider(value: string): value is OAuthLoginProvider {
+  return (OAUTH_LOGIN_PROVIDERS as readonly string[]).includes(value);
+}
+
+export interface OAuthLoginResult {
+  success: boolean;
+  error?: string;
+}
+
+/**
+ * Parse a device-code login prompt out of the CLI's streamed output.
+ * The OpenAI Codex flow — unlike the browser-loopback providers —
+ * prints a URL to open and a short code to enter rather than opening a
+ * browser itself. Detecting both lets the desktop open the page and
+ * pre-copy the code so the user only has to paste.
+ *
+ * Returns null until both parts are present. Only an `https:` URL is
+ * accepted (the value is fed to `shell.openExternal`).
+ */
+export function detectDeviceCode(
+  text: string,
+): { url: string; code: string } | null {
+  const urlMatch = text.match(
+    /Open this URL in your browser:\s*\n\s*(https:\/\/\S+)/,
+  );
+  const codeMatch = text.match(/Enter this code:\s*\n\s*(\S+)/);
+  if (urlMatch && codeMatch) {
+    return { url: urlMatch[1], code: codeMatch[1] };
+  }
+  return null;
+}
+
+// Only one interactive login can run at a time — the renderer surfaces a
+// single modal. Tracked so the renderer can cancel a flow the user
+// abandoned (otherwise the CLI's loopback OAuth server lingers).
+let activeProc: ChildProcess | null = null;
+
+/**
+ * Run `hermes auth add <provider> --type oauth`, streaming the CLI's
+ * stdout/stderr line-by-line to `emit`. The CLI opens the system browser
+ * for the OAuth consent step and runs a localhost loopback server to
+ * catch the redirect; this function just supervises that subprocess.
+ *
+ * Resolves `{ success: true }` on exit code 0, `{ success: false, error }`
+ * otherwise (non-zero exit, spawn failure, or cancellation).
+ */
+export function runHermesAuthLogin(
+  provider: string,
+  emit: (chunk: string) => void,
+  profile?: string,
+): Promise<OAuthLoginResult> {
+  return new Promise((resolve) => {
+    if (!isOAuthLoginProvider(provider)) {
+      resolve({ success: false, error: `Unsupported OAuth provider: ${provider}` });
+      return;
+    }
+    if (activeProc) {
+      resolve({
+        success: false,
+        error: "Another sign-in is already in progress.",
+      });
+      return;
+    }
+
+    // `--type oauth` is explicit so the CLI never falls back to an
+    // interactive "API key or OAuth?" prompt on a stdin we've closed.
+    const subArgs =
+      profile && profile !== "default"
+        ? ["-p", profile, "auth", "add", provider, "--type", "oauth"]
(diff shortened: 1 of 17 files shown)


fathah:
Thanks for this — the feature is well-structured and the test coverage is solid. A couple of things need fixing before this can merge, plus some smaller items worth addressing.

## Must fix

**1. CSS uses non-existent `--danger` / `--danger-bg` tokens**

The new `.oauth-login-result-error` rule references `var(--danger, #e5484d)` and `var(--danger-bg, …)`, but these tokens don't exist in the design system. The correct tokens are `--error` and `--error-bg` (defined in both light and dark themes). The hardcoded fallback `#e5484d` also doesn't match either theme's `--error` value, so the error banner will render with the wrong colour in light mode.

Also, the success rule uses `var(--success, #2e7d32)` — the fallback doesn't match the existing `--success` value either. Since the tokens already exist, the fallbacks aren't needed.

```css
/* replace the two new result classes with: */
.oauth-login-result-success {
  color: var(--success);
  background: var(--success-bg);
}

.oauth-login-result-error {
  color: var(--error);
  background: var(--error-bg);
}
```

**2. `provider-key-card` / `provider-keys-grid` CSS classes are used but never defined**

`Providers.tsx` renders `provider-keys-grid`, `provider-key-card`, `provider-key-card-head`, and `provider-key-card-title`, but none of these appear in `main.css`. The cards will have no layout or visual treatment. You'll need to add those rules alongside the other new CSS in `main.css`.

## Should fix

**3. `event.sender.send` is called without checking if the renderer is still alive**

In the `oauth-login` IPC handler, the progress chunk callback does:

```ts
event.sender.send("oauth-login-progress", chunk);
```

If the user closes the window mid-flow (before `cancelOAuthLogin` fires), `event.sender` will be destroyed and this throws. Guard it:

```ts
if (!event.sender.isDestroyed()) {
  event.sender.send("oauth-login-progress", chunk);
}
```

## Minor / nits

**4. `detectDeviceCode` regex uses `\s*` for indentation** — `\s*` also matches newlines, so a blank line between the label and the value would be silently consumed. `[^\S\n]*` (horizontal whitespace only) is more precise and avoids a potential false positive.

**5. The flagged `openai-codex` / `PROVIDERS_WITHOUT_API_KEYS` inconsistency** — totally fine to leave out of scope here, but worth opening a follow-up issue so it doesn't get lost.

**6. No `aria-label` on the Sign In button** in `Providers.tsx`. The close button has one; the sign-in button doesn't. Low priority but easy to add.

## What looks good

- The single-flight guard and the `settled` flag in `runHermesAuthLogin` are correct.
- `openExternalUrl` is correctly reused for the device-code URL, so it goes through the existing `isAllowedExternalUrl` security gate. The `https:` enforcement in `detectDeviceCode` is redundant but harmless.
- The `startedRef` guard in `OAuthLoginModal` is the right fix for React StrictMode double-invocation against the single-flight lock.
- Reusing `.models-modal-*` for the modal shell is fine — those classes are generic enough.
- i18n coverage across all 8 locales is complete.
- The 14 unit tests cover the important paths well (`vi.hoisted` pattern for the mock is correct).

Fix items 1 and 2 and this is good to go.

pmos69:
Thanks @fathah — pushed `afad55d` addressing the review.

**1. CSS tokens.** Fixed — both `.oauth-login-result-success` and `.oauth-login-result-error` now use `var(--success)` / `var(--success-bg)` and `var(--error)` / `var(--error-bg)` respectively, with the fallbacks dropped (the tokens exist).

**2. `provider-key-*` rules.** I think this one's a misread on the diff — the four classes are already defined in `main.css` (lines 2210–2250 on this branch): `.provider-keys-grid` (auto-fill grid), `.provider-key-card` (bordered card), `.provider-key-card-head` (flex row), `.provider-key-card-title` (truncating title). They render styled in dev. Happy to point you at any specific snippet if I'm looking at a different file than you.

**3. `event.sender.send` race.** Guarded — the progress callback now early-returns when `event.sender.isDestroyed()`, so closing the modal mid-flow no longer throws.

**4. `detectDeviceCode` regex.** Tightened — `\s*` swapped for `[^\S\n]*` on both the URL and the code patterns. A new test case asserts that a blank line between the label and the value does **not** silently match (covers both the URL-gap and the code-gap shapes).

**5. `openai-codex` / `PROVIDERS_WITHOUT_API_KEYS` inconsistency.** Noted — I'll file the follow-up issue after I've worked through your other reviews, so I don't fragment cross-cutting items across multiple bugs.

**6. `aria-label` on Sign In button.** Added — reads `"<localised Sign In> — <provider.name>"` so screen readers announce which provider the click targets.

Verification: `npm run typecheck` clean, `tests/hermes-auth.test.ts` 15/15 pass (14 original + the new blank-line case), full suite 503 passed / 3 skipped / 0 failures.

pmos69:
Quick note on the failing CI here — it's **not** in this PR's code.

When I pushed `afad55d` GitHub Actions ran typecheck against the merge result (PR + current `main`) and failed with 9 `TS2339` errors in `MessageRow.tsx`, `MessageList.tsx`-adjacent hooks, and `transcriptUtils.ts`. Verified by checking out `origin/main` at `f3105d3` locally and running `npm run typecheck`: the same 9 errors reproduce on `main` alone.

Root cause is PR #327 widening `ChatMessage` into a 4-way union (`ChatBubbleMessage | ReasoningMessage | ToolCallMessage | ToolResultMessage`) without narrowing in the consumers that still read `.content` / `.attachments` off the bare union. So every open PR sitting against this `main` will fail the same CI gate (the one #282 just installed).

I opened PR #332 to fix it cleanly — small, no runtime change, hoists the `isBubble` predicate that already lives inside `MessageList` into `types.ts` as `isBubbleMessage`, narrows the four broken sites, drops the `as Extract<…>` casts. typecheck clean; full suite 562 passed / 3 skipped.

Once #332 lands, this PR's CI should flip green on a re-run (no code change needed on `feat/oauth-login-ui` — the review feedback in `afad55d` is unaffected).

PR #293, opened 2026-05-21
fix(chat): resume gateway sessions via X-Hermes-Session-Id header

### LINKED ISSUE #226: Session data loss and incorrect messageCount
After using Hermes Desktop for a while, the following issues were observed:
1. The `messageCount` field in `sessions.json` does not match the actual number of messages (generally lower than actual)
2. Some session files (`sessions/*.json`) are mysteriously missing
3. After closing and reopening the client, some session content is incomplete

## Steps to Reproduce

1. Use Hermes Desktop for multiple conversations
2. Close the client normally
3. Reopen and check `~/.hermes/desktop/sessions/` directory and `sessions.json`

## Expected Behavior

- `messageCount` should accurately reflect the actual number of messages in each session
- All session files should be retained completely
- Content should not be lost after closing

## Actual Behavior

- `messageCount` is generally too small (e.g., records only 15 messages when there are actually 200+)
- At least 1 session file is completely missing (`api-6e2f193c79997b6e.json`)
- Some sessions have incomplete message history

## Environment

- OS: macOS 26.3.1
- Hermes Desktop version: (please fill)
- Configuration: Using custom model provider (tokenbank.c2c.run:7100)

## Possible Causes

Suspect that the client, when saving/updating `sessions.json`:
- Does not correctly calculate `messageCount`
- Concurrent writes cause data overwrite or loss
- Does not flush to disk on close

## Temporary Workaround

1. Manually backup important session files
2. Use `hermes dashboard` (web version) as an alternative

pmos69:
## Summary

The gateway resumes an existing session from the **`X-Hermes-Session-Id` request header** — the `session_id` field in the request body is **ignored**. The desktop only sent the body field, so **every message forked a new server-side session**.

That single root cause produces every symptom in #226:

- **History lost on reopen** — reloading a session shows only the turns from when it was momentarily "current"; later turns forked into other sessions.
- **`messageCount` too low** — most sessions end up holding only a turn or two.
- **Sessions "missing"** — actually *proliferating*, roughly one per turn.

The fix sends the resume id as the `X-Hermes-Session-Id` request header in `sendMessageViaApi`. The gateway already echoes it back in the response header (which the desktop reads), so continuity is now stable.

This is hermes-agent's own documented mechanism — `gateway/run.py`'s api_server proxy client sets the same header, with the comment: *"The remote api_server can maintain session continuity via X-Hermes-Session-Id, so it loads its own history."*

## Verification

Direct probes against the local gateway:

- `POST /v1/chat/completions` with `session_id` in the **body** → a *new* `X-Hermes-Session-Id` on every call (forks).
- Same request with `X-Hermes-Session-Id` as a **request header** → the *same* id echoed back (resumes).

End-to-end in the app with the fix: a 4-turn conversation persisted as **one** session (8 messages appended under a single id), instead of four separate 2-message sessions.

Fixes #226.

## Test plan

- [x] New chat, multiple turns → all turns persist under one session id (verified in `state.db`).
- [x] Switch away and reopen the session → full history intact.
- [x] `npm run typecheck` and `vitest` pass.

### DIFF
--- a/src/main/hermes.ts
+++ b/src/main/hermes.ts
@@ -312,6 +312,14 @@ function sendMessageViaApi(
     "Content-Type": "application/json",
     ...getRemoteAuthHeader(),
   };
+  // Session continuity: the gateway resumes an existing session via the
+  // `X-Hermes-Session-Id` *request header* — the `session_id` body field
+  // above is not honoured. Without this header every request forks a new
+  // server-side session, fragmenting stored history and messageCount
+  // (issue #226). The gateway echoes the id back in the response header.
+  if (_resumeSessionId) {
+    headers["X-Hermes-Session-Id"] = _resumeSessionId;
+  }
   // Local API server key (API_SERVER_KEY in the profile's .env /
   // config.yaml) only applies in local mode — in remote/SSH mode the
   // remote endpoint's own auth header (set above) is authoritative and
--- a/tests/hermes-api.test.ts
+++ b/tests/hermes-api.test.ts
@@ -189,4 +189,51 @@ describe("sendMessageViaApi forwards resumeSessionId", () => {
 
     expect(parsed).not.toHaveProperty("session_id");
   });
+
+  it("sends the X-Hermes-Session-Id request header when resuming", async () => {
+    const testSessionId = "session-abc-123";
+
+    await sendMessage(
+      "hello",
+      {
+        onChunk: () => {},
+        onDone: () => {},
+        onError: () => {},
+      },
+      "default",
+      testSessionId,
+    );
+
+    const chatRequest = capturedRequests.find((r) =>
+      r.url.includes("/v1/chat/completions"),
+    );
+    expect(chatRequest).toBeDefined();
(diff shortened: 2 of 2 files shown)


openapphub:
Important update, looking forward to an early merge.

pmos69:
Thanks @openapphub — good to know it's biting others too.

Status for whoever picks this up: I re-tested the branch against the current `main`.

- **Mergeable** — branch is on the current `main` tip + the two commits here; GitHub shows it `MERGEABLE` / clean, no conflicts.
- **Typecheck** — passes (node + web).
- **Tests** — full suite green: 490 passing, 3 skipped. `tests/hermes-api.test.ts` includes a regression test asserting the `X-Hermes-Session-Id` header is sent on resume, so the fix is locked against a future refactor silently dropping it.

The change is small and self-contained — `src/main/hermes.ts` (+8) plus the test (+47). Happy to rebase or adjust if anything's needed.

fathah:
@pmos69  thanks for this PR.

PR #303, opened 2026-05-21
Render agent-generated media in chat (closes #299)

### LINKED ISSUE #299: Image generation problem
After the image is generated, it cannot be downloaded. Can the image be reduced in size? Click to view and download.

pmos69:
## Summary

Closes #299. The agent delivers generated files (images, PDFs, etc.) but the desktop app rendered the delivery path as literal text. This PR makes that media appear inline in the chat.

**Detection** — three signals in an agent reply are recognised:
1. **`MEDIA:<path-or-url>` tokens** — hermes-agent's explicit delivery protocol. Trusted, rendered eagerly. Quoted forms (backtick / single / double) handle paths with spaces.
2. **Inline bare absolute paths** — an absolute path with a known extension named anywhere in the prose. The matcher is anchored: a negative lookbehind blocks matches that start mid-token or inside a URL, the extension must be followed by whitespace/sentence punctuation, and matches inside fenced or inline code are skipped.
3. **Whole-line bare paths** — a line that is exactly an absolute path, which also covers paths containing spaces.
4. Markdown image syntax (`alt`) is rendered too, via an `img` override.

**False-positive safety** — every *inferred* bare path (inline or whole-line) is verified to point at a real, reachable file (`mediaFileExists` IPC) before it is rendered as media. Until verified — and if verification fails — the original text is shown verbatim, so a path merely mentioned in prose never silently becomes media. Explicit `MEDIA:` tokens and URLs are trusted as-is.

**Rendering**
- Images render inline (size-capped), click to open a zoom/lightbox with Save and close actions.
- Non-image media renders as a compact download chip.
- Local paths are resolved to data URLs through the main process; a 25 MB guard applies.

**Right-click menu** — images, the lightbox image, and download chips now have a native context menu: **Open** (local file → OS default handler; web URL → browser) and **Save as…** (data: URLs are save-only). Labels go through i18n — new `chat.media.*` keys for all eight locales.

## Notes

- Detection covers both whole-line *and* inline path mentions; there is no assumption that the agent always places a filename on its own line.
- The new right-click menu and the generic copy/paste context menu (separate PR #301) are independent. When both land, whichever merges second should add a one-line guard so the media menu wins on media elements — flagged here for the reviewer.

## Test plan

- [x] Full test suite — 513 passed / 3 skipped; `mediaUtils` parser has 25 unit tests
- [x] Live: image renders inline, lightbox + save work
- [x] Live: bare PDF path renders as a download chip
- [x] Live: inline path mention detected when the file exists
- [x] Live: right-click Open / Save as on images and download chips

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -15,6 +15,7 @@ import icon from "../../resources/icon.png?asset";
 import type { Attachment } from "../shared/attachments";
 import { stageAttachment, clearStagedAttachments } from "./attachment-staging";
 import { discoverProviderModels } from "./model-discovery";
+import { readMediaAsDataUrl, saveMedia } from "./media";
 import {
   checkInstallStatus,
   verifyInstall,
@@ -852,6 +853,14 @@ function setupIPC(): void {
   ipcMain.handle("copy-to-clipboard", (_event, text: string) => {
     clipboard.writeText(typeof text === "string" ? text : "");
   });
+
+  // Media — render agent-generated images and save them to disk (#299).
+  ipcMain.handle("read-media-file", (_event, filePath: string) =>
+    readMediaAsDataUrl(filePath),
+  );
+  ipcMain.handle("save-media-file", (event, src: string, name: string) =>
+    saveMedia(src, name, BrowserWindow.fromWebContents(event.sender)),
+  );
 
   // Attachment staging — for pasted blobs that have no filesystem origin.
   ipcMain.handle(
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -865,6 +865,50 @@ function setupIPC(): void {
     mediaFileExists(filePath),
   );
 
+  // Native right-click menu for a rendered media element (#299): "Open"
+  // hands the file to the OS default handler (or a web URL to the browser),
+  // "Save as…" writes a copy elsewhere. Labels are passed in from the
+  // renderer so the menu honours the active UI locale.
+  ipcMain.on(
+    "show-media-menu",
+    (
+      event,
+      src: string,
+      name: string,
+      labels: { open: string; saveAs: string },
+    ) => {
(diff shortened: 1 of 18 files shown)


fathah:
conflict

fathah:
The parser logic, false-positive safety net, and IPC design are all solid. No correctness bugs — happy to merge with a few follow-up items noted below.

### Follow-up (post-merge)

**1. Memoize `parseMediaTokens` in `MessageRow`**
`parseMediaTokens(msg.content)` is called directly in the render body. `MessageRow` is wrapped in `memo` but still re-runs on any prop change (e.g. `isLoading` toggling). For long conversations this runs the full regex pipeline on every message on every render. Simple fix:

```tsx
const segments = useMemo(() => parseMediaTokens(msg.content), [msg.content]);
```

**2. Index-based keys are unstable during streaming**

```tsx
parseMediaTokens(msg.content).map((segment, i) => ... key={i} ...)
```

While a message is streaming, a `MEDIA:` token appearing mid-content will shift all subsequent segment indices, causing unnecessary re-mounts (and spurious `mediaFileExists` IPC calls). A key derived from the segment's start offset or `src` would be stable. Consider exposing `start` on `MediaSegment`.

**3. `describeImageSrc` hardcodes `isImage: true`**
`alt` written in markdown will hit the `img` override → `MediaImage` → `readMediaAsDataUrl` → `null` (PDF not in `MIME_BY_EXT`) → `⚠ Could not load file.pdf`. Low-probability edge case, but `describeImageSrc` should check `IMAGE_EXT` the same way `toToken` does so it routes to `DownloadChip` instead.

**4. Track the #301 context-menu conflict**
The PR description flags that whichever of #301/#303 lands second needs a one-line guard so the media menu wins on media elements. Please open a tracking issue before #301 merges.

Issue #285, opened 2026-05-20
SSH-mode Gateway Start collides with systemd hermes.service → infinite crash-loop


andreab67:
## Summary

When the desktop app is connected to a remote Hermes via **SSH tunnel mode** and the user clicks **Start** on the Gateway tab, `sshStartGateway` runs:

```ts
// src/main/ssh-remote.ts:1086
await sshExec(config, `nohup hermes gateway start > $HOME/.hermes/gateway.log 2>&1 &`);
```

This spawns a detached gateway process **outside any systemd unit's cgroup**. The process writes `~/.hermes/gateway.pid` and holds the lock.

If the VPS is provisioned with the standard `hermes.service` unit (gateway + API server, `Restart=on-failure`), that unit detects the lock on its next start attempt and exits 1:

```
❌ Gateway already running (PID 403160).
   Use 'hermes gateway restart' to replace it…
systemd[1]: hermes.service: Main process exited, code=exited, status=1/FAILURE
systemd[1]: hermes.service: Scheduled restart job, restart counter is at 43.
```

systemd then crash-loops the unit every ~10s indefinitely. The desktop's gateway probably appears "running" from the user's perspective (because it is — under their `nohup` orphan), so they have no signal that anything is wrong on the VPS until they happen to check `journalctl`.

## Repro

1. Provision a VPS with `hermes.service` enabled (the standard installer setup).
2. Connect from hermes-desktop in **SSH tunnel mode**.
3. Go to **Gateway** tab → click **Start**.
4. `ssh root@<vps> 'journalctl -u hermes.service -n 50 --no-pager'` → see the crash-loop.

Manual recovery on the VPS:

```bash
sudo systemctl stop hermes.service
sudo kill <rogue-pid> 2>/dev/null
rm -f /opt/hermes/.hermes/gateway.pid
sudo systemctl start hermes.service
```

## Why this matters

- The user has no in-app indication anything is wrong; the Gateway chip stays green.
- systemd's restart counter climbs forever, polluting journal logs and `systemctl status`.
- Any feature that depends on the *systemd-managed* gateway PID (rather than the orphan) sees the unit as "failed".
- Recovery requires SSH + sudo, which is the exact friction SSH-tunnel mode was meant to eliminate.

## Suggested fix

Detect a systemd-managed hermes.service on the remote and prefer it over `nohup`. Sketch:

```ts
async function sshHasSystemdHermesUnit(config: SshConfig): Promise<boolean> {
  try {
[...]

PR #304, opened 2026-05-21
Route SSH-mode gateway ops through systemd when hermes.service exists (#285)

pmos69:
Addresses #285.

## The bug

In SSH mode, `sshStartGateway` unconditionally ran:

```sh
nohup hermes gateway start > $HOME/.hermes/gateway.log 2>&1 &
```

That spawns a detached gateway **outside any systemd cgroup**. On a VPS provisioned with the standard `hermes.service` unit, the orphan grabs the gateway lock, so the unit's own start attempts exit 1 and systemd crash-loops it every ~10s indefinitely — with no in-app signal, since the orphan genuinely *is* running. `sshStopGateway` and `sshGatewayStatus` were equally systemd-blind, so the three operations disagreed on which gateway they were managing.

Full diagnosis and repro are in #285 (thanks @andreab67 for the detailed report).

## The fix

Detect a systemd `hermes.service` on the remote and route **start / stop / status** through systemd when one is present. Detection + action happen in a single shell `if`, so the decision is atomic and costs only one SSH round-trip — no extra calls, no client-side caching.

- **start / stop** — try `sudo -n systemctl …` first, then a direct `systemctl …` (the SSH user may be root).
- **No `nohup` fallback once systemd is detected.** This is deliberate, and differs from the fix sketch in the issue: if `systemctl` can't be reached (non-root, no passwordless sudo) the start command does *nothing*. An unmanaged orphan that crash-loops the unit is worse than a gateway that simply didn't start — and the status check then honestly reports it down. `rm -f gateway.pid && nohup …` would also risk two gateways racing the API port.
- **Hosts without systemd** keep the original `nohup` / pidfile path unchanged — a missing `systemctl` makes the detection test fail, so they fall straight through.
- `status` reports the unit's `is-active` state when systemd-managed, the pidfile liveness check otherwise.

The shell command strings are built by exported helpers (`buildGateway{Start,Stop,Status}Command`) so the systemd-vs-plain branching is unit tested without a live host.

## Testing

- [x] `npm run typecheck` — clean
- [x] Full test suite — 492 passed / 3 skipped; 4 new tests in `tests/ssh-remote.test.ts` cover the command-string branching (notably: the `nohup` fallback is asserted *absent* from the systemd branch)
- [ ] **Not yet verified against a real systemd VPS.** I don't have an Ubuntu host with `hermes.service` installed. @andreab67 — you have the exact repro environment; could you check this branch resolves the crash-loop? Draft until then.

## Follow-up (not in this PR)

The renderer companion noted in #285 — surfacing systemd unit status (active / failed / not-installed) in `screens/Gateway/*` when SSH mode is active — is left for a separate change to keep this PR focused on the lifecycle collision.

### DIFF
--- a/src/main/ssh-remote.ts
+++ b/src/main/ssh-remote.ts
@@ -1329,17 +1329,92 @@ export async function sshDeleteProfile(
 }
 
 // ── Gateway ───────────────────────────────────────────────────────────────────
+//
+// In SSH mode the remote gateway may be owned by a systemd `hermes.service`
+// unit — the standard VPS installer sets this up. Starting our own detached
+// `nohup` gateway then strands that unit in a restart crash-loop (issue
+// #285). Each operation below therefore asks the remote, in a single shell
+// `if`, whether such a unit is installed and routes the request through
+// systemd when it is — one SSH round-trip, atomic decision. The command
+// strings are built by the exported helpers below so they can be unit
+// tested without a live host.
+
+/**
+ * Shell test that succeeds when a systemd `hermes.service` unit file is
+ * installed on the remote. Safe on hosts without systemd: a missing
+ * `systemctl` yields empty output, so the test simply fails and callers
+ * fall back to the plain (`nohup` / pidfile) path.
+ */
+const SYSTEMD_HERMES_UNIT_TEST =
+  "systemctl list-unit-files hermes.service 2>/dev/null | " +
+  "grep -q '^hermes\\.service'";
+
+/**
+ * Command to start the remote gateway (issue #285). When a systemd
+ * `hermes.service` exists it owns the lifecycle, so the request is handed
+ * to systemd — `hermes.service` is a system unit, so `sudo` is tried first,
+ * then a direct call for when the SSH user is root. If neither works the
+ * command does nothing on purpose: an unmanaged `nohup` orphan that
+ * crash-loops the systemd unit is worse than a gateway that simply did not
+ * start (the status check will then report it as down). The detached
+ * `nohup` start is used only when there is no unit to collide with.
+ */
+export function buildGatewayStartCommand(): string {
+  return (
+    `if ${SYSTEMD_HERMES_UNIT_TEST}; then ` +
+    `sudo -n systemctl start hermes.service 2>/dev/null || ` +
+    `systemctl start hermes.service 2>/dev/null || true; ` +
+    `else ` +
+    `(nohup hermes gateway start > $HOME/.hermes/gateway.log 2>&1 &); ` +
+    `fi`
+  );
+}
+
+/**
+ * Command to stop the remote gateway (issue #285). Routed through systemd
+ * when a `hermes.service` unit exists, so the unit is left cleanly inactive
+ * rather than the desktop killing a process systemd would just restart;
+ * otherwise it falls back to `hermes gateway stop` and, last resort, the
+ * recorded pid.
+ */
+export function buildGatewayStopCommand(): string {
+  return (
+    `if ${SYSTEMD_HERMES_UNIT_TEST}; then ` +
+    `sudo -n systemctl stop hermes.service 2>/dev/null || ` +
+    `systemctl stop hermes.service 2>/dev/null || true; ` +
+    `else ` +
+    `hermes gateway stop 2>/dev/null || ` +
+    `(if [ -f $HOME/.hermes/gateway.pid ]; then ` +
+    `pid=$(python3 -c "import json; d=json.load(open('$HOME/.hermes/gateway.pid')); print(d['pid'] if isinstance(d,dict) else d)" 2>/dev/null); ` +
+    `[ -n "$pid" ] && kill $pid 2>/dev/null; fi); true; ` +
+    `fi`
+  );
+}
+
+/**
+ * Command to report remote gateway state (issue #285). For a systemd-managed
+ * gateway this is the unit's `is-active` state (`active` when up); otherwise
+ * it is a liveness check on the recorded pid. Prints `active` or `running`
+ * when up, anything else when not.
+ */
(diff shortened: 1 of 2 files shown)


andreab67:
@pmos69 thanks for picking this up — quick interim report while I get a window to point this at the live VPS:

**Tested so far (Windows host, no remote):**
- Checked out `pmos69/fix/ssh-gateway-systemd` clean on top of `main`.
- `npm install` → succeeds. One non-fatal warning worth flagging: `better-sqlite3@12.8.0` declares engines `20.x || 22.x || 23.x || 24.x || 25.x`, and I'm on Node v26.1.0 → `EBADENGINE`. `@electron/rebuild` still rebuilt it cleanly against Electron 39, so it's cosmetic for now, but the engines field is going to bite anyone on the Node 26 train until better-sqlite3 ships a bump. Not your PR's problem — just noting it.
- `npm run dev` → main/preload/renderer all build, Electron launches.

**Not yet exercised (the part that actually matters for #285):**
- Connecting the desktop to the hermes VPS in SSH tunnel mode and exercising gateway start / stop / status against the live `hermes.service`. That's the only path that proves the systemd routing actually resolves the crash-loop the issue describes — I'll run it as soon as I'm back in front of the VPS and follow up here with the unit state before/after, whether `sudo -n systemctl` works for my SSH user (it doesn't have NOPASSWD, so I expect the direct `systemctl` branch to take over — useful signal for the PR), and whether `sshGatewayStatus` now agrees with `systemctl is-active`.

Leaving as Draft on your side is the right call until that bake is done. Flagging this thread so I don't lose it.

andreab67:
Real-host findings now in — exercised against my VPS, which is the same one that produced the crash-loop in #285. SSHed in as the `hermes` service user (what the desktop does in SSH tunnel mode) and ran your exported `buildGateway{Start,Stop,Status}Command` strings verbatim.

### Setup
- `hermes.service` is enabled + running, drop-ins present (`api-server-key.conf`, `drain-timeout.conf`, `override.conf`), `MainPID=445621`, `ActiveEnterTimestamp=Wed 2026-05-20 16:45:45 MDT` going in.
- SSH user is `hermes` (uid 1001). Key auth.

### Detection
- `systemctl list-unit-files hermes.service | grep -q '^hermes\.service'` → matches. `SYSTEMD_HERMES_UNIT_TEST` correctly takes the `if-then` arm on this host.

### Status (`buildGatewayStatusCommand`)
- Returns `active`. `sshGatewayStatus` parses that as `true`. ✅

### Auth-path probe (this is the interesting part)
As the `hermes` user, **both** branches of the start/stop chain fail:

| Branch | Result |
| --- | --- |
| `sudo -n systemctl ...` | `sudo: a password is required` (exit 1) — no NOPASSWD entry for `hermes` |
| plain `systemctl start hermes.service` | `Failed to start hermes.service: Interactive authentication required.` (polkit, exit 1) |
| plain `systemctl is-active hermes.service` | `active` (exit 0) — reads are allowed without auth |

So on this VPS the start/stop commands fall all the way through to `|| true` and **silently no-op**. Which is exactly the "do nothing on purpose" behavior you called out in the PR description — confirming that trade-off is real and observable, not just theoretical.

### Stop (`buildGatewayStopCommand`)
- Exit 0 (via `|| true`). After running: `MainPID=445621`, `ActiveEnterTimestamp` unchanged, `is-active=active`, `NRestarts=0`. Service untouched. ✅

### Start (`buildGatewayStartCommand`)
- Exit 0 (via `|| true`). Critically: `pgrep -af "hermes gateway"` shows only the one systemd-managed PID 445621 under `0::/system.slice/hermes.service` — **no orphan `nohup`'d gateway spawned**, no extra `python3 .../hermes gateway` outside the unit's cgroup. `NRestarts` still 0. The old `nohup hermes gateway start &` path that caused the #285 crash-loop is provably not executed. ✅

### Verdict
The fix does what it sets out to do on the exact host that exhibited the bug — it neutralizes the crash-loop without needing to actually drive the unit. The acknowledged trade-off (Start/Stop buttons effectively become no-ops on hosts where the SSH user has neither NOPASSWD sudo nor a polkit grant for the unit) is the right call — "service correctly running, button does nothing" is a much better failure mode than "service crash-looping forever, button doesn't visibly help."

I'd say this is ready to come out of Draft when you're ready.

### Small follow-up suggestion (separate PR, as you noted)
Since `is-active` works without auth but `start`/`stop` need it, the renderer companion you mentioned for #285 could surface that asymmetry — e.g. when the unit is detected, show the systemd state as the truth and, on Start/Stop click failure, hint "this SSH user cannot manage the unit; run `sudo systemctl start hermes` from a shell." Avoids the silent no-op being confusing. Out of scope here.

pmos69:
Thanks @andreab67 — that's exactly the test this needed, and on the very host that produced the #285 crash-loop. The auth-path probe is especially useful: confirming that both `sudo -n systemctl` and plain `systemctl start/stop` fail for the `hermes` SSH user (no NOPASSWD, polkit needs interactive auth) while `is-active` reads fine — and that the chain therefore falls through to `|| true` and no-ops without spawning an orphan — is precisely the behaviour the PR aims for. "Service correctly running, button does nothing" beating "service crash-looping forever" is the trade-off, and it's good to see it observed rather than assumed.

I've taken it out of Draft — it's ready for review.

On the follow-up: noted, and I agree it's the right next step. The renderer companion from #285 should surface that asymmetry — when a systemd unit is detected, treat `is-active` as the source of truth, and on a Start/Stop no-op hint that this SSH user can't manage the unit (e.g. "run `sudo systemctl start hermes` from a shell"). I'll keep that as a separate PR so this one stays focused on neutralising the crash-loop.

The `better-sqlite3` / Node 26 `EBADENGINE` note is also appreciated — agreed it's unrelated to this PR; worth its own issue if it starts biting people.

PR #305, opened 2026-05-21
SSH tunnel: probe venv (un-dotted) and ~/.local/bin for remote hermes CLI (closes #284)

### LINKED ISSUE #284: SSH tunnel: hermes probe only checks `.venv`, misses installs using `venv`
## Summary

In SSH tunnel mode, the desktop probes for the remote `hermes` CLI at three hardcoded paths in `src/main/ssh-remote.ts` (around lines 1604–1608):

```ts
const candidates = [
  "$HOME/hermes-agent/.venv/bin/hermes",
  "$HOME/.hermes/hermes-agent/.venv/bin/hermes",
  "/opt/hermes/hermes-agent/.venv/bin/hermes",
];
```

All three use `.venv` (dotted). However, installs that create the virtualenv as `venv` (no dot) are not detected, even when `hermes` is fully working from an interactive shell.

When this happens, the Settings → Debug Dump shows:

> `SSH dump failed: ERR: hermes CLI not found on remote PATH or in any known venv location`

And the Kanban tab shows the misleading message:

> "Kanban requires a local Hermes install or SSH tunnel mode. Plain remote (HTTP + API key) mode does not yet expose the kanban API."

…even though SSH tunnel mode is actually selected and the tunnel itself works fine. (The Kanban tab's copy might be worth tightening too — it implies the user is in the wrong mode when the real problem is the remote probe.)

## Reproduction

On a remote host where the install lives at `~/.hermes/hermes-agent/venv/bin/hermes` (i.e. `venv`, not `.venv`):

1. Set up SSH Tunnel mode in Desktop Settings pointing at that host.
2. Click **Debug Dump** → see the error above.
3. Click **Kanban** → see the "requires SSH tunnel" message.

`ssh user@host '~/.hermes/hermes-agent/venv/bin/hermes --version'` works fine, confirming the install is healthy.

## Why this happens

The non-interactive SSH command path used by the desktop doesn't source `~/.bashrc` / `~/.profile`, so any `PATH` additions (e.g. `~/.local/bin`) made there are invisible. The `command -v hermes` fallback therefore fails, and the hardcoded probe list never matches because `.venv` ≠ `venv`.

## Workaround

Symlink `.venv` to the actual venv on the remote:

```bash
ln -sf ~/.hermes/hermes-agent/venv ~/.hermes/hermes-agent/.venv
```

After this, the probe hits path #2 and everything (Kanban, Debug Dump, Run Diagnosis) works.

## Suggested fix

Add the non-dotted variants to the candidate list in `buildRemoteHermesCmd`:

```ts
const candidates = [
  "$HOME/hermes-agent/.venv/bin/hermes",
  "$HOME/hermes-agent/venv/bin/hermes",
  "$HOME/.hermes/hermes-agent/.venv/bin/hermes",
  "$HOME/.hermes/hermes-agent/venv/bin/hermes",
  "/opt/hermes/hermes-agent/.venv/bin/hermes",
  "/opt/hermes/hermes-agent/venv/bin/hermes",
];
```

Alternatively, probe `~/.local/bin/hermes` as well, since many `pip install --user` / install-script flows put a wrapper there.

## Environment

- Desktop: v0.4.5 (macOS arm64)
- Remote: Ubuntu, Hermes Agent v0.14.0 (2026.5.16), Python 3.11.15
- Connection mode: SSH Tunnel

pmos69:
Closes #284.

## The bug

In SSH tunnel mode the desktop locates the remote `hermes` CLI via `buildRemoteHermesCmd`, which probed only `.venv` install paths:

```ts
const candidates = [
  "$HOME/hermes-agent/.venv/bin/hermes",
  "$HOME/.hermes/hermes-agent/.venv/bin/hermes",
  "/opt/hermes/hermes-agent/.venv/bin/hermes",
];
```

A remote whose virtualenv directory is `venv` (no dot) is missed entirely — even when `hermes` runs fine from an interactive shell. The `command -v hermes` fallback doesn't save it either: the desktop's non-interactive SSH doesn't source `~/.profile` / `~/.bashrc`, so `PATH` additions made there are invisible.

Symptoms (from the report — thanks @daveonkels for the precise diagnosis):
- Settings → Debug Dump fails with *"hermes CLI not found on remote PATH or in any known venv location"*.
- The Kanban tab shows *"Kanban requires … SSH tunnel mode"* — misleading, since SSH tunnel mode **is** active and the tunnel works; the real failure is the probe.

## The fix

Probe each install base with **both** `.venv` and `venv`, and add `~/.local/bin/hermes` where `pip install --user` flows place a wrapper:

```ts
const candidates = [
  "$HOME/hermes-agent/.venv/bin/hermes",
  "$HOME/hermes-agent/venv/bin/hermes",
  "$HOME/.hermes/hermes-agent/.venv/bin/hermes",
  "$HOME/.hermes/hermes-agent/venv/bin/hermes",
  "/opt/hermes/hermes-agent/.venv/bin/hermes",
  "/opt/hermes/hermes-agent/venv/bin/hermes",
  "$HOME/.local/bin/hermes",
];
```

Each candidate is a cheap `[ -x path ]` test, so the extra entries cost nothing. `/usr/local/bin/hermes` is still deliberately **not** probed — that's the sudo-wrapper the function exists to bypass (see its header comment). `buildRemoteHermesCmd` is now exported so the probe list can be unit tested without a live host.

## Testing

- [x] `npm run typecheck` — clean
- [x] Full suite — 492 passed / 3 skipped; 4 new tests in `tests/ssh-remote.test.ts` assert both venv spellings + `~/.local/bin` are probed and `/usr/local/bin` is not
- [x] No VPS needed — `buildRemoteHermesCmd` is a pure command-string builder; the fix only widens an existing `[ -x … ]` probe list

## Not in this PR

The reporter also notes the Kanban tab's copy is misleading when the real fault is the remote probe. That's a separate renderer-copy change in `screens/Kanban/*` — left out to keep this fix focused; happy to follow up.

### DIFF
--- a/src/main/ssh-remote.ts
+++ b/src/main/ssh-remote.ts
@@ -1887,11 +1887,25 @@ export async function sshListCachedSessions(
 // Probe the well-known venv install paths first; fall back to bare `hermes`
 // on PATH only if none of those exist, preserving the old behavior for
 // non-installer deployments.
-function buildRemoteHermesCmd(args: string[], extraShell = ""): string {
+//
+// Each install base is probed with both `.venv` and `venv` — the venv
+// directory name is not fixed, and an install that uses the un-dotted
+// `venv` was otherwise invisible even when fully working (issue #284).
+// `~/.local/bin/hermes` is also probed, where `pip install --user` flows
+// place a wrapper. `command -v hermes` alone is not enough: the desktop's
+// non-interactive SSH does not source `~/.profile`/`~/.bashrc`, so any
+// PATH additions made there are not visible.
+//
+// Exported for unit testing the probe list without a live remote host.
+export function buildRemoteHermesCmd(args: string[], extraShell = ""): string {
   const candidates = [
     "$HOME/hermes-agent/.venv/bin/hermes",
+    "$HOME/hermes-agent/venv/bin/hermes",
     "$HOME/.hermes/hermes-agent/.venv/bin/hermes",
+    "$HOME/.hermes/hermes-agent/venv/bin/hermes",
     "/opt/hermes/hermes-agent/.venv/bin/hermes",
+    "/opt/hermes/hermes-agent/venv/bin/hermes",
+    "$HOME/.local/bin/hermes",
   ];
   const quotedArgs = args.map((a) => shellQuote(a)).join(" ");
   const probe = candidates
--- a/tests/ssh-remote.test.ts
+++ b/tests/ssh-remote.test.ts
@@ -4,6 +4,7 @@ import {
   buildGatewayStartCommand,
   buildGatewayStopCommand,
   buildGatewayStatusCommand,
+  buildRemoteHermesCmd,
 } from "../src/main/ssh-remote";
 import type { SshConfig } from "../src/main/ssh-tunnel";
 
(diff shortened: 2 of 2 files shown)


fathah:
conflict

PR #309, opened 2026-05-21
Add i18n to the Kanban screen (en + pt-PT)

pmos69:
## Summary

The Kanban screen was the one screen with **no internationalization** — `Kanban.tsx` didn't import `useI18n` and every user-facing string was hardcoded English, so it stayed English regardless of the selected language.

This adds a `kanban` translation namespace and routes the screen through it.

- New `kanban` namespace — `en/kanban.ts` (the source) and `pt-PT/kanban.ts`.
- Wired into the shared i18n `resources` for `en` and `pt-PT`.
- `Kanban.tsx` refactored — `useI18n` added, all ~57 strings resolved via `t()` (column/status labels, card tooltips, the create-task and new-board modals, the task-detail modal, prompts, confirmations, errors).
- Interpolated strings use `{{count}}` / `{{title}}` / `{{from}}`+`{{to}}` placeholders.

## Scope

- Only **en + pt-PT** are translated. The other six locales (es, id, ja, pt-BR, zh-CN, zh-TW) **fall back to English** via the existing `t()` resolver (`translated ?? fallback ?? key`) — nothing breaks; they can be added in follow-up PRs. This is deliberate: no unreviewed machine translations are shipped.
- The pt-PT translation was reviewed by a native speaker.
- Compact codes are intentionally **not** localized — priority tags (`P0`/`P1`/`P2`) and relative-age units (`5s`/`3m`/`2h`/`4d`).
- Cancel buttons reuse the existing `common.cancel` key.

## Test plan

- [x] Full test suite green (488 passed / 3 skipped)
- [x] i18n resolves: `kanban` namespace loads for en + pt-PT; missing locales fall back to English

### DIFF
--- a/src/renderer/src/screens/Kanban/Kanban.tsx
+++ b/src/renderer/src/screens/Kanban/Kanban.tsx
@@ -11,6 +11,7 @@ import {
   RotateCcw,
   Sparkles,
 } from "../../assets/icons";
+import { useI18n } from "../../components/useI18n";
 
 interface KanbanProps {
   profile?: string;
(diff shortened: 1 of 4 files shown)


fathah:
conflict

Issue #256, opened 2026-05-18
Workspace/Office does not inherit locale or default Hermes model configuration


chnczu-jpg:
Hi, I found two issues in Hermes Desktop v0.4.3 on macOS.

1. UI language is not fully applied to Workspace / Office

After switching the desktop app language to Chinese, the main Hermes Desktop UI changes correctly, but the Workspace / Office / Claw3D interface remains mostly English.

It looks like the main desktop renderer and `~/.hermes/hermes-office` are two separate frontends. The desktop app locale does not propagate into the Office frontend, and many Office strings appear to be hardcoded in English instead of using the same i18n/locale system.

Expected:
When the desktop language is set to Chinese, Workspace / Office should also use Chinese UI strings.

Actual:
Main app UI is Chinese, but Office still shows English labels such as:
- Welcome to Claw3D
- Remote gateway
- Building Directory
- Agent Event Console
- New session
- Attach / Send
- Marketplace / Analytics
- etc.

2. Workspace / Office does not inherit Hermes default model/provider

My normal Hermes chat is configured to use Grok, e.g.:

model.default: grok-4.3
provider: xai-oauth

But the Office runtime adapter defaults to `hermes` / `hermes-agent` instead of using the same configured default model.

I found that `~/.hermes/hermes-office/.env` contains:

HERMES_MODEL=hermes

and `server/hermes-gateway-adapter.js` also falls back to `hermes`.

Expected:
Office / Workspace should inherit the main Hermes config model by default, or at least expose a clear setting to sync with Hermes Desktop’s selected/default model.

Actual:
Office shows/uses `hermes` or `hermes-agent`, while the main chat uses Grok. This can lead to confusing failures such as AggregateError or model mismatch.

Suggested fix:
- Share locale/i18n between Hermes Desktop and Office frontend, or pass desktop locale into Office startup.
- Replace hardcoded English Office strings with translation resources.
- Let `hermes-office` read `~/.hermes/config.yaml` model.default when HERMES_MODEL is not explicitly set.
- Avoid shipping `.env` with `HERMES_MODEL=hermes` as a hard default.
- In the Office model list, prefer the configured Hermes default model over the generic `hermes-agent` entry.

pmos69:
Thanks for the detailed write-up — both problems are real, and they split cleanly by which codebase owns the fix.

[...]

PR #312, opened 2026-05-22
Start Hermes Office with the desktop's configured model (#256)

pmos69:
Addresses the **model half** of #256.

## The bug

`writeClaw3dSettings()` in `src/main/claw3d.ts` generates the `hermes-office/.env` and **hardcoded the model**:

```js
`HERMES_MODEL=hermes`,
```

So Hermes Office always ran a generic `hermes` agent, ignoring whatever model the desktop chat is configured with. When the desktop is set to (say) Grok or a Codex model, Office silently uses something else — which the reporter notes can surface as `AggregateError` / model-mismatch failures.

## The fix

- **Write the configured model.** `writeClaw3dSettings()` now resolves the model via `getModelConfig()` and writes it into the Office `.env`. Falls back to `hermes` only when no model is configured.
- **Refresh the `.env` on start.** Previously it was written only at first setup / port change / wsUrl change — so an existing Office install kept whatever model was frozen in at setup time, forever. `startAll()` now rewrites the `.env` before launching the processes, so existing installs pick up the right model on the next Office start, and a later desktop model change is reflected on the next Office restart.
- **Testable.** The `.env` builder is extracted into an exported `buildOfficeEnv()` with unit tests covering the `HERMES_MODEL` handling.

## Scope

This is the **model** half of #256 only. The other half — Office UI not following the desktop language — belongs to the separate `fathah/hermes-office` frontend (it has no i18n system); the desktop can't translate it. That's flagged on the issue.

## Test plan

- [x] `tests/office-env.test.ts` (3 new) + existing `claw3d` tests pass
- [x] Live: with the desktop model set to `gpt-5.5`, Stop→Start Office — `~/.hermes/hermes-office/.env` rewrites to `HERMES_MODEL=gpt-5.5`, and Office's model selector then shows `hermes/gpt-5.5` instead of the generic `hermes`

### DIFF
--- a/src/main/claw3d.ts
+++ b/src/main/claw3d.ts
@@ -11,7 +11,11 @@ import { homedir } from "os";
 import { createConnection } from "net";
 import { getEnhancedPath, HERMES_HOME } from "./installer";
 import { stripAnsi, safeWriteFile } from "./utils";
-import { getApiServerKey, getConnectionConfig } from "./config";
+import {
+  getApiServerKey,
+  getConnectionConfig,
+  getModelConfig,
+} from "./config";
 import http from "http";
 
 const HERMES_OFFICE_REPO = "https://github.com/fathah/hermes-office";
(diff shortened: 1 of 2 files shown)


fathah:
conflict

Issue #298, opened 2026-05-21
Feature request: Right-click context menu for selected text (Copy/Paste)


dido85-ai:
Please add a right-click context menu (dropdown) in the TUI that appears when text is selected:

Right-click on selected text should show a dropdown menu with:
- Copy
- Paste
- Select All
- Copy All (entire chat content)

This is the standard Windows UX pattern for desktop apps. Currently the TUI has no mouse interaction support for copy/paste, which makes it difficult to use on Windows where users expect this behavior.

Shift+right-click or Ctrl+right-click could also work if there are technical limitations with the default right-click in Ink/React.

Deepseek-v4-flash

Thanks!

Issue #27, opened 2026-04-28
Feature Request: Add Image Attachment, File Browser, and Context Folder Support to Hermes Desktop


creativef:
Hermes Desktop already wraps Hermes Agent, which has full vision and file-handling capabilities. This request is about exposing those features in the UI.

1. Image Attachment in Chat

Problem: Users can't attach images to chat messages.

Proposed Solution: Add a "+" button or drag-and-drop zone in the chat input bar. When an image is attached:

Encode it (base64 or file path)

Pass it to the Hermes Agent with the appropriate vision flag

Display a thumbnail in the message composer

Priority: High — lowest-hanging fruit, immediately useful.

2. File Browser / Attachment Panel

Problem: No way to give the AI access to files or folders.

Proposed Solution: Add a sidebar panel (collapsible) where users can:

Drag and drop files or folders

Browse and select files from the filesystem

The AI can then read, analyze, or reference these files

Priority: Medium — enables code analysis, document review, etc.

3. Context Folder (Working Directory)

Problem: The AI doesn't have persistent access to a project directory.

Proposed Solution: Allow users to set a "context folder" in settings (similar to Claude Desktop). This folder is mounted/accessible to the AI for all conversations, enabling:

Project-aware code completion

File creation and modification

Cross-file analysis

Priority: Medium — essential for developer workflow.

Implementation Notes:

Modify the React renderer to handle attachment UI

Pass attachments through the IPC bridge to the main process

The main process spawns Hermes with --vision or similar flags for image input

For file/folder access, use Node.js fs APIs or spawn Hermes with the appropriate working directory

This is a really great project. I, for one, would love to see the continued development of it. It has given people who don't have the skillset to deploy and run Hermes either on a VPS or their local Docker on their machine, or know what a terminal is,  so I appreciate your work. Thank you.

creativef:
I built this out myself, where you can attach files to the chat or folder for the ai to read.

jadams777:
+1 for adding file attachments.

pmos69:
https://github.com/fathah/hermes-desktop/pull/258 should close this

PR #315, opened 2026-05-22
feat(chat): queue follow-up messages typed while the agent is busy

shochaso:
## Problem

While a turn is in flight, the chat input drops anything you submit — `handleSend` early-returns on `isLoading`, so pressing Enter mid-run does nothing. There is no way to line up a follow-up; you have to wait for the agent to finish before you can type your next instruction.

## What this does

Adds a **client-side follow-up queue**:

- **`ChatInput`** stays usable while the agent runs. Enter / the send button now **enqueue** the message instead of dropping it. While busy, the send button renders as a secondary (outlined) "queue" button next to Stop, with a `Queue — sends after the current turn finishes` tooltip.
- **`Chat`** drains the queue one item per turn: when a turn finishes, the next queued message sends automatically as a fresh turn. Multiple follow-ups chain until the queue empties.
- A **`{n} queued for next turn`** indicator above the input shows pending follow-ups.
- **Stop** (abort) clears the queue — stop means stop everything.
- **New chat / clear** also clears the queue.

The queue lives entirely in the renderer. No gateway, IPC, or protocol changes — it works in both local and remote modes.

## Why client-side

The desktop app talks to the agent over the stateless OpenAI-compatible `/v1/chat/completions` endpoint, which has no mid-run steering primitive. A renderer-side queue delivers the "type your next prompt while it works" UX without needing a new transport. Delivery is next-turn rather than mid-stream injection.

## Notes

- New i18n keys (`chat.queueForNext`, `chat.queued`) added to `en` and `ja`; other locales fall back to `en`.
- `chat.queued` uses i18next `{{count}}` interpolation.

## Testing

- `npm run typecheck` — pass
- `npm run lint` — no new errors/warnings from these changes
- `npm test` — 488 passed / 3 skipped, 39 files
- `npm run build:mac` — packages cleanly

### DIFF
--- a/src/renderer/src/assets/main.css
+++ b/src/renderer/src/assets/main.css
@@ -1657,6 +1657,29 @@ body {
   background: #dc2626;
 }
 
+/* Send button shown while the agent is busy — queues a follow-up rather
+   than sending immediately, so it reads as secondary, not a primary send. */
+.chat-queue-btn {
+  background: transparent;
+  border: 1px solid var(--accent);
+  color: var(--accent);
+}
+
+.chat-queue-btn:hover:not(:disabled) {
+  background: var(--accent);
+  color: #fff;
+}
+
+/* "N queued for next turn" hint above the input. */
+.chat-queue-indicator {
+  font-size: 12px;
+  color: var(--text-secondary);
+  padding: 4px 10px;
+  margin-bottom: 6px;
+  background: var(--accent-subtle);
+  border-radius: var(--radius-md);
+}
+
 .chat-btw-btn {
   width: 34px;
   height: 34px;
--- a/src/renderer/src/screens/Chat/Chat.tsx
+++ b/src/renderer/src/screens/Chat/Chat.tsx
@@ -12,8 +12,14 @@ import { useFastMode } from "./hooks/useFastMode";
 import { useLocalCommands } from "./hooks/useLocalCommands";
 import { useI18n } from "../../components/useI18n";
 import type { ChatMessage, UsageState } from "./types";
+import type { Attachment } from "../../../../shared/attachments";
 
 export type { ChatMessage } from "./types";
+
+interface QueuedMessage {
+  text: string;
+  attachments: Attachment[];
+}
 
 interface ChatProps {
   messages: ChatMessage[];
@@ -41,6 +47,23 @@ function Chat({
   const [remoteMode, setRemoteMode] = useState(false);
   const dragCounter = useRef(0);
   const chatInputRef = useRef<ChatInputHandle>(null);
+  // Follow-up messages typed while the agent is busy. They send automatically,
+  // one per turn, once the current turn finishes.
+  const queueRef = useRef<QueuedMessage[]>([]);
+  const [queuedCount, setQueuedCount] = useState(0);
+
+  const handleEnqueue = useCallback(
+    (text: string, attachments: Attachment[]) => {
+      queueRef.current.push({ text, attachments });
+      setQueuedCount(queueRef.current.length);
+    },
+    [],
+  );
+
+  const clearQueue = useCallback(() => {
+    queueRef.current = [];
+    setQueuedCount(0);
+  }, []);
 
   useEffect(() => {
     let cancelled = false;
@@ -115,7 +138,8 @@ function Chat({
     setHermesSessionId(null);
     setUsage(null);
     setToolProgress(null);
-  }, [isLoading, hermesSessionId, sessionId, setMessages]);
+    clearQueue();
+  }, [isLoading, hermesSessionId, sessionId, setMessages, clearQueue]);
 
   const localCommands = useLocalCommands({
     profile,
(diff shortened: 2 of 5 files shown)


fathah:
Thanks for the thorough writeup and the repro steps — this is a real bug and the `safeSend` approach is exactly right. The belt-and-suspenders pattern (`isDestroyed()` pre-check + `try/catch` for the race window) is the idiomatic Electron solution here, and aborting the upstream generation on the first dead-sender chunk is the correct behaviour.

A couple of things to fix before this can merge:

**1. `contextFolder` is dropped (regression)**

The PR removes `contextFolder` from both the `ipcMain.handle("send-message", ...)` parameter list and from the `sendMessage(...)` call. In `main`, `contextFolder` is the 7th argument and drives the working-directory system message (issue #27). Without it, context-folder-scoped conversations silently stop injecting the folder path into every request. Please restore both the handler parameter and the trailing argument:

```ts
// handler signature — restore contextFolder
async (event, message, profile?, resumeSessionId?, history?, attachments?, contextFolder?) => {

// sendMessage call — restore trailing arg
await sendMessage(message, { ...callbacks }, profile, resumeSessionId, history, attachments, contextFolder);
```

**2. `currentChatAbort` not nulled in the `onChunk` abort path**

When `safeSend` returns `false` in `onChunk` and `currentChatAbort()` is called, `currentChatAbort` itself is not set to `null`. `onDone`/`onError` both do `currentChatAbort = null` first — the `onChunk` path should too, before calling abort, to match that invariant.

```ts
onChunk: (chunk) => {
  fullResponse += chunk;
  if (!safeSend("chat-chunk", chunk)) {
    if (currentChatAbort) {
      currentChatAbort();
      currentChatAbort = null; // ← add this
    }
  }
},
```

Fix those two and this is good to go.

---

## PR #315 — feat(chat): queue follow-up messages typed while the agent is busy

The feature itself is well-conceived — a renderer-side queue is the right approach given the stateless `/v1/chat/completions` transport, and the drain-one-per-turn pattern is clean. Happy to merge this once a few things are addressed.

**1. `contextFolder` dropped (regression — must fix)**

`useChatActions` no longer receives or passes `contextFolder`, and `Chat.tsx` loses the `contextFolder` state, its reset effects, and the `ChatHeader` props (`contextFolder`, `showContextFolder`, `onPickFolder`, `onClearFolder`). This silently breaks context-folder-scoped conversations (issue #27). Please restore the full `contextFolder` wiring — it's unrelated to the queue feature and shouldn't have been touched.

**2. "Copy entire chat" and "Select All bubble" IPC listeners removed (regression — must fix)**

The fork's `Chat.tsx` drops the `buildChatTranscript` import, the `messagesRef` + `onContextMenuCopyChat` listener, and `onContextMenuSelectBubble` (issue #298). These are unrelated to this PR and need to be restored.

**3. i18n keys missing from `en/chat.ts` (regression — must fix)**

The fork's `en/chat.ts` is missing: `thinking`, `toolCall`, `toolResult`, `setContextFolder`, `contextFolderActive`, `removeContextFolder`. These are used in other components and their absence will cause fallback rendering or runtime key errors. Please add the two new keys (`queueForNext`, `queued`) without removing the existing ones.

**4. Queue not cleared on session switch**

`Chat.tsx` main has a `useEffect` on `[sessionId]` that resets `hermesSessionId` and `contextFolder`. The PR removes this effect. As a side effect, if the user switches to a different session while messages are queued, those items will drain into the wrong session. The `clearQueue()` call should be added to the `sessionId` reset effect.

**5. Drain effect depends on `handleSend` identity**

```ts
useEffect(() => {
  if (isLoading || queueRef.current.length === 0) return;
  const next = queueRef.current.shift();
  setQueuedCount(queueRef.current.length);
  if (next) void handleSend(next.text, next.attachments);
}, [isLoading, handleSend]);
```

`handleSend`'s identity changes when `hermesSessionId` updates (which happens right after a turn ends). This creates a window where the effect could re-run on an already-drained queue. Safer to read `handleSend` via a ref that's updated on every render, and only depend on `[isLoading]`.

**6. Slash commands mid-run are silently dropped**

`handleSlashSelect` in `ChatInput` still calls `onSubmit` directly, so slash commands submitted while the agent is busy are dropped rather than queued. This is probably intentional (slash commands like `/new` or `/clear` shouldn't be queued), but it's worth making explicit — either a comment or a deliberate guard.

Minor nits:
- `queued: "{{count}} queued for next turn"` is grammatically fine for 1 but reads oddly for many. Consider `"{{count}} message(s) queued"` or proper i18next plural forms.
- No CSS for `.chat-queue-indicator` / `.chat-queue-btn` — these will render unstyled unless they intentionally inherit from existing rules.

Items 1–3 are must-fix regressions. Items 4–5 are correctness issues worth addressing. 6 and the nits are up to you.

pmos69:
Thanks for this PR and for clearly describing the UX gap. The core idea here was valid: messages typed while the agent is busy should not be silently dropped.

This has now been implemented on `main` via b03ce43787b54527d61d61b4f83177b47170fea6, with the queue work rebuilt on top of the newer chat code so it preserves the context-folder wiring, copy-chat/select-bubble IPC listeners, existing i18n keys, and the newer session-switch behavior. A later follow-up on `main` also tightened the queued-send error path.

One behavior difference from this PR is intentional/acceptable: Stop cancels the active generation, but a queued follow-up can still start immediately afterward. That matches the behavior of Claude Desktop and supports the workflow where the user changes their mind, queues the next instruction, and stops the current response.

Since the feature is already covered on current `main` and this branch is now stale/conflicting against the newer chat implementation, I’m closing this as superseded rather than asking you to keep rebasing it. Appreciate the contribution; it helped identify and shape the final behavior.

PR #326, opened 2026-05-23
feat: add Claude Code CLI as OAuth-backed provider

jason-edstrom:
Adds `claude-code` as a selectable provider in the Models / Setup / Providers screens, mirroring the existing `openai-codex` pattern.

## Why

The hermes-agent backend already recognizes `claude-code` as an alias of the Anthropic provider ([plugins/model-providers/anthropic/__init__.py](https://github.com/NousResearch/hermes-agent/blob/main/plugins/model-providers/anthropic/__init__.py)) and reads credentials from `CLAUDE_CODE_OAUTH_TOKEN` / the OAuth credential pool. But hermes-desktop had no UI surface for it — users on the Claude Pro/Max subscription tier couldn't pick it without hand-editing `config.yaml`.

This wires it through the desktop UI the same way `openai-codex` is wired through, so picking "Claude Code CLI" in the provider dropdown is enough once the user has run `claude` locally to log in.

## Changes

- **`src/main/providers.ts`** — `claude-code` added to `PROVIDERS_WITHOUT_API_KEYS`, so the install gate doesn't ask for an API key.
- **`src/main/model-discovery.ts`** — `claude-code` added to `NON_DISCOVERABLE_PROVIDERS`. Anthropic's `/v1/models` needs `x-api-key`, which OAuth users don't have, so we skip discovery (matches `openai-codex` / `qwen-oauth` behavior).
- **`src/renderer/src/constants.ts`** — `claude-code` added to `PROVIDERS.options`, `PROVIDERS.labels`, and `PROVIDERS.setup` as an OAuth-backed entry (`needsKey: false`, no `envKey`, no `url`).
- **i18n** — `claudeCodeName` / `claudeCodeDesc` / `claudeCodeTag` strings added to `en`, `es`, `pt-BR`, `zh-CN` (matching the existing Codex coverage). Other locales fall back to English.
- **Tests** — `providers.test.ts`, `constants.test.ts`, `model-discovery.test.ts`, and `installer-utils.test.ts` extended to cover `claude-code` end-to-end.

## Verification

- `npx tsc --noEmit` — clean
- `npx vitest run` — 489 tests pass, 3 skipped (full suite)
- `npm run lint` — no new errors/warnings introduced (2 pre-existing warnings remain in `Install.tsx` and `Schedules.tsx`)

## User flow after merge

1. User installs Claude Code CLI: `npm install -g @anthropic-ai/claude-code`
2. User logs in once: `claude` (browser OAuth for Pro/Max)
3. Open hermes-desktop → Providers → pick **Claude Code CLI**
4. Done — no API key field appears, no install-gate nag.


fathah:
Conflict

pmos69:
Until June 15, claude -p calls count against your regular subscription usage limits — the big, generous interactive bucket.

Starting June 15, 2026, Agent SDK and claude -p usage no longer counts toward your plan's usage limits; those limits stay reserved for interactive use of Claude Code, Claude Cowork, and Claude. So an autonomous agent doing high-volume non-interactive -p calls moves from "shares the large interactive allowance" to "limited to a $20–$200 dollar credit at API rates, then either pays API rates or halts.
(API rates are estimated to be around 40x the subscription rates., which are heavily subsidized)

Furthermore, in case Anthropic starts banning usage that does not follow the TOS, what will  be banned? Users or the product they use? Anthropic has shown it uses both levers, through different mechanisms:

There are three distinct things that can get hit:

1. The usage path / credential (the primary lever, used most). This is what's happened in practice. On January 9, 2026, Anthropic deployed server-side safeguards that blocked subscription OAuth tokens from working outside the official Claude Code CLI. The token simply gets rejected with an error when used in an unauthorized tool. This doesn't touch your account or kill the tool — it just blocks that specific unauthorized use. Your subscription keeps working normally for legitimate use (interactive Claude Code, Claude.ai). [Openclaw](https://openclaw.rocks/blog/anthropic-oauth-ban)

**2. The product/project (via legal pressure). This is what happened to OpenCode — Anthropic sent legal requests to the maintainers demanding they strip out the subscription-OAuth integration, and the project complied by removing the auth plugin entirely. So a third-party tool can be targeted directly, but the mechanism is legal/cease-and-desist against whoever distributes it, not a technical action against you.**

3. The user account (the heaviest lever, used sparingly). Account suspension or ban is possible under the Usage Policy and ToS for repeated or egregious violations, and some coverage flagged "possible account action" as a risk for users who didn't migrate. But in the actual OAuth crackdown, the dominant approach was blocking the credential and pressuring the products — not mass-banning subscribers. When Anthropic enforced more broadly in April, it emailed affected users and offered a one-time credit equal to one month's subscription cost plus optional discounted usage bundles as compensation — i.e., it cut off the coverage, not the accounts

PR #327, opened 2026-05-23
feat: show tool calls, tool output, and reasoning in chat history (collapsible) (ref:#330)

jason-edstrom:
## Summary

Resumed sessions now show the **full** transcript — not just the chat bubbles. Tool calls, tool results, and assistant reasoning are surfaced inline as native `<details>` collapsibles, defaulting to closed so the chat view stays clean but everything is one click away.

Previously `getSessionMessages` selected only `id, role, content, timestamp` and filtered to `role IN ('user','assistant') AND content IS NOT NULL` — so all the rich agent state in the local DB was dropped on read.

## What's new in resumed sessions

- 🟣 **Thinking** — assistant reasoning (from `reasoning` / `reasoning_content` / `reasoning_details` columns)
- 🔵 **Tool call** — pretty-printed JSON arguments per `tool_calls` entry
- 🟢 **Tool result** — full `role='tool'` row content with bounded `max-height` scroll

Each row carries the tool name as a chip, a one-line summary in the collapsed header, and the full payload in the expanded body. Long tool outputs (think `read_file` dumps) scroll inside their own section instead of stretching the page.

### Main process — `src/main/sessions.ts`

`getSessionMessages` was rewritten to:

1. Widen the SELECT to include `tool_call_id, tool_calls, tool_name, reasoning, reasoning_content, reasoning_details`.
2. Include `role='tool'` rows.
3. Drop the `content IS NOT NULL` filter (tool-only assistant turns have null content).
4. Return `HistoryItem[]` — a discriminated union — instead of the old flat `SessionMessage[]`.

Three new **exported pure helpers** make this trivially unit-testable without booting better-sqlite3 (which is an Electron-only native module):

- `pickReasoning(row)` — priority: `reasoning` → `reasoning_content` → flatten `reasoning_details` (Anthropic/OpenRouter signed-block JSON; extracts `text` / `thinking` fields, drops signatures).
- `parseToolCalls(json)` — defensive JSON parse; bad rows yield `[]` so a malformed `tool_calls` cell never blocks history rendering.
- `expandRowsToHistory(rows)` — pure row → `HistoryItem[]` expansion with the correct emit order: `(reasoning → assistant_bubble → tool_call*) → tool_result`.

### SSH remote — `src/main/ssh-remote.ts`

The Python script we run over SSH for remote profiles got the same widening — it now mirrors `expandRowsToHistory` inline so SSH sessions get the same UI. `decodeContent` / `pickReasoning` / `parseToolCalls` are reimplemented in Python following the TS source as the canonical reference.

### Renderer

- **`src/renderer/src/screens/Chat/types.ts`** — `ChatMessage` becomes a discriminated union. Live messages keep their existing `{ id, role, content, attachments? }` shape (now `ChatBubbleMessage`); `reasoning` / `tool_call` / `tool_result` are new variants tagged by `kind`. Existing live-stream construction sites compile unchanged.
- **`src/renderer/src/screens/Chat/HistoryRow.tsx`** (new) — `ReasoningRow`, `ToolCallRow`, `ToolResultRow` share a single `CollapsibleSection` primitive built on `<details>`/`<summary>`, so default-closed and toggle behavior are native (no event juggling, accessibility for free).
- **`src/renderer/src/screens/Chat/MessageList.tsx`** — dispatches by `item.kind`. Bubbles still get the 'has content' filter (live placeholders stay hidden); history rows pass through unconditionally.
- **`src/renderer/src/screens/Layout/Layout.tsx`** — `handleResumeSession` adapts the new `HistoryItem[]` shape into renderer `ChatMessage` variants.

### CSS

New `.chat-history*` classes. Each variant gets a colored left border so users can visually scan a session without expanding anything: purple = reasoning, blue = tool_call, green = tool_result.

### i18n

`chat.thinking` / `chat.toolCall` / `chat.toolResult` added to `en`. Other locales fall back to English — happy to add proper translations on review feedback or in a follow-up.

## Tests

- **`tests/sessions-history-items.test.ts`** (new, **17 tests**) — covers the row → `HistoryItem` expansion: ordering, reasoning column priority, `reasoning_details` flattening, malformed `tool_calls` JSON, empty assistant content with tool calls, `session_meta` drops, user-row filtering.
- **`tests/sessions-delete-session.test.ts`** — DB schema bumped to match the production schema (the new columns are referenced by the widened SELECT).

Full results: **505 passed, 3 skipped (508 total)**, `tsc --noEmit` clean, no new ESLint warnings on changed files.

## Scope notes

This PR covers **history display only**. Live streaming of tool calls and reasoning into the chat view as they happen is a larger project — it'd require widening the IPC wire protocol in `useChatIPC.ts` and the streaming callbacks in `src/main/hermes.ts` — and is intentionally out of scope here. If maintainers want it, happy to do it as a follow-up PR.

## Screenshots / demo

Test by resuming a session that used the `terminal` or `read_file` tool — every tool call gets its own collapsed row above and below the assistant bubble, expandable for the full args / output.


jason-edstrom:
closes #330

PR #282, opened 2026-05-20
ci: add GitHub Actions workflow for typecheck + tests

pmos69:
## Summary

Adds a CI workflow that runs **typecheck + tests** on every PR and on pushes to `main`. Closes #278.

Until now nothing ran a PR's suite automatically — a PR's `MERGEABLE / CLEAN` badge only means *no git conflict*, not *the code compiles or tests pass*. As a result, broken changes reached `main` unnoticed:

- three failing tests merged in (a duplicate provider option + a test mock missing a method) — fixed in #277;
- a PR that imported a non-existent file (`en/kanban.ts`) sat looking mergeable because it had no conflict.

CI makes "does it compile / do tests pass" automatic and visible on every PR.

## What it runs

`.github/workflows/ci.yml` — on `pull_request` and `push` to `main`:

- `npm ci` (Node 22 — matches `release.yml`)
- `npm run typecheck` (node + web)
- `npm test` (Vitest)
- `npm run lint` — **non-gating** (`continue-on-error`). The codebase has a large prettier/line-ending warning backlog (0 errors); lint is surfaced as annotations but doesn't block merges, so the backlog can be cleared separately without holding up CI adoption.

Includes a `concurrency` group so a newer push supersedes an in-flight run.

## Verification

Already exercised end-to-end on a real `ubuntu-latest` runner (via a throwaway PR on my fork) — **all steps green**: checkout, Node setup, `npm ci` (incl. the `better-sqlite3` native rebuild), typecheck, test, lint. So this should go green on its own PR here too.

`main` is currently green (typecheck clean, 488 tests passing) — so adopting CI now won't immediately flag a red baseline.

## Note

A CI workflow was briefly attempted before in #78, but its author self-closed it within ~20 minutes (it ran while `main` had failing tests, so it went red on pre-existing breakage — not a maintainer rejection). With #277 having fixed those, CI now starts from a green baseline.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #332, opened 2026-05-23
Chat: narrow ChatMessage union at every .content / .attachments site (CI gate unblock)

pmos69:
## Why

PR #327 widened `ChatMessage` from a single bubble shape into a union of four variants:

```ts
export type ChatMessage =
  | ChatBubbleMessage
  | ReasoningMessage
  | ToolCallMessage
  | ToolResultMessage;
```

…but four downstream consumers were left reading `.content` / `.attachments` off the bare union. They typechecked locally under some toolchains by accident; **under the CI gate added in #282 they fail 9 TS2339 errors on every PR**, blocking the entire merge queue (verified by checking out `origin/main` at `f3105d3` and running `npm run typecheck`).

This is a CI-gate unblock; no runtime behaviour change.

## What

- **Hoist** the `isBubble` predicate that already lived inside `MessageList` to `types.ts` as the exported `isBubbleMessage` type guard, so every consumer narrows by the *same* rule.
- **`MessageRow`**: tighten the prop type from `ChatMessage` to `ChatBubbleMessage`. The MessageList parent already only dispatches bubbles here — every other variant has its own `*Row` component — so this just makes the existing contract explicit.
- **`MessageList`**: replace the `(msg as { kind?: string }).kind` discriminant + the four `as Extract<…>` casts with direct discriminated-union narrowing on `msg.kind`. TS narrows exhaustively; no casts left.
- **`useChatActions`**: filter the outgoing history with `isBubbleMessage` before mapping — the agent has its own reasoning / tool rows in its state and the bubble `.content` shape doesn't apply to them.
- **`useChatIPC`**: only chunk-append to the last message when it's a bubble. Live chunks never target a sub-row.
- **`transcriptUtils`**: filter to bubbles before building the user-visible transcript — reasoning / tool_call have no `.content`.

## Tests

- Extends `transcriptUtils.test.ts` with a mixed-kind input asserting the three sub-row variants are filtered out (`reasoning`, `tool_call`, `tool_result`). The existing 5 cases still pass unchanged.
- `isBubbleMessage` is exercised transitively by the new test (transcript builds correctly only if the filter narrows correctly).

## Verification

- `npm run typecheck` — clean (was 9 errors on `main`).
- Full suite — **562 passed / 3 skipped / 0 new failures**. The `locale-persistence.test.ts` "main process restarts" flake under full-suite load is unrelated and passes in isolation (documented previously on PR #321).
- Verified the diagnosis end-to-end: checked out `origin/main` at `f3105d3`, ran typecheck, got the exact 9 errors this PR resolves.

## Scope discipline

I deliberately did not change any other behaviour — no new UI for the reasoning / tool_call / tool_result sub-rows (those still flow through their existing `*Row` components added in #327), no signature changes to `sendMessage` or `onChatChunk`, no new exports beyond the one shared type guard.

### DIFF
--- a/src/renderer/src/screens/Chat/MessageList.tsx
+++ b/src/renderer/src/screens/Chat/MessageList.tsx
@@ -2,6 +2,7 @@ import { memo, useMemo } from "react";
 import { HermesAvatar, MessageRow } from "./MessageRow";
 import { ReasoningRow, ToolCallRow, ToolResultRow } from "./HistoryRow";
 import type { ChatMessage } from "./types";
+import { isBubbleMessage } from "./types";
 
 interface MessageListProps {
   messages: ChatMessage[];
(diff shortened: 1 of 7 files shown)


pmos69:
Closing as obsolete — the back-merge in 25d414a brought the Analytics commit (`380f30d`) and its per-file ChatMessage union narrowing onto `main`, so the four sites this PR was targeting (`MessageRow.tsx`, `useChatActions.ts`, `useChatIPC.ts`, `transcriptUtils.ts`) are already covered. Verified `npm run typecheck` on the current `origin/main` reports zero TS errors (the only remaining diagnostic is `Cannot find module 'posthog-js'` until a fresh `npm install` pulls the new dep from the lockfile).

Thanks for sorting out the back-merge, @fathah.

PR #335, opened 2026-05-23
feat(sidebar): Superset-style layout with collapsible sections and local Workspaces

shochaso:
## Motivation

The current sidebar is a flat 14-item nav with no live state (you can't tell at a glance whether the gateway is up, which sessions are running, etc.) and no concept of a "working folder" for the agent. This PR restructures the sidebar into a Superset/Linear-style layout with collapsible sections, live status dots, and a new local Workspaces feature so the agent can run inside a chosen folder.

## Sidebar restructure

- Profile switcher chip at the top (avatar + active profile + dropdown of `listProfiles()` — switches via `setActiveProfile`)
- Flat nav: **Office / Kanban / Chat**
- `+ New Workspace` action row
- Three collapsible sections, each with an icon, count badge, and chevron — collapse state persisted in `localStorage`:
  - **Workspaces** — list of registered local folders
  - **Sessions** — `listSessions(8, 0)` rendered inline with a live status dot (`endedAt === null` → pulsing green = running; otherwise hollow = idle) and per-row message count. Click resumes the session
  - **Configuration** — Profiles / Models / Providers / Skills / Persona / Memory / Tools / Schedules / Gateway. The Gateway row carries a live dot from `gatewayStatus()`
- **Settings + Help** pinned at the very bottom

CSS uses `display: grid; grid-template-rows: auto minmax(0,1fr) auto` on `.sidebar` so the scroll region shrinks correctly and the footer always stays pinned.

## New: local Workspaces

A "workspace" is a local folder the user registers via the `+ New Workspace` button. Folders are persisted to `desktop.json` under `workspaces` / `activeWorkspace`. Five new IPC channels (`list-workspaces` / `add-workspace` / `remove-workspace` / `get-active-workspace` / `set-active-workspace`).

The selected workspace is threaded through `sendMessage(... , workspace)`:

- **CLI path (`sendMessageViaCli`)** — uses the workspace as the spawn `cwd` (replaces the previous `cwd: HERMES_REPO`). The agent really does run in that folder.
- **API path (`sendMessageViaApi`, OpenAI-compatible `/v1/chat/completions`)** — the wire format has no working-directory field, so the workspace is surfaced as a leading system message: `"The user's active workspace directory is: <path>"`. This is a soft context hint, not a real `chdir`. Honest about the limitation; a deeper fix would need a gateway-side extension.

## Accessibility

- All controls are keyboard-operable (`Enter` / `Space` activate section headers, workspace rows, and session rows via a shared `activateOnKey` helper — same handler as `onClick`).
- `:focus-visible` ring on all sidebar controls.
- Profile dropdown dismisses on outside click and `Escape`.
- ARIA: `aria-haspopup` / `aria-expanded` on the profile chip, `aria-expanded` on section headers.

## i18n

New keys (`newWorkspace`, `sectionWorkspaces`, `sectionSessions`, `sectionConfiguration`, `help`) translated across all 8 locales (en / ja / es / id / pt-BR / pt-PT / zh-CN / zh-TW).

## Verification

- `npm run typecheck` — clean
- `npm run lint` — 0 errors (warnings unchanged from baseline)
- `npm test` — 488 passed / 3 skipped, including `tests/preload-api-surface.test.ts` which auto-verifies new methods exist in both `preload/index.ts` and `index.d.ts`
- `npx electron-vite build` — production build succeeds
- Manual: ran `npm run dev`, confirmed the full sidebar renders with live data (profile chip, 3 sections with counts and carets, sessions with green pulsing dots for in-flight runs, footer Settings + help). Keyboard-tested section collapse (`Tab` to header → `Enter` collapses, caret rotates ▼→▶).

## Files

19 files, +1015 / −99. New: `src/renderer/src/screens/Layout/Sidebar.tsx`. Touched: main process (`config.ts` / `hermes.ts` / `index.ts`), preload (`index.ts` / `index.d.ts`), renderer (`Layout.tsx` / `Chat.tsx` / `useChatActions.ts` / `assets/main.css` / `assets/icons/index.tsx`), 8 i18n navigation locales.


pmos69:
Solid restructure — the live status dots and the `grid-template-rows: auto minmax(0,1fr) auto` pattern for the pinned footer are clean wins.

One framing question worth pinning down before this lands, since yesterday's [v0.5.0](https://github.com/fathah/hermes-desktop/releases/tag/v0.5.0) shipped #292 (per-conversation `contextFolder` in the Chat screen): how do **Workspaces** and **`contextFolder`** relate?

I think they're actually two layers of the same idea, not competing concepts — same split Claude Desktop makes between Projects (chat tab) and per-session folders (Claude Code tab):

| Layer | Scope | Purpose |
|---|---|---|
| **Workspace** (this PR) | Per-user, "active" one at a time | Organisational container — *"I'm working on Project X"* |
| **`contextFolder`** (#292) | Per-conversation | Operational override — *"this conversation works in this specific folder"* |

The cleanest unification I can see:

- `contextFolder` defaults to `activeWorkspace.folder` when a conversation opens.
- User can still override per-conversation in the Chat header (small chip), e.g. to work in a subfolder.
- The `cwd` / system-message threading you already wire up reads from `contextFolder` instead of `activeWorkspace.folder` directly — same behaviour 99% of the time, and survives the override case for free.

That keeps both surfaces useful (Workspaces in the sidebar for *"where I'm camped this week"*, `contextFolder` in Chat for *"what this conversation specifically reads"*) without doubling up the folder-picker UX.

Heads-up that I've been prototyping a file browser pane attached to the conversation's context folder as a follow-up to #292. After your PR lands the cleaner home for it is probably as a **workspace-level** file panel (closer to Claude's Cowork tab semantically) — happy to fold my parked work into whichever shape this discussion settles on.

fathah:
conflict

PR #336, opened 2026-05-23
fix: quote remote Hermes SSH command script safely

sokratisg:
## Summary

- Quote the full remote `bash -c` script generated by `buildRemoteHermesCmd()` while preserving per-argument shell quoting.
- Preserve upstream's expanded Hermes CLI probe list for `.venv`, `venv`, `~/.local/bin/hermes`, and bare PATH fallback.
- Add SSH command-construction tests for multi-word titles, multiline markdown bodies, single quotes, existing `extraShell` redirects, and upstream's venv probe coverage.

## Root Cause

SSH tunnel mode wrapped the whole remote script in raw single quotes while each Hermes CLI argument was also single-quoted. User input containing spaces, newlines, or single quotes could terminate the outer script string and be interpreted as shell syntax. Multiline markdown bodies could then produce errors like `bash: line N: -: command not found`.

## Validation

- `npm run test -- tests/ssh-remote.test.ts` (17 tests passed)
- `npm run lint` exits 0 with 2 existing warnings
- `npm run typecheck`
- `npm run test` (57 files passed, 663 tests passed, 3 skipped)
- `npm run dev` built main/preload and started the renderer dev server at `http://localhost:5173/`; the command exited after `starting electron app...` in this headless session, so no manual UI interaction was performed
- Live SSH validation against a private remote Hermes test host through the rebased fixed command shape:
  - created, verified, and archived a temporary multiline-body task
  - final active task list had no leftover test tasks

### DIFF
--- a/src/main/ssh-remote.ts
+++ b/src/main/ssh-remote.ts
@@ -1911,7 +1911,8 @@ export function buildRemoteHermesCmd(args: string[], extraShell = ""): string {
   const probe = candidates
     .map((p) => `[ -x ${p} ] && exec ${p} ${quotedArgs}${extraShell}`)
     .join("; ");
-  return `bash -c '${probe}; command -v hermes >/dev/null && exec hermes ${quotedArgs}${extraShell}; echo "ERR: hermes CLI not found on remote PATH or in any known venv location" >&2; exit 1'`;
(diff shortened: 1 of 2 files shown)


fathah:
Conflict

PR #366, opened 2026-05-25
feat: improve local LLM discovery and Codex OAuth defaults

Logical-Leap:
## Summary

This improves Hermes Desktop support for local OpenAI-compatible runtimes and OAuth-backed model providers.

- Allow model discovery for local OpenAI-compatible providers without requiring a static API key when the configured base URL points at the local machine.
- Centralize local-provider and loopback URL detection in `provider-registry.ts` instead of inlining provider-specific or port-specific checks in model discovery.
- Add `GPT-5.5 Codex (OAuth)` as a seeded default model entry using the existing `openai-codex` provider path.
- Preserve existing behavior for remote/custom endpoints: non-loopback custom providers still require an explicit API key or matching env var.

## Why

Local LLM servers such as Ollama, LM Studio, llama.cpp, and vLLM commonly expose OpenAI-compatible `/v1/models` endpoints on user-configured local ports and do not require bearer tokens. Hermes Desktop previously returned `status="no-key"` before attempting discovery, so local model autocomplete/listing failed even though the endpoint was reachable.

OAuth providers are already represented in the app/provider list; seeding a Codex OAuth default model makes that path easier to select without confusing it with the static-key OpenAI provider.

## Implementation notes

- Loopback detection checks URL host identity, not a specific port, so it works with user-configured ports.
- IPv4 loopback is handled as the full `127.0.0.0/8` range; IPv6 loopback is handled as `::1`; `localhost` is also supported.
- The local no-key fallback only applies when both conditions are true:
  1. provider is a local OpenAI-compatible provider (`custom`, `ollama`, `lmstudio`, `vllm`, `llamacpp`)
  2. base URL resolves to the local machine
- Remote custom providers keep the existing `no-key` behavior.

## Test plan

- `npm run typecheck`
- `npx vitest run tests/model-discovery.test.ts tests/provider-registry.test.ts tests/oauth-model-discovery.test.ts`
- `npm run build`

### DIFF
--- a/src/main/default-models.ts
+++ b/src/main/default-models.ts
@@ -5,7 +5,7 @@
  * on fresh install. Format:
  *   { name: "Display Name", provider: "provider-key", model: "model-id", baseUrl: "" }
  *
- * Provider keys: openrouter, anthropic, openai, custom
+ * Provider keys: openrouter, anthropic, openai, openai-codex, custom
  * For openrouter models, use the full path (e.g. "anthropic/claude-sonnet-4-20250514")
  * For direct provider models, use the provider's model ID (e.g. "claude-sonnet-4-20250514")
  */
@@ -40,6 +40,14 @@ const DEFAULT_MODELS: DefaultModel[] = [
     provider: "openai",
     model: "gpt-4.1",
     baseUrl: "",
+  },
+
+  // ── OpenAI Codex OAuth / ChatGPT subscription ────────────────────────
+  {
+    name: "GPT-5.5 Codex (OAuth)",
+    provider: "openai-codex",
+    model: "gpt-5.5",
+    baseUrl: "https://chatgpt.com/backend-api/codex",
   },
 ];
 
--- a/src/main/model-discovery.ts
+++ b/src/main/model-discovery.ts
@@ -26,7 +26,11 @@ import {
 // PROVIDER_BASE_URLS lives in its own module so `config.ts` can use the
 // same lookup without pulling in this whole file (and triggering a
 // circular import via `model-discovery → config → ...`).
-import { PROVIDER_BASE_URLS } from "./provider-registry";
+import {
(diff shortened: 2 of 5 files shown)


fathah:
Conflict

Logical-Leap:
Merge conflict resolved and branch updated with latest main. Ready for review.

pmos69:
Thanks for this contribution. I walked through the current state of `main` plus the open local-provider work, and I think this PR has now been superseded by later changes.

The main pieces here are covered by these commits/PRs:

- `673d07a` (`feat: support OpenAI Codex provider setup`) added `openai-codex` as a first-class Desktop provider/setup option, including the no-static-key provider classification.
- `1ce6ece` (`feat(models): live model-discovery autocomplete for OpenAI-compatible providers (#265)`) added the general `/v1/models` discovery/autocomplete path for OpenAI-compatible providers.
- `ea333e3` (`feat(models): discover models for OAuth/subscription providers`) added OAuth/subscription provider model discovery, including the `openai-codex` path and the GPT-5.5 Codex model surfacing through discovery rather than a seeded default row.
- `c2e87cf` (`fix(providers): route built-in OpenAI-compat providers correctly end-to-end`) moved the built-in OpenAI-compatible provider routing/base URL behavior into the shared provider registry path.
- #402 / `ee8567b` (`expose local model providers`) covers the remaining local-provider behavior from this PR: LM Studio/Ollama/vLLM/llama.cpp are exposed as local providers and model discovery can run without a static key for those providers, plus loopback `custom` URLs. It also does this without sending a fake `Authorization: Bearer no-key-required` header when there is no real key.

So the useful intent of this PR has landed or is represented in the newer #402 path, and keeping this branch open now mostly risks reintroducing older/conflicting versions of the same logic.

Closing as superseded. Thanks again for pushing on this area; it helped identify the exact local-discovery gap that #402 now addresses.

Issue #367, opened 2026-05-25
# Nous Portal support is broken in v0.5.1 — no UI to add API key, dropdown selection silently fails, credential pool saves with wrong schema


iandabeast333:
# Nous Portal support is broken in v0.5.1 — no UI to add API key, dropdown selection silently fails, credential pool saves with wrong schema

## TL;DR

The Providers tab lists "Nous Portal" in the inference-provider dropdown, but offers **no working way for a user to authenticate it from the UI**:

- No **API Key** card for Nous Portal in the *LLM Providers* section.
- No **Sign in** card for Nous Portal in the *Subscription / OAuth Plans* section.
- The Credential Pool dropdown does include "Nous Portal", but when a key is added it writes a malformed schema that the gateway can't consume.
- Selecting "Nous Portal" from the inference provider dropdown writes `model.provider: "nous"` to `config.yaml` (the OAuth variant) — so the gateway proceeds to run, the user starts chatting, and only then sees `Hermes is not logged into Nous Portal. Run hermes model to re-authenticate.`

The end result is that a non-technical user can't get Nous Portal working through the desktop UI at all on Windows. The only fix is to drop into PowerShell and edit `.env` / run `hermes auth add nous` by hand.

## Environment

- **Hermes Desktop:** v0.5.1
- **Hermes Engine:** v0.14.0 (released 2026-05-16)
- **OS:** Windows 11
- **Python:** 3.11.15 (bundled with engine)
- **Profile home:** `C:\Users\<user>\Desktop\Hermes\profiles\<profile>` (HERMES_HOME points to a profile dir, not `~/.hermes`)

### 1. Missing "Nous Portal API Key" card in *LLM Providers* section

The engine plugin at `plugins/model-providers/nous/__init__.py` declares:

```python
nous = NousProfile(
    name="nous",
    aliases=("nous-portal", "nousresearch"),
    env_vars=("NOUS_API_KEY",),
    ...
)
```

`NOUS_API_KEY` is also referenced by `hermes_cli/doctor.py`, `hermes_cli/dump.py`, and several test files — it's a first-class env-var-based credential. The `Credential Pools` doc on the Hermes website (`/user-guide/features/credential-pools`) explicitly says:

> *"For providers that support both API keys and OAuth (Anthropic, Nous, Codex), the add flow asks which type."*

[...]

Issue #353, opened 2026-05-24
Warning: Unknown toolsets: hermes


jinjun:
Is the desktop version fully tested for Windows? The TUI version works so well but as soon as I launch the desktop app, I start to see all sorts of errors. 

1) 50% I'll see "Warning: Unknown toolsets: hermes" aftert I type in any chat msg;
2) another 50% I would see "Error: Session continuation requires API key authentication. Configure API_SERVER_KEY to enable this feature.". 
3) I can no longer tell if the agent is still running. But in most cases, normal answer starts to appear after the error msg.

Please take a look. Thank you!

pmos69:
Three separate things in this report — quick triage:

**(2) "Session continuation requires API key authentication"** — this is a known bug, already reported on #302 / #333 / #346 / #348 and now here. PR #337 fixes it and is merged into `main`, so it'll be in the next release. Until then, 30-second workaround:

1. Open `%LOCALAPPDATA%\hermes\config.yaml`, find the value after `  token:` under `api_server:` (a long random string).
2. Open or create `%LOCALAPPDATA%\hermes\.env`.
3. Add: `API_SERVER_KEY=<paste the token>`
4. Restart Hermes Agent.

Once #337 ships in a release, this workaround stops being needed.

**(1) "Warning: Unknown toolsets: hermes"** — genuinely new, and not coming from the desktop. The desktop only writes the `platform_toolsets.cli:` block; the warning is emitted by the Python hermes-agent CLI when it parses an unrecognised entry in the top-level `toolsets:` list of `config.yaml`. Could you paste the top-level `toolsets:` block from your `config.yaml`? It should look something like:

```yaml
toolsets:
- hermes-cli   # or hermes, or something else
```

That confirms whether it's a stale default entry from `hermes setup` (in which case the warning is cosmetic and safe to ignore until upstream cleans it up), or something else.

**(3) "Can no longer tell if the agent is still running"** — almost certainly a knock-on effect of (2) firing mid-stream. The auth error appears inline before the answer, which can leave the spinner/loading state ambiguous. Once (2) goes silent (workaround above, or next release), this should largely resolve on its own.

Thanks for the clear report — three problems cleanly separated is unusually helpful.

Sawyer007:
ask codex or claude to fix it like I did. :D

PR #369, opened 2026-05-26
fix(config): defensive consistency hardening — gateway key injection, migration, validation, audit

### LINKED ISSUE #423: [Help] Hermes-desktop refuses to continue chatting!
Hello,

I downloaded **hermes-desktop** and tried to use it with various models via a locally installed **Ollama** as a service provider. Eventually, I figured out that not all models are compatible, mainly due to lack of tools support and large enough token context size. Eventually, I settled on **"qwen3:4b"** and that seemed to work upon first try, with **Ollama** running in the background.

Upon trying to continue chatting, I received the following error:

> Error: Session continuation requires API key authentication. 
> Configure API_SERVER_KEY to enable this feature.

<hr> 

I've looked around and even asked **hermes** for an answer. It responded with:

Image

<hr> 

Problem: [console.hermes.ai](https://console.hermes.ai/) does not seem to be a working URL and 
the main domain is also for sale.

Image

<hr> 

**Question:** 
Since I am using Ollama locally with local models, how do I set the **"API_SERVER_KEY?"**

Thank you,
MB

NOTE: Will uninstall until problems are resolved.

pmos69:
## The bug

Multiple users on Telegram + Reddit have been hitting this on v0.5.1 since the day it shipped:

```
Error: Session continuation requires API key authentication.
Configure API_SERVER_KEY to enable this feature.
```

Reporters across recent threads:
Belanda, S S, Journey, Jen Jen, Josh Müller, Kursadk (SSH variant), plus Marat (related session proliferation symptom).

Fathah on the maintainers' Telegram (2026-05-25): *"session and API key related issues, still exists in the latest version too. + need to validate the model, provider and API Keys before hitting send button, so users doesn't get stuck in the error loop."*

## Root cause

Verified live via `scripts/repro-session-cont-error.js`.

The desktop's `getApiServerKey` reads `API_SERVER_KEY` from **six** candidate locations — top-level `config.yaml`, `.env`, and nested `api_server.token` (each per-profile and default). The upstream gateway's `APIServerAdapter` (`api_server.py:647`) reads only `api_server.extra.key` or `os.getenv("API_SERVER_KEY")` at startup, and `gateway/run.py:608-610` bridges only *top-level* config.yaml keys into env.

So when a user's key lives in the nested `api_server.token` location (which is what `hermes setup` writes, and where the existing precedence routes a fresh install), the desktop has it but the gateway never does. PR #357 — which made every fresh chat send `X-Hermes-Session-Id` — made this latent divergence user-visible on every send.

## The six-piece defense (+ Nous follow-up)

snap-369-chat-banner
snap-369-after-apply-fix

| # | Commit | Lines | Role |
|---|---|---:|---|
| 0 | `fix(gateway): inject resolved API_SERVER_KEY into gateway spawn env` | 33 | The bug fix — bridge the desktop's resolved value into the gateway's spawn env so `os.getenv("API_SERVER_KEY")` always sees what the desktop sees. |
| 2 | `refactor(shared): extract URL_KEY_MAP to src/shared/url-key-map.ts` | 177 | Single source of truth for URL → env var name (previously duplicated across 3 sites). |
| 1a | `feat(config): track source in API_SERVER_KEY resolution + helpers` | 269 | `resolveApiServerKeyWithSource()` + `maskKey()` + `appendConfigFixLog()`. Foundation for migration + audit. |
| 1b | `feat(config): migrate non-canonical API_SERVER_KEY to .env on read` | 213 | When `getApiServerKey` resolves from a non-`.env` source, copy the value to `.env` (additive, idempotent, audit-logged). |
| 3 | `fix(models): read back saved API key in Edit Model dialog` | 17 | Replace `setFormApiKey("")` with read-from-.env so the dialog shows reality. |
| 4 | `feat(chat): pre-send validation with disabled Send + inline reason` | 484 | New `validateChatReadiness` IPC. Disables Send + shows banner when active provider's key is missing. Fail-open — never false-blocks. |
| 5 | `feat(diagnose): startup config-health audit + auto-fix UI` | 1382 | `runConfigHealthCheck` + `autoFixIssue` IPC. Banner above chat + full panel in Settings → Diagnose with per-issue "Fix" buttons where it's safe to auto-apply. |
| — | `chore(i18n)` + `chore(test)` | 1364 | pt-PT translations + persisted CDP harness scripts. |
| **6** | **`feat(diagnose): detect missing Nous Portal credentials (issue #367)`** | **503** | **New — closes the silent-misconfiguration gap. Validation + audit now check `auth.json` (top-level `providers[<name>]` and `credential_pool[<name>]`) as a positive fallback after the env-key check, so Nous Portal in OAuth mode passes and Nous Portal with nothing configured at all blocks. Also benefits any other provider whose credentials live in `auth.json`.** |

## Verification

Six live-test scenarios using the new CDP harness (`ENABLE_CDP=1 npm run dev` + `node scripts/<name>.js`). All scripts persisted in `scripts/` with `scripts/README.md` documenting the workflow.

1. **Gateway env injection** — `node scripts/repro-session-cont-error.js` — 403 → 200 with the same bug-triggering config.
2. **Migration on read** — after running #1, `.env` is populated with `API_SERVER_KEY=…` and `~/.hermes/logs/config-fixes.log` has the migration entry.
3. **Edit Model API key read-back** — `node scripts/repro-edit-model-emptykey.js` — verdict 🔴 → ✅ (field shows saved key on reopen).
4. **Pre-send validation** — delete `DEEPSEEK_API_KEY` from .env + clear deepseek from auth.json while DeepSeek is active → Send disabled, banner shows the missing key + "Fix in Providers →".
5. **Startup health audit** — config in `api_server.token`-only state → ConfigHealthBanner appears with the issue; auto-fix populates .env; re-run shows clean.
6. **Nous credential detection (#367 Bug 4)** — `node scripts/repro-nous-detection-367.js` — provider="nous" + no env + no auth.json entry → blocks; add a properly-shaped credential-pool entry → both validation and audit clear.

Unit-test coverage: **+59 new tests** (658 → 717 passing). All ABI-compatible — existing tests untouched.

## Migration semantics

- **Additive only** — auto-migration never deletes from the source location. If your key lives in `api_server.token`, it stays there *and* gets copied to `.env`.
- **Idempotent** — second read of the same config is a no-op.
- **Per-profile scoped** — a default-profile key is never copied into a sibling profile that may have intentionally left it unset.
- **Audit-logged** — every migration + auto-fix appends a JSONL entry to `~/.hermes/logs/config-fixes.log` (capped at 1000 entries, value masked to first-4 + last-4 chars).
- **Fail-soft** — log I/O errors and migration errors are silently swallowed; nothing in this PR can block a successful read.

## What's deliberately NOT in this PR

- **Upstream gateway fix** — the right long-term fix is for the gateway to also read `api_server.token` directly. Separate PR against NousResearch/hermes-agent will follow.
- **The rest of issue #367** — missing Nous API-key card + missing Nous OAuth card + credential-pool schema bug (`{key, label}` instead of full shape). Separate desktop PR addresses those; this one only closes the *detection* gap so the existing defensive posture catches them.
- **Other locales** — only EN + pt-PT translated. Existing `t()` fallback handles the rest until the project's batched translation pass picks them up.
- **Cross-profile migration** — out of scope; users running multiple profiles may have intentionally per-profile keys.

Closes: part of #367 (Bug 4 — silent misconfiguration detection).

---

## Update — Windows ↔ WSL hermes-home drift detection (commit 0ddf38e)

Added a 7th piece to the defensive posture: `SIBLING_HERMES_HOME_DRIFT`. Detects config drift between Hermes Desktop's Windows-side `~/.hermes/` (`%LocalAppData%\hermes\`) and any `~/.hermes/` on a WSL distro's filesystem.

Motivation: issue #384 (`Unsloth Studio endpoint API returns "Invalid token payload"`). The reporter's `hermes` CLI on WSL works with "the same settings", but Hermes Desktop on Windows fails. Those are *separate* `~/.hermes/` directories on different filesystems — they don't share state. When one has a value the other lacks (e.g. `model.api_key` set on WSL but not in `%LocalAppData%\hermes\config.yaml`), the desktop's chat silently fails. The audit now catches this and offers a one-click "Copy WSL → Windows" auto-fix where the direction is unambiguous.

| What | Detail |
|---|---|
| New module | `src/main/wsl-detection.ts` — enumerates WSL distros via `\wsl$\<distro>\home\<user>\.hermes\` (filesystem-only, no `wsl.exe -l`, no waking stopped distros). 60s cache. Fail-soft. |
| New issue code | `SIBLING_HERMES_HOME_DRIFT` |
| New check | Compares 21 high-signal fields between current HERMES_HOME and each sibling — every common provider env var, `model.{provider,default,base_url,api_key}`, `api_server.token`, `API_SERVER_KEY` |
| Severities | warning + autoFixable (WSL→Win unambiguous), info (Win→WSL or both-set-but-different) |
| Auto-fix | Inserts into existing parent blocks via newly-exported `upsertBlockChild` (since `setConfigValue` declines to insert missing multi-segment leaves) |
| Cross-platform | No-op on macOS / Linux / non-WSL Windows |
| Tests | +10 in `tests/sibling-hermes-home-drift.test.ts` (mock `findSiblingHermesHomes` to point at temp dirs — exercises real check + auto-fix paths without needing a live WSL distro) |

Total tests on this PR now: **727 passed | 3 skipped** (was 717 before this commit).

---

## Update — Legacy `- hermes` toolset detection (commit 74f07f2)

Added an 8th piece: `LEGACY_TOOLSET_NAME`. The bundled hermes-agent CLI renamed its default toolset alias from `"hermes"` (legacy) to `"hermes-cli"`. Configs from an earlier engine — either a previous bundled version or a prior standalone hermes CLI install — still carry the legacy name in their top-level `toolsets:` block. The current engine's validator rejects it and prints `Warning: Unknown toolsets: hermes` on every agent invocation.

Surfaced by issue #353 and confirmed reproducing on fresh v0.5.1 + engine v0.14.0 installs (Telegram report — Itz Dingo). The agent still runs — the warning is cosmetic — but it clutters every chat session and signals "broken install" to new users.

| What | Detail |
|---|---|
| New issue code | `LEGACY_TOOLSET_NAME` |
| Scope | Top-level `toolsets:` block only — does **not** touch `platform_toolsets.cli` (which has its own validation in `tools.ts`) |
| Tolerates | Quoted forms (`- "hermes"`, `- 'hermes'`), trailing comments, zero-indent and indented list items |
| Does NOT match | `hermes-cli`, `hermes-telegram`, `hermes-discord`, etc. — those are the current canonical names |
| Severity | warning + autoFixable |
| Auto-fix | One-line YAML rewrite preserving indentation, quoting style, and trailing comments |
| Tests | +11 in `tests/config-health.test.ts` covering detection (basic, quoted, platform-suffixed siblings unaffected, platform_toolsets isolation, missing config.yaml) and rewrite (in-place, quoting/comment preservation, no-op when clean, audit log) |

**Live-verified end-to-end via dev electron + CDP**: seeded `- hermes` in real `%LocalAppData%\hermes\config.yaml`, called `rerunConfigHealth` IPC, confirmed the issue surfaces with severity=warning + autoFix=true, called `autofixConfigIssue('LEGACY_TOOLSET_NAME')`, confirmed the file rewrites to `- hermes-cli`, re-audit returns clean, audit log entry recorded. Diagnose UI renders the issue with an Apply fix button.

Closes the silent-warning gap shared by #353 + the Itz Dingo Telegram report.

Total tests on this PR now: **738 passed | 3 skipped** (was 727 before this commit).

Closes #423 — users on v0.5.1 can hit the session-continuation `API_SERVER_KEY` error after a local Ollama-backed first turn because the gateway does not reliably see the same API server key that Desktop resolved. This PR injects/migrates the resolved key and surfaces config-health guidance.

---

## Maintainer Merge Note

Suggested position: after #383 if the maintainer wants the CDP harness separated first; otherwise this can be the first functional PR in the stack.

Known stacked overlap: #383 contains the same CDP harness infrastructure that this PR used for live verification. If #383 lands first, rebase this PR and drop the duplicate CDP harness/script changes from this branch. Later PRs #400 and #434 benefit from the shared URL/env-key resolver introduced here.


pmos69:
Reviewed this pass and found two follow-ups worth fixing before merge.

1. `getApiServerKey(profile)` could copy a default-profile non-canonical `API_SERVER_KEY` source into a named profile's `.env` during migration-on-read. The read fallback itself is fine, but writing the inherited default value into the profile would create profile drift and contradict the PR's per-profile migration contract.
2. Chat readiness and config-health had slightly different local/private URL rules. Readiness allowed LAN/private URLs such as `192.168.x.x`, but config-health only skipped `localhost`/`127.0.0.1`, so local LAN model setups could show a spurious missing-key warning.

I pushed `16a0dc6 fix(config): avoid cross-profile key migration` to this PR branch. It keeps inherited default keys readable without materializing them into named profile `.env` files, moves the local/private URL check into the shared URL helper, and adds regressions for both cases. I also mocked WSL discovery in `config-health.test.ts` so those unit tests don't touch the real host WSL state; the dedicated WSL drift suite remains responsible for that path.

Validation after the fix:
- `npm test -- --run tests/api-server-key-migration.test.ts tests/config-health.test.ts tests/validation.test.ts tests/url-key-map.test.ts` -> 61 passed
- `npm run build` -> passed

fathah:
conflict

PR #376, opened 2026-05-26
fix: duplicate messages from stream-to-DB reconciliation

axing117:
## Problem

Chat messages appear duplicated after streaming completes. The root cause is in `reconcileStreamedWithDb()` — the reconciliation key uses `.trim().slice(0, 200)` to match streamed messages against DB messages. When the stream-accumulated content and the DB-finalised content differ in interior whitespace (e.g. `\n\n` vs ` ` from token concatenation, trailing newlines), the keys diverge and both copies survive the merge.

## Fix

**Two-layer defense:**

1. **Normalised reconciliation key** — added `normalizeWhitespace()` that collapses all runs of whitespace (spaces, tabs, newlines) into a single space before trimming. Applied to both bubble content and reasoning text keys.

2. **Content-level deduplication** — at the end of the merge, when a streamed bubble message has the same role + normalised content as an existing DB bubble in the result, skip it instead of appending a duplicate. Non-bubble messages (tool_call, tool_result) pass through unconditionally since they're matched by callId.

## Testing

- Verified that messages with interior whitespace differences now match correctly
- Verified that genuinely unique streamed messages (error bubbles, etc.) still pass through

### DIFF
--- a/src/renderer/src/screens/Chat/sessionHistory.ts
+++ b/src/renderer/src/screens/Chat/sessionHistory.ts
@@ -119,11 +119,21 @@ export function dbItemsToChatMessages(
  * `null` opts a message out of matching — there's no equivalent on
  * the other side and the reconciliation should treat it as unique.
  */
+/**
+ * Collapse all runs of whitespace (spaces, tabs, newlines) into a single
+ * space and trim.  This prevents the reconciliation key from diverging
+ * when the stream-accumulated string and the DB-finalised string differ
+ * only in interior whitespace (e.g. "\n\n" vs " ").
+ */
+function normalizeWhitespace(s: string): string {
+  return s.replace(/\s+/g, " ").trim();
+}
+
 function reconciliationKey(m: ChatMessage): string | null {
   if ("kind" in m) {
     switch (m.kind) {
       case "reasoning":
-        return `reasoning:${(m.text || "").trim().slice(0, 200)}`;
+        return `reasoning:${normalizeWhitespace(m.text || "").slice(0, 200)}`;
       case "tool_call":
         return `tool_call:${m.callId || m.id}`;
       case "tool_result":
@@ -133,7 +143,7 @@ function reconciliationKey(m: ChatMessage): string | null {
     }
   }
   const bubble = m as ChatBubbleMessage;
-  return `${bubble.role}:${(bubble.content || "").trim().slice(0, 200)}`;
+  return `${bubble.role}:${normalizeWhitespace(bubble.content || "").slice(0, 200)}`;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


fathah:
Hey @axing117 — thanks for digging into this and putting together a clear repro description. The whitespace-normalization fix is the right call and I'd like to get that in. I do have a concern about the second layer though, explained below.

---

## Layer 1 — `normalizeWhitespace` in `reconciliationKey` ✅

This is correct and the change I'd want to merge. `trim()` alone was always fragile here — token concatenation can produce `\n\n` where the DB finalises with a single space, and the key diverges silently. Collapsing all whitespace runs before the 200-char slice fixes that cleanly.

## Layer 2 — content-level tail dedup — I'd like to revisit this

The tail pass (`if (!consumedIds.has(m.id)) result.push(m)`) is specifically the "preserve renderer-only state" path — it's how error bubbles injected by `onChatError` survive the merge. The `seenBubbleKeys` guard added here treats any two bubble messages with the same normalised content as duplicates and drops one of them, which is too aggressive.

The concrete regression: a user who sends the same message twice in a row (e.g. two "ok" turns, or repeated short prompts) can have the second turn silently dropped if any whitespace drift caused a key miss on the first message in the main loop — because by the time we reach the tail pass, the first "ok" is already in `seenBubbleKeys` and the second gets skipped.

The existing test that was changed captures exactly this:

```ts
// before this PR:
expect(merged).toHaveLength(2);
expect(merged[0].id).toBe("a-1");
expect(merged[1].id).toBe("a-2");  // ← second message preserved
```

That test was correct behaviour — two streamed bubbles with identical content where the DB only has one should preserve both (the second is renderer-only state, which might legitimately exist). Changing the expectation to `toHaveLength(1)` converts a valid preservation into a silent drop.

**If Layer 1 lands, does Layer 2 actually remain necessary?** Once the keys normalise correctly, the main loop's FIFO bucket mechanism should match both messages to their DB counterparts, so there's nothing left in the tail pass to deduplicate. The only true "duplicate" scenario Layer 2 catches is when the *streamed array itself* has two messages with the same content and the DB only has one — and in that case the original behavior (preserve both, let the user see any renderer-injected state) seems safer than silently dropping one.

## `mergeDbMetadataIntoStreamed` ✅

Nice addition — preserving DB-only attachments on the streamed copy without swapping the React id is the right approach. One small gap: the `"kind" in streamed` early-return skips `tool_result` messages which also carry `attachments`. Not a regression today, but worth a follow-up.

---

**What I'd suggest**: land Layer 1 + `mergeDbMetadataIntoStreamed` as-is, and either drop Layer 2 entirely or replace it with a much narrower guard (e.g. only skip if `content === ""` — the true degenerate case). Happy to merge a revised version or take Layer 1 as a separate targeted commit if that's easier.

Thanks again for the thorough write-up and test coverage — the direction is right.

PR #382, opened 2026-05-26
fix(providers): Nous Portal UI cards + credential-pool schema + OAuth + discovery (#367)

pmos69:
Closes the renderer + IPC + discovery + OAuth surfaces for #367. All four bugs from the report, plus two follow-ups uncovered while live-testing against a real Nous Portal account.

| Commit | Bug / scope |
|---|---|
| `fix(credential-pool): write entries in the upstream engine schema` | **Bug 3** — `addCredentialPoolEntry` builds `{id, label, auth_type, priority, source, access_token, base_url, request_count}`. Legacy `{key, label}` entries are still read tolerantly. |
| `feat(providers): add Nous Portal API key + OAuth cards` | **Bugs 1 & 2** — the two missing surfaces. EN + pt-PT i18n. |
| `chore(test): CDP harness + live-verify script for #367 fixes` | DOM-level verifier (`scripts/repro-nous-cards-schema-367.js`). |
| `fix(auth): unblock Nous Portal in the OAuth IPC allowlist` | **Found in live testing** — the main-process `OAUTH_LOGIN_PROVIDERS` was a second allowlist that didn't include `nous`. Without this, clicking Sign-in errored with "Unsupported OAuth provider: nous". |
| `feat(models): Nous Portal model discovery + (free) flag in autocomplete` | **Follow-up from live testing** — `discoverProviderModels(\"nous\")` returned `unsupported`. Moved to OAUTH_DISCOVERY_PROVIDERS (244-model curated list now flows through). Plus: query the portal's live /v1/models with the OAuth token to flag free models — surfaced as `<option label=\"(free)\">` in the autocomplete so free-tier users don't pick a paid model by accident. |

## Bug 4 — handled by sibling PR

The fourth bug (silent misconfiguration when `model.provider: \"nous\"` is set without any creds) is the **detection** half — addressed in #369. The two PRs are independently mergeable in either order.

## Verification

- `npm run typecheck` clean
- `npm test` — **670 passed | 3 skipped** (was 658 baseline; +12 new tests in `tests/credential-pool-schema.test.ts` and `tests/model-discovery.test.ts`)
- **Live verified end-to-end with @pmos69's free Nous Portal subscription** (signed in via Google OAuth, real Nous inference endpoint):
  - ✅ `Nous Portal API Key` card renders in LLM Providers section
  - ✅ `Nous Portal (OAuth)` Sign-in card renders + button is clickable
  - ✅ OAuth device-code flow opens browser → user signs in with Google → tokens land in `auth.json` (`providers.nous` + `credential_pool.nous[0]` in canonical shape with `access_token`, `refresh_token`, `agent_key`, `auth_type: \"oauth\"`, `source: \"device_code\"`, `inference_base_url`, etc.)
  - ✅ `discoverProviderModels(\"nous\")` returns 244 models with `freeModels: [\"deepseek/deepseek-v4-flash:free\", \"openrouter/owl-alpha\"]`
  - ✅ Chat round-trip: switch `model.provider: \"nous\"` + `default: \"deepseek/deepseek-v4-flash:free\"` → send a prompt → real reply from `inference-api.nousresearch.com`: **\"NOUS WORKS\"**

Live-verify scripts persisted in `scripts/`:
  - `drive-nous-oauth.js` — drives the Sign-in card, captures verification URL + code, watches for completion
  - `drive-nous-chat.js` — switches the active model and sends a one-line probe
  - `verify-nous-discovery.js` — pins the discovery + free-flag contract against a real account
  - `repro-nous-cards-schema-367.js` — DOM-only check for card rendering + schema (no account needed)

## Migration safety

- Existing `{key, label}` entries in users' `auth.json` are **read** tolerantly but not rewritten. They stay as-is until the user re-adds them.
- Adding a Nous key through the UI now produces a working credential the engine can resolve — that's the new behaviour. Pre-fix, the same user action produced a non-functional entry.

Closes part of #367 (Bugs 1, 2, 3, plus the OAuth IPC + discovery infrastructure to make them work end-to-end). Sibling PR #369 handles Bug 4 (detection).

### DIFF
--- a/scripts/drive-nous-chat.js
+++ b/scripts/drive-nous-chat.js
@@ -1,0 +1,133 @@
+/**
+ * Drive a Nous Portal end-to-end chat round-trip via the dev electron.
+ *
+ *   1. Backup config.yaml + remember active model.
+ *   2. Switch model.provider → "nous", model.default → "Hermes-4-405B"
+ *      (or the smallest available Nous model name; the engine resolves
+ *      it). Clear base_url so the engine uses portal's inference URL
+ *      from auth.json.
+ *   3. Send a one-liner chat.
+ *   4. Wait for chat-done; capture the agent's reply.
+ *   5. Restore the original model selection.
+ */
+const { attach } = require("./e2e-attach");
+const fs = require("fs");
+const path = require("path");
+const os = require("os");
+
+const CONFIG = path.join(
+  os.homedir(),
+  "AppData",
+  "Local",
+  "hermes",
+  "config.yaml",
+);
+const CONFIG_BAK = CONFIG + ".nous-chat-bk";
+
+const NOUS_MODEL = process.env.NOUS_MODEL || "Hermes-4-405B";
+const PROMPT =
+  process.env.PROMPT ||
+  "Reply with exactly two words: 'NOUS WORKS'. Nothing else.";
+
+(async () => {
+  fs.copyFileSync(CONFIG, CONFIG_BAK);
+
+  try {
+    // Mutate config: provider = nous, default = NOUS_MODEL, blank base_url
+    const cfg = fs.readFileSync(CONFIG, "utf-8");
+    const cfgNew = cfg
+      .replace(/^(model:\s*\n(?:.*\n)*?\s*provider:\s+).*$/m, `$1"nous"`)
+      .replace(
+        /^(model:\s*\n(?:.*\n)*?\s*default:\s+).*$/m,
+        `$1"${NOUS_MODEL}"`,
+      )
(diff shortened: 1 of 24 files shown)


fathah:
**`cryptoRandomId` uses `Math.random()`**

Pool entry IDs should be unpredictable. `crypto` is already a Node built-in used elsewhere in the codebase — no overhead concern:

```typescript
import { randomBytes } from "crypto";

function cryptoRandomId(): string {
  return randomBytes(4).toString("hex"); // 8 hex chars, same output length
}
```

pmos69:
Good catch — opened #394 with the suggested fix. Same output shape (8 lowercase hex chars), now actually crypto-random.

Issue #363, opened 2026-05-25
Remote-mode feature parity for all-Windows deployments (Sessions / Kanban / Models / etc.)


pmos69:
## Summary

Hermes Desktop currently has two connection modes for non-local hermes-agent backends:

- **Remote** (HTTP + bearer token) — chat-only. Sessions / Kanban / Models / Profiles / Memory / Tools / Persona / Settings all show *"Connected to remote Hermes — this data is not accessible through the API yet"*.
- **SSH Tunnel** — full functionality, but the entire `src/main/ssh-remote.ts` layer assumes a Unix remote shell (`$HOME`, `python3`, `kill -0`, `systemctl`, `[ -f ... ]` test brackets, `hermes` with no `.exe`). It does not work against a Windows-native `hermes.exe` host. PR #305 widened the venv probe paths but those are still Unix-only.

For users whose **hermes-agent runs on a Windows host** (and there are several recent reports — #334, #348 partially, the original #284), neither mode delivers the full experience.

## Canonical use case (#334 thread, @ssatan34)

> hermes-agent installed on a 24/7-online Windows mini-PC at home; another Windows PC at home accesses it on the LAN; from work, accessed via Lucky reverse proxy (HTTP). Wants Sessions / Kanban / Models / Profiles to all work.

All-Windows, no SSH server, no Linux/macOS box involved anywhere. The Lucky reverse proxy already terminates HTTPS in front of the gateway — extending Remote mode with a richer HTTP API would be a natural fit (no new transport, no SSH dependencies, no Windows shell variance).

### Path A — Windows-native `ssh-remote.ts`

Add Windows shell variants throughout `src/main/ssh-remote.ts`:

- Detect the remote OS (probe `uname` vs `ver` / `$PSVersionTable`).
- Branch every command builder on OS — `buildGatewayStatusCommand`, `buildGatewayStartCommand`, `remoteEnvPath`, `buildRemoteHermesCmd`, etc.
- Probe `hermes.exe` paths: `%LOCALAPPDATA%\hermes\hermes-agent\hermes.exe`, `%ProgramFiles%\hermes\`, etc.
- Use PowerShell or `cmd` syntax depending on the remote OpenSSH default shell.

**Prerequisite (host-side):** OpenSSH Server installed, started, autostart enabled, firewall port 22 open, default shell registry value set (PowerShell vs cmd, security-relevant), `authorized_keys` provisioned (with the Windows admin-authorized-keys quirk for elevated users).

[...]

PR #355, opened 2026-05-24
Remote-mode parity for the Codex /api backend

to3tho:
### Why

The desktop currently drops most tabs to a generic "is not
available in remote mode" notice when connected to a remote
Hermes backend. Codex's PR
[NousResearch/hermes-agent#23742](https://github.com/NousResearch/hermes-agent/pull/23742)
and the four follow-up PRs in that branch line up the backend
side; this PR is the matching App-side migration so the desktop
becomes a first-class client of that surface.

After this PR all **14 sidebar tabs** avoid the generic
remote-mode unavailable notice. **11 are newly wired** to the
backend `/api/*` telemetry / read surfaces — Memory, Tools,
Schedules, Kanban, Gateway, Sessions, Profiles, Providers,
Skills, Persona, Settings; **Office / Chat / Models remain on
their existing working paths** (no change in this PR). Five of
the newly-wired tabs ship with full write surfaces: Memory
(entries + USER.md), Persona / Soul, Tools (per-profile
toggle), Cron, Kanban CRUD.

This is large for a single PR. Plan to review it as five
**logical phases**, each commit-prefixed for navigation, rather
than a flat commit list. If splitting into separate PRs would
help, happy to do that — see "Split if requested" below.

### Phase map

| Phase | What | Commits |
|---|---|---|
| **A — Foundation + initial 5 tabs** | `Envelope<T>` shared types, `TelemetryClient` (mode-aware base URL), `CapabilitiesProvider` + `useCapability` hook, `useTelemetryQuery` hook, `TelemetryCard` wrapper, mount in Layout, wire 5 initial tabs (Memory / Tools / Schedules / Kanban / Gateway) to the new state-system, shared DTOs + IPC + preload for them | 17 |
| **C — Read Phase 2** | Sessions, Profiles, Providers, Skills, Persona TelemetryViews; matching IPC handlers, preload methods, shared DTOs; ipc-surface test coverage | 6 |
| **D — Events + Usage + Settings fallback** | Recent Activity card + Usage card on the Gateway tab; populate Engine / Released / Python / OpenAI SDK on the Settings tab in remote mode; ipc-surface coverage | 6 |
| **E1–E2 — Cron + Kanban CRUD UI** | Mutation client, `MutationResult<T>` shape, Cron CRUD UI on Schedules tab, Kanban CRUD UI on Kanban tab, `hermesAPI.cron.*` + `hermesAPI.kanban.*` preload surfaces, ipc-surface coverage | 9 |
[...]

PR #396, opened 2026-05-26
Phase A: telemetry foundation + initial 5 read-only tabs (depends-on #23742, #31125)

to3tho:
## Phase A: foundation + three wired read-only tabs for the Codex /api/* Remote Management API surface

This PR ships Phase A of a migration that wires hermes-desktop's
remote-mode tabs to the Codex `/api/*` Remote Management API
surface (NousResearch/hermes-agent#23742, plus the
`/api/gateway/status` enrichment from #31125).

This PR was revised in response to @pmos69's line-by-line
review (see the comment thread below). The original Phase A
submission predates the `/api/*` adapter migration on the
multi-phase reference branch. The hardening commits cherry-pick
the necessary parts of that migration onto Phase A's surface.

---

### Scope after the hardening pass

Wired (read-only) in remote mode:

| Tab | Wire path |
|---|---|
| **Memory** | `GET /api/memory` |
| **Tools** | `GET /api/tools/toolsets` |
| **Gateway** | `GET /api/gateway/status` + `GET /v1/capabilities` |

Out of scope in this PR — surface a `RemoteNotice` in remote mode rather than a broken / empty local-screen fallback:

- **Schedules** — no upstream contract in the cited dependency chain. `/api/jobs` exists in #23742 but doesn't carry a `SchedulesTelemetry` contract (no `kind` field, fragile mapping). Belongs in a later phase with an explicit read-summary endpoint.
- **Kanban** — `/api/kanban/*` lives in follow-up #31641, not in the cited #23742 + #31125 dependency chain. Belongs in the Phase E1-E2 split.
- **Sessions / Profiles / Providers / Skills / Persona / Settings** — unchanged from existing behaviour, not part of Phase A scope.

---

### Architecture: adapter pattern

Each `fetch*` function in `src/main/telemetry/subsystems.ts`
calls the real `/api/*` endpoint and adapts the raw shape to
the renderer's `*Telemetry` DTO via a small `adapt()` helper.
React views stay unchanged when upstream evolves — only the
adapter needs to follow.

`fetchGatewayStatus` additionally union-merges
`status.subsystem_capabilities` (the canonical hook from
#31125) with `/v1/capabilities.features.remote_*` flags (older
fallback signal) into the renderer's `capabilities` list.

### Foundation layer

- **`Envelope<T>`** discriminated union for the telemetry wire
  contract (`{available:true,data:T}` vs
  `{available:false,reason,detail?}`).
- **`TelemetryClient`** with mode-aware base URL +
  error-to-envelope mapping (404 → `not-implemented`, 5xx →
  `upstream-error`, etc.). Wraps a raw-JSON fallback when the
  backend doesn't emit the envelope contract.
- **`CapabilitiesProvider`** runs the gateway-status probe
  once on connect; re-probes when a `probeKey` derived from
  the full connection signature changes (mode + URL +
  `hasApiKey` + ssh fields).
- **`useTelemetryQuery`** 4-state machine
  (`loading | empty | error | data`), reads the capability
  gate before fetching.
- **`TelemetryCard`** renders all four states with consistent
  empty / error / loading affordances.

### Defence-in-depth: payload validation

`telemetryGet` accepts an optional `validateShape` hook that
runs on BOTH wrapper paths:

- **Envelope path**: when the backend sends
  `{available:true, data:...}` the validator runs on `data`
  before forwarding.
- **Raw-JSON fallback**: when the backend forgets the envelope
  the validator runs on the parsed body before it's wrapped
  as `available:true`.

Required keys only — unknown fields are tolerated by design
(forward-compat with backend additions). On reject the envelope
is `available:false, reason:upstream-error` with no `detail`
string; the verdict goes to `console.warn` in the main process
for diagnostics, not to the user-visible `RemoteNotice` detail
line.

For the Phase A endpoints:
- `/api/gateway/status` requires `ok` + `running` booleans
- `/api/tools/toolsets` requires `toolsets` array; per-item
  shape requires a non-empty string `key`
- `/api/memory` requires `memory` + `user` objects

The per-item toolset check catches payloads like
`{toolsets: [{}]}` that would otherwise reach React as
`<li key={undefined}>`.

---

### Tests

690 vitest tests, 3 skipped, 0 failures across 61 files
(global `npm run lint`: 0 errors, 2 pre-existing react-hooks
warnings unrelated to this PR).

#### New: `tests/telemetry-contract.test.ts`

@pmos69's explicit request: a regression guard against upstream
wire-shape drift. For each of the 3 wired endpoints the file
mocks an HTTP server at the documented path and returns the
documented live shape (verified against deploy/codex-stack on
2026-05-25), then runs the real adapter against it. Three assertion classes per endpoint:

- happy-path live shape → adapter returns the expected DTO
- path-tracker asserts the adapter hits `/api/<endpoint>` —
  not a legacy `/v1/telemetry/*` regression
- wrong-shape payload → adapter returns `upstream-error`,
  doesn't silently wrap garbage

Fixtures carry the redundant snake_case + camelCase fields,
the empty arrays, and the `<redacted>` `user.content` from
the sanitiser (#31568) — exactly as the live backend ships
them.

---

### Hardening commits (on top of the original Phase A 16)

| Commit | Subject |
|---|---|
| `fc12b22` | `fix(telemetry): migrate Phase A adapters to /api` |
| `f9a188b` | `fix(telemetry): validate envelope.data + raw payloads` |
| `69e7282` | `chore(lint): drop unused profile prop from ToolsTelemetryView` |
| `ce5893b` | `fix(telemetry): re-probe capabilities on backend swap` |
| `2b3afdf` | `fix(layout): RemoteNotice for Schedules + Kanban in remote` |
| `f8c57de` | `fix(telemetry): uptime "—" when backend doesn't expose it` |
| `7194aef` | `chore(format): clean telemetry prettier drift` |
| `64094db` | `test(telemetry): contract tests for /api/* shapes` |

Each commit subject describes its scope; the body links back
to which review finding it addresses.

---

### Depends on

Two upstream backend PRs on `NousResearch/hermes-agent`:

- **[#23742](https://github.com/NousResearch/hermes-agent/pull/23742)**
  — Codex's Remote Management API foundation. **Hard
  dependency** — the three read endpoints
  (`/api/memory`, `/api/tools/toolsets`, `/api/gateway/status`)
  ship in this PR.
- **[#31125](https://github.com/NousResearch/hermes-agent/pull/31125)**
  — `/api/gateway/status` enrichment with
  `subsystem_capabilities` + version / python_version /
  openai_sdk_version / released fields.

Both are Draft on upstream. This PR stays Draft until they land
or the maintainer indicates a preferred merge strategy.

### Acknowledgements

@pmos69's two rounds of detailed review (initial paths +
DTO/wrapper drift) shaped this version end-to-end. The
contract tests in particular were his explicit request and
catch exactly the regression class his review surfaced.


pmos69:
@to3tho — read this end-to-end as offered in #355. **Foundation *architecture* is solid; the *integration* with the cited upstream PRs is broken at three layers — paths, DTO shapes, and a runtime crash path.** A path-rename pass alone won't fix it. Substance first, design notes after.

## 🔴 Blocking — integration mismatches against #23742 + #31125

Verified by reading #23742's and #31125's diffs/bodies directly. The PR's wire contract is incompatible with the dependencies it cites at three layers; fixing any one of them leaves the others.

### 1. Endpoint paths

The code calls `/v1/telemetry/*`; upstream #23742 exposes `/api/*`. None of the five read paths can resolve against the dependency this PR claims.

| Subsystem | This PR (`subsystems.ts:24-43`) | What #23742 actually adds |
|---|---|---|
| Gateway status | `/v1/telemetry/gateway-status` | `/api/gateway/status` |
| Tools | `/v1/telemetry/tools` | `/api/tools/toolsets` |
| Memory | `/v1/telemetry/memory` | `/api/memory` |
| Schedules | `/v1/telemetry/schedules` | **not in #23742** (see #4) |
| Kanban | `/v1/telemetry/kanban` | **not in #23742** (see #4) |

Your PR body also says the paths are `/api/{...}` and `/api/gateway/status`, which matches upstream — so the description and code disagree with each other too.

### 2. DTO shapes don't match — path-renaming alone leaves the integration broken

After flipping the paths to `/api/*`, the bodies upstream returns are completely different shapes from what this PR models. Verified from #23742's diff:

| Endpoint | What #23742 actually returns | What this PR models in `telemetry-types.ts` |
|---|---|---|
| `/api/gateway/status` | `{ok, running, pid, gateway_state, platforms, active_agents, exit_reason, updated_at, version, python_version, openai_sdk_version, released, subsystem_capabilities}` (with #31125 enrichment) | `{service, version, uptimeSeconds, capabilities, upstreamProviders[]}` |
| `/api/memory` | `{memory: {entries[], charCount, charLimit, ...}, user: {charCount, charLimit, ...}, stats: {totalSessions, totalMessages}}` | `{provider, configured, itemCount?, sizeBytes?, lastUpdatedAt?}` |
| `/api/tools/toolsets` | `{toolsets: [{key, name, label, description, enabled, available, configured, tools}], platform}` | `{toolsets: [{key, label, description, enabled, source, mcpServer?}]}` |

For memory in particular: **zero field overlap** between the upstream payload and the renderer DTO. The shared `MemoryTelemetry` type isn't a subset, an alias, or a restructured view of the upstream shape — it's a completely different summary the upstream endpoint doesn't produce.

This isn't a rename. It needs explicit adapters from the upstream `/api/*` payloads to the renderer DTOs (and matching tests against the real upstream shape), or the shared types + views have to be rewritten to consume the upstream shapes directly.

### 3. The defensive envelope-wrapper makes the DTO mismatch silent — and a crash path

`client.ts:120-126` does:

```ts
if (json && typeof json === "object" && "available" in json) {
  resolve(json);
  return;
}
// Treat as raw data — wrap it ourselves.
resolve({ available: true, data: json as unknown as T });
```

I called this defensive wrapping "a strength" in my first read. With the DTO mismatch in #2 above, it inverts: a real `/api/memory` response (no `available` key) gets wrapped as `{available: true, data: theRawWrongShape}` and flows into the renderer as **successful telemetry data**. The renderer then accesses fields that don't exist:

- **`GatewayTelemetryView.tsx` will hard-crash on render.** Reads `data.capabilities.length === 0`, `data.capabilities.join(", ")`, `data.upstreamProviders.length > 0`, `data.upstreamProviders.map(...)`. When upstream sends `subsystem_capabilities` (no `capabilities`) and no `upstreamProviders`, both `.length` and `.join` / `.map` are `TypeError: Cannot read properties of undefined`. Same for `useCapability.ts:25` (`ctx.data.capabilities.includes(key)`).
- **`MemoryTelemetryView.tsx` will render blanks silently.** The guards (`{data.itemCount !== undefined && (…)}`) are correctly defensive against missing fields — so the user sees a Memory tab that loads "successfully" but displays nothing.

So depending on which tab the user opens, the integration either crashes the whole Layout (no error boundary covers these views as far as I can see) or returns an unbroken-but-empty UI. Neither is what the user or the maintainer would want; both look like Phase A "works" when it doesn't.

The intent of the wrapper — tolerate backends that don't speak the envelope yet — is fine. The problem is that the same code path can't distinguish "old backend forgot the envelope but the shape is right" from "different backend's raw payload with the wrong shape." Suggested: tighten the fallback to **validate the wrapped data matches the expected DTO** before resolving `available: true` (zod-style or hand-rolled), and surface a parse-error envelope otherwise. This also gives you a hook for the missing field-shape test (see below).

### 4. Schedules and kanban have no matching upstream contract

#23742 does expose cron/job management routes under `/api/jobs` (`GET /api/jobs`, `POST /api/jobs`) — but it does not expose this PR's `/v1/telemetry/schedules` or an `/api/schedules` read-summary endpoint matching `SchedulesTelemetry`. Kanban likewise has no matching upstream route in #23742. So even after fixing paths/adapters for gateway, memory, and tools, the **schedules** and **kanban** telemetry tabs still need either a new upstream contract or a different source.

Phase A's scope claim — *"five tabs share the same Envelope wire shape and could be wired in one coherent commit batch"* — turns out to only hold for three of them (tools, memory, gateway-status). Schedules and kanban need either a separate upstream PR proposal (a read-summary endpoint on top of `/api/jobs` for schedules, a brand-new endpoint for kanban) or a different read source (e.g. the existing local IPC handlers in `src/main/{cronjobs,kanban}.ts` proxied over the SSH tunnel).

### How the tests didn't catch this

`tests/telemetry-client.test.ts` spins up its own HTTP server with `/v1/telemetry/*` routes (lines 8, 21, 36, …) and asserts the client talks to *those*, returning bodies the PR's own DTOs expect. So the suite validates the client against an internal idealised wire format and DTO shape, never against the actual paths or payloads #23742 publishes. Solid internal test, but it can't see the protocol drift at any of the three layers.

Recommend: a single regression test that mocks the upstream **as documented in #23742's body** — exact response shape, exact path — and runs the client against it. That would have surfaced #1, #2, and the silent-failure path from #3 on day one.

## 🟢 What's working really well

The foundation layer's *architecture* is clean and well-thought-out. The issues above don't touch architecture — they're contract-against-reality rather than rework.

- **`TelemetryEnvelope<T>` discriminated union** (`telemetry-types.ts:32-38`). The `{available:true,data}` vs `{available:false,reason,detail?}` split is exactly the right shape — exhaustive checks at every consumer, no in-band sentinels, no `data | null` ambiguity. The `TelemetryUnavailableReason` enum (`not-configured` / `not-implemented` / `remote-mode-blocked` / `upstream-error`) is well-chosen: each carries genuinely different user-facing meaning.

- **Drift fallback for 404 → `not-implemented`** in `client.ts:99-101`. Catches the "old backend predates the endpoint" case cleanly. (As noted in #3 above, the *other* drift fallback — raw-JSON-wrapping — needs tightening, but the 404 mapping is solid.)

- **`useTelemetryQuery`'s four-state machine** (`useTelemetryQuery.ts:30-38`). `loading | empty(reason) | error(message) | data(T)` covers every UI state without conflating them — the empty-vs-error distinction always trips people up, and it's clean here.

- **`probeKey` re-trigger in `CapabilitiesProvider`** (`CapabilitiesProvider.tsx:83`). Cleanest possible invalidation hook for the mode-switch case.

- **`cancelled` flag** in both `useTelemetryQuery` and `CapabilitiesProvider` prevents stale-setState-after-unmount, in the standard React-effect pattern.

- **Read-only scope discipline.** No write surface, no `mira-uitest` allowlist, no CodeMirror dependency dragged in for Persona/Soul. Exactly the carve-out shape I suggested in #355 — you held the line on it, which is what makes this slice actually reviewable.

## 🟡 Design observations (not blockers, but worth raising)

- **`useCapability` collapses `unavailable` reasons into `"absent"`** (line 27). The doc justifies this as conservative — and it is — but the user sees identical empty-states for "backend doesn't have this feature" vs "backend is unreachable." Different actionable advice. A `present | not-implemented | network-error` trichotomy at the UI layer would let the empty-state copy diverge ("Update your Hermes" vs "Check your connection"). Doesn't have to ship in Phase A, but worth a TODO.

- **No `AbortSignal` plumbed through the fetcher**. `useTelemetryQuery` sets `cancelled=true` to drop the response, but the underlying HTTP request keeps running to completion. For SSH-tunneled deployments over high-latency links this is wasted bandwidth, and a tab-switch storm can pile up duplicate in-flight requests. `client.ts` already constructs the `http.request` object, so plumbing an `AbortController` through `telemetryGet(path, {signal})` and into `useTelemetryQuery`'s effect cleanup is mechanical. Worth a follow-up.

- **`TelemetryUnavailableReason` collapses transport-failed and got-an-error-response into the same `"upstream-error"`** bucket. Same UX-divergence point as the `useCapability` note — but trickier because the client side already loses that distinction at the catch boundary. If you ever want to surface "backend offline" vs "backend said 500" differently, the union needs splitting.

- **`probeKey?: unknown`** (`CapabilitiesProvider.tsx:44`). Works but `<T>` generic with `probeKey?: T` would preserve type info at the call site. Tiny.

- **Cap-list-vs-endpoint-404 asymmetry.** If `/api/gateway/status` itself 404s, the whole context lands in `unavailable.not-implemented` and `useCapability` returns `"absent"` for everything — right answer. But if the cap list says `present` and a per-subsystem endpoint 404s, the user-facing message is "this feature requires a newer backend" even though the backend is the right version, the endpoint just briefly disagreed. Hints that the endpoint behavior, not the cap list, should be authoritative. Worth thinking about whether the cap list should be advisory (skip the request) or authoritative (refuse the request).

## Path forward suggestion

The foundation architecture survives in place. What needs to change is not mechanical:

1. **Decide whether the renderer DTOs consume the upstream raw shape directly, or whether you add adapters in `subsystems.ts` that translate upstream `/api/*` payloads into the existing renderer DTOs.** Either is defensible — adapters give you a stable internal contract that survives upstream wire changes; direct consumption gives you fewer moving parts but more renderer changes when upstream evolves. Either way, **#23742's actual response shape is the source of truth** for what fills these DTOs — not the PR's invented summary shapes.
2. **Rebase paths to `/api/*` and the capability field to `subsystem_capabilities`**, *after* deciding #1.
3. **Drop schedules + kanban from this PR's tab list**, or carve a new upstream PR proposing `/api/{schedules,kanban}` endpoints (you have working IPC handlers locally for both — the schema is already there, it's an exposure question on the upstream side).
4. **Tighten the envelope-wrapper's raw-JSON fallback** to validate the shape before resolving `available: true`, so a misaligned upstream payload surfaces as a parse error instead of a silent renderer crash.
5. **Add one integration test** that mocks the upstream wire shape exactly as documented in #23742's body — including the `subsystem_capabilities` field name — so this kind of drift can't recur silently.

With those changes Phase A goes from "depends on two open upstream PRs that don't match the code" to "depends on the same two upstream PRs and actually integrates with them." The 1991-line foundation's *architecture* stays in place; the *contract layer* gets a real revision.

Substantively: the architecture is right, the test discipline at the unit level is right, the read-only scope is right. The wire layer + DTO layer + the silent-failure path through the envelope wrapper all need work. Happy to keep reviewing follow-ups.

to3tho:
@pmos69 — thanks for the second pass. It caught the important mismatches in the Phase A carve-out, and I pushed 8 hardening commits on top of the original 16 that address them directly. PR body updated with the revised scope + commit chain; the high points:

**Wire contract realignment**

The original Phase A predated the `/api/*` adapter migration on the multi-phase reference branch — the diff you were reading was, in effect, a Phase A carve-out from before that migration. The first hardening commit (`fc12b22`) brings over the necessary adapter pieces: paths flip from `/v1/telemetry/*` to `/api/*`, an adapter layer in `subsystems.ts` maps the live Codex shapes into the renderer's DTOs, `GatewayStatusTelemetry` picks up `released` / `pythonVersion` / `openaiSdkVersion`, capabilities union-merge `subsystem_capabilities` with `/v1/capabilities.features.remote_*`.

**Wrapper hardening (your point 3 in the second review)**

The envelope-wrapper used to wrap raw JSON as `available:true, data:raw` permissively. `f9a188b` adds an opt-in `validateShape` hook that runs on BOTH wrapper paths — `env.data` on the envelope path, and the parsed body on the raw-JSON fallback. Required-key-only by design so backend additions don't have to revisit this layer. On reject the envelope is `available:false, reason:upstream-error` with NO `detail` string — the verdict goes to `console.warn` in main, not to `RemoteNotice`'s user-visible detail line. Toolsets validator gained a per-item shape check (catches `[{}]` payloads that would otherwise reach React as `<li key={undefined}>`).

**Scope reduction**

Schedules + Kanban are out of Phase A. As you flagged: `/api/jobs` exists in #23742 but doesn't carry a `SchedulesTelemetry` contract (no `kind` field, fragile mapping). `/api/kanban/*` lives in the follow-up #31641, not in the cited dependency chain. Both tabs render `RemoteNotice` in remote mode (`2b3afdf`) instead of falling through to the local screen — which would have reached for a local SQLite/cron daemon that doesn't exist on the remote backend.

**Contract tests**

`tests/telemetry-contract.test.ts` (`64094db`) — exactly the regression guard you asked for. Per endpoint:

- mocks an HTTP server at the documented path,
- returns the documented live shape (verified against deploy/codex-stack on 2026-05-25 — `/api/gateway/status`, `/api/tools/toolsets`, `/api/memory` plus a `/v1/capabilities` fixture for the merge step),
- runs the real adapter from `subsystems.ts` against it,
- asserts (a) right DTO out, (b) right path was called (a `pathsHit` tracker catches a legacy-path regression), (c) wrong-shape payload → `upstream-error`, not silent wrap.

Fixtures keep the redundant snake_case + camelCase fields and the `<redacted>` `user.content` from the sanitiser (#31568) — exactly as the live backend ships them.

**Smaller fixes from your design observations**

- `probeKey` (`ce5893b`) — was a `remoteMode` boolean; now a `JSON.stringify(connectionConfig)` signature so a backend swap in Settings re-probes even when the mode stays `remote`. Documented limitation: `hasApiKey` is a boolean (renderer never sees the raw key), so a value-only key swap on the same URL isn't captured — could be closed with a config-changed event from Settings in a follow-up.
- Uptime UI honesty (`f8c57de`) — adapter no longer fakes `uptimeSeconds: 0` when the backend doesn't expose uptime. The DTO field becomes optional and `formatUptime(undefined)` renders `"—"`. A real `0` (gateway just restarted) still renders as `"0s"`.
- `ToolsTelemetryView` unused prop drop (`69e7282`) — the `_props` parameter that ESLint flagged is gone; Layout caller drops the prop too. When cross-profile scoping arrives via #31607 in a later phase, the prop comes back with real adapter use.
- Plus a prettier-only sweep (`7194aef`) so the global `npm run lint` task is clean (0 errors, 2 pre-existing react-hooks warnings unrelated to this PR).

**Tests**

690 vitest tests, 3 skipped, 0 failures across 61 files. `npm run typecheck` clean.

---

Two review points are intentionally left out of this Phase A patch, not because I disagree with them, but because they change the foundation contract rather than the Phase A adapter wiring: the `useCapability` absent/network-error distinction, and the transport-vs-domain-error split in `TelemetryUnavailableReason`. I agree both are valid follow-ups; I kept this PR focused on correcting the documented `/api/*` wiring, validation, and remote-mode fallbacks so it can land without expanding the base envelope contract in the same pass.

If you have appetite for another pass, the Windows live-test you offered is the one I can't replicate locally. Once #23742 and #31125 are mergeable upstream — or if a maintainer wants to test against a backend running the deploy stack we've been deploying internally — your scenario in #363 (mini-PC + Lucky reverse proxy) is the canonical integration target.

Thanks again for the careful read. The contract-test discipline in particular is a pattern I want to keep across the rest of the phases.

pmos69:
@to3tho — thanks for the hardening pass. I re-read the adjusted diff against the points from the previous review, and the substantive blockers I raised look addressed now.

The important pieces are in much better shape:

- `/v1/telemetry/*` is gone from the actual adapter paths for the wired tabs; the Phase A reads now target `/api/gateway/status`, `/api/tools/toolsets`, and `/api/memory`.
- The adapter layer in `subsystems.ts` is the right compromise here: the renderer keeps a stable DTO while the main process owns the upstream `/api/*` shape.
- The raw-JSON fallback is no longer permissive for these subsystem calls. `validateShape` on both envelope and raw paths closes the silent-success/crash path I was worried about.
- Dropping schedules + kanban from the wired Phase A scope is the right call. Rendering the remote-mode notice instead of falling through to local SQLite/cron paths avoids a whole class of false “remote” behavior.
- `tests/telemetry-contract.test.ts` is exactly the kind of adapter-level regression guard I was asking for: documented path, documented live shape, adapter output, and wrong-shape rejection.

I ran the telemetry-focused suite and typecheck locally:

```bash
npm test -- --run tests/telemetry-client.test.ts tests/telemetry-contract.test.ts tests/telemetry-ipc-surface.test.ts src/renderer/src/hooks/useTelemetryQuery.test.tsx
npm run typecheck
```

Both passed.

I also ran the full Vitest suite once. It had a timeout in `tests/locale-persistence.test.ts`, but that same test passed when rerun in isolation, so I’m treating that as an existing/environmental timing flake rather than a telemetry regression.

Only small thing I noticed on this pass:

`src/main/telemetry/subsystems.ts` picks `lastUpdatedAt` with the first present timestamp:

```ts
mem.last_modified ?? mem.lastModified ?? usr.last_modified ?? usr.lastModified
```

If both `MEMORY.md` and `USER.md` exist and `USER.md` changed later, the Memory tab can show the older memory timestamp. The contract fixture already has this shape: memory is older than user, but the assertion only checks that the output is ISO-shaped. I’d take the max finite epoch across both memory/user timestamps and assert the fixture maps to the newer one.

That’s not a blocker from me. The original integration mismatch is fixed in substance, and the new contract tests should keep it fixed.

to3tho:
Thanks, good catch. I fixed that in 769aa01.

`fetchMemory` now takes the max finite epoch across the memory/user timestamps instead of the first-present value, with Number() coercion for defensive handling. The contract fixture already had memory older than user, so I tightened the assertion to require the newer user timestamp exactly rather than only checking ISO shape.

Targeted telemetry tests and typecheck are still passing. I also removed the accidental CIFS executable-bit artifact before pushing, so the remote tree keeps `src/main/telemetry/subsystems.ts` at 100644.

pmos69 (commented):
Thanks for the follow-up. I did a smaller re-check of the hardening commit and the workflow risk surface.

The `fetchMemory` timestamp fix in `769aa01` looks good to me: taking the max finite epoch across `memory` and `user` is the right behavior, and tightening the contract test to assert the newer USER.md timestamp makes this a real regression guard instead of a shape-only check.

I also checked the workflow-approval angle. This PR does not touch `.github/workflows`, `package.json`, lockfiles, or `scripts`, and the accidental executable-bit artifact appears to be gone (`src/main/telemetry/subsystems.ts` is back to `100644`). So from my side it is OK to run the normal PR workflow.

Only non-blocking note I still see: a few comments/types still refer to the old `/v1/telemetry/*` endpoint names even though the actual Phase A adapters now target `/api/*`. That is documentation drift, not a functional blocker, but it would be nice to clean up before merge.

pmos69:
image
image
image

to3tho:
Fixed in 8419a08 — corrected all stale /v1/telemetry refs, Schedules/Kanban marked as not wired in Phase A. Thanks for catching the drift.

PR #400, opened 2026-05-26
fix(compat): write host-derived <VENDOR>_API_KEY for custom providers

### LINKED ISSUE #360: Custom OpenAI-compatible provider: API key env-var name mismatch between Desktop UI and Agent runtime + invalid YAML in default config
## Summary

When using a custom OpenAI-compatible provider whose base URL is **not** one of the hard-coded vendors (e.g. `https://www.arccodex.com/api/codex/v1`), the API key entered in the Desktop UI is never picked up by the Agent runtime, resulting in a permanent `401 Invalid or missing API key` from the upstream provider.

There are actually **two independent bugs** that together make custom OpenAI-compatible providers unusable out of the box.

## Environment

- App: Hermes Agent Desktop (latest, fathah/hermes-desktop)
- Provider in UI: **OpenAI Compatible / Local**
- Base URL: `https://www.arccodex.com/api/codex/v1`
- Model: `gpt-5.5`
- OS: Windows

---

### What the UI does

In `src/renderer/src/screens/Models/Models.tsx`, when the user saves an API key for a custom OpenAI-compatible model, the key is written to `.env` under a UI-derived variable name. For unrecognized hosts the UI falls back to a generic name:

```
CUSTOM_API_KEY=<user key>
```

(For arccodex.com it ends up under `CUSTOM_API_KEY` because the host is not in the UI's known-vendor list.)

### What the Agent runtime does

In `hermes_cli/runtime_provider.py`, `_host_derived_api_key()` derives the env-var name **from the Base URL host** and looks up:

```
<VENDOR>_API_KEY
```

For `https://www.arccodex.com/api/codex/v1` it tries to read `ARCCODEX_API_KEY`, which the UI never writes. `CUSTOM_API_KEY` is not checked for this code path, so the runtime sends the request **without an `Authorization` header**, and the upstream returns:

```
Error code: 401 - {'error': 'Arc Codex: Invalid or missing API key. Or your API key is not enabled.
Please login to Arc Codex Dashboard https://www.arccodex.com to check your API key and try again.'}
```

In the UI the key field also appears empty after reopening the Edit Model dialog (it only shows a placeholder), which makes the problem look like “the UI is losing my key” when in fact the key is stored, just under the wrong name.

### Suggested fix

When saving a custom OpenAI-compatible model, write the key under **both** names, or always fall back to the host-derived name that the runtime expects:

```ts
// renderer/src/screens/Models/Models.tsx (sketch)
const host = new URL(baseUrl).hostname;          // www.arccodex.com
const vendor = host.split('.').slice(-2, -1)[0]  // arccodex
                 .toUpperCase();
const envKey = `${vendor}_API_KEY`;              // ARCCODEX_API_KEY

await writeEnv(envKey, apiKey);
await writeEnv('CUSTOM_API_KEY', apiKey); // keep for backward compatibility
```

Alternatively, make the runtime fall back to `CUSTOM_API_KEY` when `<VENDOR>_API_KEY` is missing for non-recognized hosts.

### UX hint

The Edit Model dialog says:

> Stored as an environment variable. Picks the matching env key based on the URL, or CUSTOM_API_KEY otherwise.

This is misleading because the runtime does **not** fall back to `CUSTOM_API_KEY`. Either update the runtime to honor that promise, or update the hint to reflect what really happens.

---

## Bug 2 — Invalid YAML in default `config.yaml` (streaming block)

In the default `config.yaml` shipped with the app, the streaming-related section contains a typo / invalid YAML structure (the `api_key:` line is commented in a way that breaks the parent mapping, and an inline value is duplicated on the same line as the key). This causes the runtime to either ignore the section or, worse, to silently skip the api_key override even when the user does set `ARCCODEX_API_KEY` in `.env`.

Repro: open the bundled `config.yaml`, locate the streaming block, run `python -c "import yaml,sys; yaml.safe_load(open('config.yaml'))"` — it either errors or produces an unexpected structure depending on the exact bytes.

Suggested fix: clean up the streaming block so that:

```yaml
streaming:
  enabled: true
  # api_key: ${ARCCODEX_API_KEY}   # uncomment to override per-provider
```

is valid YAML and the `api_key` override actually works when uncommented.

---

## Workaround for users (until fixed)

1. Open the app's `.env` and add:
```
   ARCCODEX_API_KEY=sk-arc-...your key...
```
   (use the `<VENDOR>_API_KEY` name that matches your base URL host).
2. Fix `config.yaml` so the `streaming:` block is valid YAML; if you want to force the key from there, uncomment `api_key: ${ARCCODEX_API_KEY}`.
3. Fully restart the app (env vars are read at startup).

---

## Severity / impact

- Any user trying to use a **custom OpenAI-compatible provider** that is not in the hard-coded vendor list (Arc Codex, self-hosted vLLM/Ollama behind a custom domain, third-party gateways, etc.) hits this immediately.
- Symptoms (401 + key “disappearing” from the UI) strongly suggest a credentials problem on the provider side, so users are likely to blame the provider rather than the app.

Happy to test a patch against arccodex.com if helpful.

pmos69:
## Summary

Engine compat fix. Upstream `hermes-agent` past v2026.5.16 introduced `hermes_cli/runtime_provider.py::_host_derived_api_key()` which derives the env var name for custom-provider chat from the base URL host (`api.deepseek.com` → `DEEPSEEK_API_KEY`) and refuses to forward `OPENAI_API_KEY` to non-openai hosts. The desktop today only writes `CUSTOM_PROVIDER_<NAME>_KEY` to `.env` and only sets `OPENAI_API_KEY` in the child env at chat time — so users with a custom provider on `api.deepseek.com`/`.groq.com`/`.mistral.ai`/etc. lose chat the moment the bundled engine bumps. Same regression-on-upgrade shape as #360 — engine bump makes the open report visibly worse.

**Strategy:** additive compat — write both env-var forms. Old engine reads `OPENAI_API_KEY` (unchanged); new engine reads `<VENDOR>_API_KEY` via host-derive. Old engine ignores the extra var; new engine ignores the extra `OPENAI_API_KEY` for non-openai hosts (its own leak-prevention). No engine-version detection branch — both engines see one consistent disk + env state.

## Two paths fixed

| Path | Where | Fix |
|---|---|---|
| **Runtime chat spawn (CLI fallback)** | `src/main/hermes.ts::sendMessageViaCli` | After resolving the API key, also set the host-derived env-var in the child process env. |
| **Gateway persistence (primary chat flow)** | `src/main/models.ts::seedDefaults` | When persisting a custom-provider key to `.env`, write both `CUSTOM_PROVIDER_<NAME>_KEY` and the host-derived `<VENDOR>_API_KEY`. The long-running gateway only sees `.env` at startGateway time, so the host-derived form has to live there (not just be set at chat-time). |

Both call sites share a tiny new module `src/main/host-derived-env.ts` (`hostDerivedEnvKeyForUrl(baseUrl)`) so the URL → env-var contract can't drift between the two write sites.

**Guard rails:**
- Doesn't shadow `OPENAI_API_KEY` / `ANTHROPIC_API_KEY` via this path (those belong to canonical providers, not custom)
- Won't overwrite a pre-existing host-derived key (user's real `DEEPSEEK_API_KEY` is safe from a custom-provider seed clobber)
- No-op for unknown vendor hosts (e.g. `api.unsloth.ai` — no false-positive vendor binding)
- Won't propagate `no-key-required` placeholder to a vendor-scoped slot

## Tests

**+19 new tests, all passing. Full suite: 691 passed | 3 skipped.**

- **`tests/compat-host-derived-key.test.ts`** (13 tests) — URL → env-var mapping. Pins the contract for every vendor (deepseek, groq, mistral, together, fireworks, cerebras, perplexity, openrouter, anthropic, openai, huggingface) plus null returns for local LLMs / unknown hosts / case sensitivity.
- **`tests/compat-host-derived-key-persistence.test.ts`** (6 tests) — `seedDefaults` writes both env-var names, doesn't shadow canonical keys, no false-positive vendor binding for unknown hosts, idempotent re-runs don't duplicate entries.

## Live test — verified against BOTH engines

The dual-engine compat rig is documented in `.personal/engine-update-audit.md` Phase 0 (local-only doc). Two separate engine installs:
- `~/AppData/Local/hermes-oldengine` — v2026.5.16 release tag (the version v0.5.1 bundles, pre-host-derive)
- `~/AppData/Local/hermes-newengine` — current upstream main (post-host-derive)

Both run via `$env:HERMES_HOME=...; npm run dev`. The verify script (`scripts/verify-compat-host-derived-key.js`) seeds a fake-key deepseek custom provider, triggers `seedDefaults`, and confirms both env-var names land in `.env`.

### OLD engine (v2026.5.16 — bundled in v0.5.1)
```
HERMES_HOME: C:\Users\pmos6\AppData\Local\hermes-oldengine
Engine:     Hermes Agent v0.14.0 (2026.5.16)
[A2] Stripped pre-existing DEEPSEEK_API_KEY so the fix's write is observable
[B] Appended custom_providers entry to config.yaml
[C] Deleted models.json to force re-seed
[C] listModels picked up entry: id=86de9907-... baseUrl=https://api.deepseek.com/v1

[D] .env contents (relevant lines):
   CUSTOM_PROVIDER_COMPATTESTDEEPSEEK_KEY=sk-fake-deepseek-compat-test-12345
   DEEPSEEK_API_KEY=sk-fake-deepseek-compat-test-12345

✅ PASS (OLD engine): both env-var names persisted to .env.
```

### NEW engine (upstream main, post-host-derive)
```
HERMES_HOME: C:\Users\pmos6\AppData\Local\hermes-newengine
Engine:     Hermes Agent v0.14.0 (2026.5.16)
[A2] Stripped pre-existing DEEPSEEK_API_KEY so the fix's write is observable
[B] Appended custom_providers entry to config.yaml
[C] Deleted models.json to force re-seed
[C] listModels picked up entry: id=b2d1cb5b-... baseUrl=https://api.deepseek.com/v1

[D] .env contents (relevant lines):
   CUSTOM_PROVIDER_COMPATTESTDEEPSEEK_KEY=sk-fake-deepseek-compat-test-12345
   DEEPSEEK_API_KEY=sk-fake-deepseek-compat-test-12345

✅ PASS (NEW engine): both env-var names persisted to .env.
   → The long-running gateway will find DEEPSEEK_API_KEY on next startup.
   → On the new engine, _host_derived_api_key() can resolve it.
```

Independent cross-check: invoked the new engine's `_host_derived_api_key()` directly with `DEEPSEEK_API_KEY=sk-fake-... python -c "from hermes_cli.runtime_provider import _host_derived_api_key; print(_host_derived_api_key('https://api.deepseek.com/v1'))"` → returns the key. Confirms the new engine actually consumes the env var the desktop is writing.

## Why ship this *before* the engine bump

- The CAUTION items in the audit landed an action plan: this is the highest-priority one because it's the only one with a *user-visible regression on upgrade*. Users with a custom DeepSeek/Groq/Mistral/etc. provider lose chat the moment the bundled engine bumps if this isn't already in.
- Pure additive code paths — no behavior change for users on the current bundled engine.
- 0 risk to the existing `CUSTOM_PROVIDER_<NAME>_KEY` flow — every test confirms it stays in place alongside the new write.

Closes part of the engine-bump prep tracked downstream. Doesn't claim to fix #360 but reduces its blast radius on engine upgrade.

---

## Maintainer Merge Note

Suggested position: after #369.

Why: this PR is conceptually part of the same provider key-consistency hardening as #369 and is easiest to reason about once #369's shared URL/env-key resolver and config-health behavior have landed.

### DIFF
--- a/scripts/verify-compat-host-derived-key.js
+++ b/scripts/verify-compat-host-derived-key.js
@@ -1,0 +1,190 @@
+/**
+ * Live-verify the dual-engine compat fix for host-derived <VENDOR>_API_KEY.
+ *
+ * Confirms that when the desktop spawns the hermes-agent subprocess for a
+ * custom-provider chat against a known-vendor host (e.g. api.deepseek.com),
+ * the child env contains BOTH:
+ *   - OPENAI_API_KEY (so old engine v2026.5.16 / v0.14.0 keeps working)
+ *   - <VENDOR>_API_KEY (so post-2026.5.16 engines with
+ *     hermes_cli/runtime_provider.py::_host_derived_api_key find a key)
+ *
+ * Probe strategy: seed a custom_providers.json with a fake-key deepseek
+ * provider, kick a chat via IPC, and capture the gateway error. With the
+ * fix in place:
+ *
+ *   - Old engine (v2026.5.16): reads OPENAI_API_KEY, forwards to
+ *     api.deepseek.com, deepseek returns 401 "Invalid API key" — key
+ *     reached the API.
+ *   - New engine (post-host-derive): reads DEEPSEEK_API_KEY via host-derive,
+ *     forwards to api.deepseek.com, same 401 from deepseek.
+ *
+ * Without the fix on a host-derive-aware engine: chat fails earlier with
+ * "no-key-required" or missing-key error before reaching deepseek.
+ *
+ * Prereqs:
+ *   1. Dev electron running with ENABLE_CDP=1 (port 9222 by default)
+ *   2. Main process built from `fix/compat-host-derived-api-key` branch
+ *   3. For the "old engine" leg: HERMES_HOME=%LocalAppData%/hermes-oldengine
+ *      (set BEFORE `npm run dev` — see .personal/engine-update-audit.md
+ *      Phase 0 for setup instructions)
+ *   4. For the "new engine" leg: HERMES_HOME=%LocalAppData%/hermes-newengine
+ *
+ * Run:   node scripts/verify-compat-host-derived-key.js
+ */
+
+const { chromium } = require("playwright");
+const fs = require("fs");
+const path = require("path");
+const os = require("os");
+
+const FAKE_KEY = "sk-fake-deepseek-compat-test-12345";
+const PROVIDER_NAME = "_compat_test_deepseek";
+const BASE_URL = "https://api.deepseek.com/v1";
+
+async function attach() {
+  const cdpUrl = `http://127.0.0.1:${process.env.CDP_PORT || "9222"}`;
+  const browser = await chromium.connectOverCDP(cdpUrl);
+  const page = browser.contexts()[0].pages()[0];
+  return { browser, page };
+}
+
+(async () => {
+  const { browser, page } = await attach();
+
+  // Phase A — observe which engine + HERMES_HOME the desktop is wired to
+  const home = await page.evaluate(
+    async () => await window.hermesAPI.getHermesHome(),
+  );
+  const engine = await page.evaluate(
+    async () => await window.hermesAPI.getHermesVersion(),
+  );
+  console.log("HERMES_HOME:", home);
+  console.log("Engine version:", engine);
+  const isOldEngine = home.toLowerCase().includes("hermes-oldengine");
+  const isNewEngine = home.toLowerCase().includes("hermes-newengine");
+  const flavor = isOldEngine ? "OLD" : isNewEngine ? "NEW" : "DEFAULT";
+  console.log(`Test leg: ${flavor}`);
+
+  // Phase B — seed a custom provider via direct JSON write (no add-provider
+  // IPC exists in the renderer surface; the desktop loads custom_providers.json
+  // on the next listModels call).
+  const cpFile = path.join(home, "custom_providers.json");
+  let cps = [];
+  if (fs.existsSync(cpFile)) {
+    try {
+      cps = JSON.parse(fs.readFileSync(cpFile, "utf-8"));
+    } catch {
+      cps = [];
+    }
+  }
+  const existing = cps.findIndex((c) => c.name === PROVIDER_NAME);
+  const entry = {
+    name: PROVIDER_NAME,
+    provider: "custom",
+    model: "deepseek-chat",
+    baseUrl: BASE_URL,
+    apiKey: FAKE_KEY,
+    apiMode: null,
+  };
+  if (existing >= 0) cps[existing] = entry;
+  else cps.push(entry);
+  fs.writeFileSync(cpFile, JSON.stringify(cps, null, 2));
+  console.log(`[B] Seeded ${PROVIDER_NAME} → ${BASE_URL} into ${cpFile}`);
+
+  // Trigger a listModels so the desktop picks up the new entry and writes
+  // the CUSTOM_PROVIDER_<NAME>_KEY env var into .env.
+  const modelsAfter = await page.evaluate(
+    async () => await window.hermesAPI.listModels(),
+  );
+  const seeded = modelsAfter.find((m) => m.name === PROVIDER_NAME);
+  console.log(
+    "[B] listModels picked up entry:",
+    seeded ? `id=${seeded.id}` : "NO (failed to seed)",
+  );
+
+  // Phase C — activate the provider + send chat
+  console.log("\n[C] Activating provider + sending probe chat…");
+  await page.evaluate(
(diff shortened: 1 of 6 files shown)


fathah:
conflict

pmos69:
@greptileai please review the rebased branch when you get a chance. This was updated onto current main after conflict resolution, with the focused compatibility tests and typecheck passing locally.

pmos69:
@greptileai please re-review the latest update when you get a chance. I addressed your finding by skipping the no-key-required sentinel in the persistence path as well, and added a regression test for it. Focused tests and typecheck pass locally.

PR #402, opened 2026-05-27
Expose local model providers in model setup

### LINKED ISSUE #362: How we setup the ollama ?
Hi,

I dont find in settings where we can setup the ollama, to use the ollama models there! how i cant load it?

### LINKED ISSUE #500: # Bug: Model discovery broken for local providers (Ollama, LM Studio) in v0.5.2
**Repo:** https://github.com/fathah/hermes-desktop
**Affected version:** v0.5.2
**Working version:** v0.4.5
**Component:** `src/main/model-discovery.ts`

## Description

After updating the Hermes Desktop shell from v0.4.5 to v0.5.2, local model providers that don't require an API key (Ollama, LM Studio, vLLM, llama.cpp, custom) stop showing models. The Model field in the provider configuration dropdown remains empty or broken.

## Root cause

The new `model-discovery.ts` file (introduced in v0.5.2) performs model detection by sending HTTP requests to the provider's `/models` endpoint. The logic unconditionally checks for an API key before making the request:

```
1. Resolve provider baseUrl (from settings or PROVIDER_BASE_URLS table)
2. Resolve API key (from override OR from .env)
3. IF API key is empty → return { status: "no-key", models: [] }
```

Step 3 returns an empty model list for **all** providers when the API key is empty — including local providers that don't need one (Ollama, LM Studio, vLLM, llama.cpp, custom).

## Evidence

In `src/main/providers.ts`, there is already a defined set of keyless providers:

```typescript
export const PROVIDERS_WITHOUT_API_KEYS = new Set([
  "custom",
  "lmstudio",
  "ollama",
  "vllm",
  "llamacpp",
  "openai-codex",
]);
```

However, `model-discovery.ts` never references `PROVIDERS_WITHOUT_API_KEYS`. It treats every provider identically — requiring an API key even for local-only providers.

## Reproduction steps

1. Install hermes-desktop v0.5.2
2. Run Ollama locally on http://127.0.0.1:11434
3. Set provider to "custom" with base_url: http://127.0.0.1:11434/v1
4. Leave API key empty (as expected for local Ollama)
5. Observe: Model field shows no models or errors out

Same applies to LM Studio (http://127.0.0.1:1234/v1).

## Proposed fix

In `src/main/model-discovery.ts`, before returning `{ status: "no-key", models: [] }`, check if the provider type belongs to `PROVIDERS_WITHOUT_API_KEYS`. If so, skip the API key check and proceed with the HTTP request without an Authorization header:

```typescript
if (!apiKey && !PROVIDERS_WITHOUT_API_KEYS.has(provider)) {
  return { status: "no-key", models: [] };
}
```

## Files involved

- `src/main/model-discovery.ts` — new detection logic introduced in v0.5.2
- `src/main/providers.ts` — contains `PROVIDERS_WITHOUT_API_KEYS` set (already defined, unused in detection)

pmos69:
## Summary
- add explicit local provider choices for LM Studio, Ollama, vLLM, and llama.cpp in model setup
- prefill canonical local base URLs and auto-detect known local provider ports
- allow local provider model discovery without API keys and report unreachable local servers as discovery errors instead of empty model lists

## Testing
- npm test -- tests/constants.test.ts tests/detect-provider.test.ts tests/provider-registry.test.ts tests/set-model-config-base-url.test.ts tests/model-discovery.test.ts
- npm run typecheck
- npm run build
- npm test
- live tested with LM Studio local server on localhost:1234 and verified model discovery/population in the desktop UI

Closes #362
Closes #500

Refs #423 — this improves the local Ollama/local-provider setup path called out in that report. The specific session-continuation `API_SERVER_KEY` failure is handled by #369.

---

## Maintainer Merge Note

Suggested position: after #400.

No known stacked conflict from the pre-release rebuild. This is mostly independent, but it is cleaner after the provider/key consistency work because it exposes local-provider setup paths that rely on the same model/provider configuration surface.

### DIFF
--- a/src/main/model-discovery.ts
+++ b/src/main/model-discovery.ts
@@ -246,6 +246,13 @@ const _cache = new Map<string, CacheEntry>();
 // with the main cache via `_clearCache` for test isolation.
 const _freeCache = new Map<string, string[]>();
 
+const LOCAL_NO_KEY_PROVIDERS = new Set([
+  "lmstudio",
+  "ollama",
+  "vllm",
+  "llamacpp",
+]);
+
 function cacheKey(provider: string, baseUrl: string): string {
   return `${provider.toLowerCase()}|${baseUrl.replace(/\/+$/, "").toLowerCase()}`;
 }
@@ -330,7 +337,13 @@ function buildUrl(base: string): string {
   return `${trimmed}/models`;
 }
 
+interface FetchModelsResult {
+  models: string[];
+  reachable: boolean;
+}
+
 function authHeaders(provider: string, apiKey: string): Record<string, string> {
+  if (!apiKey) return {};
   const lower = provider.toLowerCase();
   if (lower === "anthropic") {
     // Anthropic uses x-api-key + an API version header on /v1/models.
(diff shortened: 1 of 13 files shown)


pmos69:
Greptile, could you please re-review this PR? I pushed updates addressing the local-provider base URL persistence issue and the Atomic Chat backend registry/no-key discovery gap.

pmos69:
@greptileai please re-review the latest PR head. The previous review was on \\634345\\; the current head is \\6151d22\\ and addresses those findings.

Issue #408, opened 2026-05-27
Suggestion for Conversation Deletion Function


jasoncow007:
Looking forward to the implementation of the conversation deletion feature in Hermes Desktop.

PR #404, opened 2026-05-27
fix(linux): set SUID on chrome-sandbox in .deb/.rpm postinst (#395)

pmos69:
## Problem

Closes #395.

Reporter on Ubuntu 26.04 LTS installs the `.deb` v0.5.1 and observes Electron crashing on close. Inspection of the installed file shows:

```
$ ls -la "/opt/Hermes Agent/chrome-sandbox"
-rwxr-xr-x 1 root root 15328 May 24 22:46 chrome-sandbox
```

No SUID bit. The reporter's workaround (`chmod 4755`) confirmed the diagnosis.

## Why this happens

Electron's setuid sandbox helper needs **either**:
1. The SUID root bit on `chrome-sandbox`, or
2. `kernel.unprivileged_userns_clone=1` (unprivileged user namespaces enabled)

Modern Ubuntu (24.04+) ships AppArmor + kernel hardening that disable unprivileged user namespaces by default, so the SUID path becomes mandatory. Same story on Fedora 40+.

electron-builder doesn't ship this postinst step by default — it has to be wired up via `{deb,rpm}.afterInstall`. Other Electron apps (VS Code, Slack, Discord, Signal, …) all do the same thing.

## Fix

Three small changes:

1. **`build/linux-after-install.sh`** (new) — `chmod 4755` on `/opt/Hermes Agent/chrome-sandbox`. Guarded by `[ -f ]` so it's a no-op if the file isn't there. Set executable in the git index (`git update-index --chmod=+x`) so the .deb and .rpm builders preserve the bit.

2. **`electron-builder.yml`** — wire the script into both `deb:` and `rpm:` target sections via `afterInstall`. AppImage doesn't need this (self-mounting, different sandbox path); snap doesn't either (its own confinement).

3. **`.gitattributes`** — pin `*.sh` to `eol=lf`. CRLF on a shell script breaks the shebang on Linux (`bad interpreter: /bin/bash\r: No such file or directory`), which would silently fail the postinst hook. Defensive against future Windows-side edits.

## Verification

This change only takes effect on `.deb` / `.rpm` built from this commit, and end-to-end verification requires Ubuntu 26.04 (or similar hardened distro) + the resulting package. I'm on Windows so I can't run that myself.

Once this lands and a release goes out, the reporter on #395 can confirm with:

```bash
ls -la "/opt/Hermes Agent/chrome-sandbox"
# → -rwsr-xr-x   (the 's' in place of 'x' for the owner is the SUID bit)
```

And the close-crash should disappear.

[...]

PR #415, opened 2026-05-27
fix(chat): compress oversize images + set Content-Length on chat POSTs (#405)

### LINKED ISSUE #405: [Bug]: Desktop TUI image upload returns Invalid JSON in request body
## Bug Description
Desktop TUI sends images that cause `400 Invalid JSON in request body` error.

## Environment
- Platform: macOS (Apple Silicon)
- hermes-agent: latest
- Connection: Desktop TUI / WebUI -> API Server

## Steps to Reproduce
1. Paste or send an image in Desktop TUI
2. Error: `Error: Invalid JSON in request body`

## Root Cause
Likely in the frontend (hermes-desktop), not the backend API server.

Another frontend client successfully parses and sends images. The issue appears to be in how this frontend encodes images before sending to the API.

## Expected Behavior
Should be able to send images to agent for vision processing.

pmos69:
## Problem

Closes #405.

Lemonseaa reports that pasting an image in Hermes Desktop returns `400 Invalid JSON in request body`. The gateway's error message is misleading — the real cause is the **request body is too large** for the gateway's 10 MB cap (`MAX_REQUEST_BYTES`), but two compounding bugs hide the real failure mode from the user.

## Diagnosis

After live reproduction against an aiohttp instance with the same `client_max_size=10_000_000`:

| Body size + framing | Status | Error |
|---|---|---|
| 5 MB body, Content-Length set | 200 | OK |
| 12 MB body, Content-Length set | 413 | `body_too_large` (correct) |
| 12 MB body, chunked encoding | **400** | **`Invalid JSON in request body`** ← #405 symptom |

The chunked path is the bug. `hermes.ts` was using `req.write(string); req.end();` which Node frames as `Transfer-Encoding: chunked` with **no Content-Length**. The gateway's `body_limit_middleware` only checks Content-Length, so it became a no-op for the desktop's requests, and oversized bodies fell through to aiohttp's `client_max_size` overflow — caught by a broad `except (json.JSONDecodeError, Exception):` that returns the misleading "Invalid JSON" 400.

Compounding it: the desktop accepted up to 20 MB images even though base64 inflation puts the encoded body well past the 10 MB gateway cap. So any photo over ~7 MB hit this consistently.

### Wire-level (the actual #405-causing bug)

**`src/main/hermes.ts`** — encode the JSON body to a `Buffer` and set `Content-Length` explicitly on the request headers. Done for both the streaming chat path and the non-streaming probe-error path. Node won't fall back to chunked once Content-Length is present.

Verified live via Node-level capture:
- Pre-fix pattern (`req.write(string)`): no Content-Length, `Transfer-Encoding: chunked` — reproduces the bug
- Post-fix pattern (`req.write(buffer)` + explicit CL): Content-Length set correctly, no chunked framing → gateway's size middleware now catches oversized payloads with a clean 413

### Renderer-level (avoid hitting the cap in the first place)

**`src/renderer/src/screens/Chat/attachmentUtils.ts`** — new `compressImageToFit(file, targetBytes)` using HTML Canvas. Iteratively drops JPEG quality (0.85 → 0.5 in 0.15 steps), then scales dimensions down by 20% per round until under cap. Short-circuits and returns the original `File` reference if `file.size <= targetBytes`, so the 99% of pastes that don't need compression incur zero decode/re-encode overhead.

- **Opaque images** → re-encoded as JPEG (huge size win)
- **Images with alpha** (detected via strided pixel sampling) → kept as PNG, downsampled only
- **Animated GIFs** → rejected with the new `image-uncompressible` error (canvas would only capture the first frame; silently emitting a static thumbnail would be worse UX than a clear error)

**`src/shared/attachments.ts`** constants:
- `MAX_IMAGE_INPUT_BYTES = 50 MB` — sanity cap on input (only rejects pathologically large files; was 20 MB and rejected immediately)
- `MAX_IMAGE_TARGET_BYTES = 5 MB` — compression target (sized so `× 4/3` base64 inflation fits the gateway's 10 MB cap with headroom for history)

**`AttachmentChip.tsx`** — tooltip now shows "12 MB → 4.2 MB, compressed" when compression ran, so users know quality changed.

**i18n** — `attachImageTooLarge` cap bumped to 50 MB; new `attachImageUncompressible` string. 7 locales (en, es, id, ja, pt-BR, pt-PT, zh-CN).

## Testing

- **Unit tests:** short-circuit path (1 MB image passes through untouched, no `originalSize` set), input-cap path (>50 MB rejected without attempting compression). Existing tests for text/path-ref attachments still pass.
- **Wire-level live test:** confirmed via Node http capture that the post-fix `req.write(Buffer) + Content-Length` pattern lands a real `Content-Length` header and disables chunked encoding. Repro file: `.personal/verify-405-content-length.js`.
- **Compression algorithm:** the canvas-based loop isn't testable from JSDOM (no real canvas), so it's covered by visual inspection in the dev electron + integration via the wire-level path. If the algorithm has a bug, the worst case is the new Content-Length still gives a clean 413 — strict improvement over the pre-fix state.

## Upstream follow-up (separate, non-blocking)

The gateway's `except (json.JSONDecodeError, Exception):` clause should be split so future clients get a descriptive error rather than the "Invalid JSON" red herring. Worth filing on `NousResearch/hermes-agent` — but this desktop fix doesn't depend on it.

## CI flake note

Two tests (`Skills.test.tsx > surfaces the CLI error...` and `locale-persistence.test.ts > reloads the saved locale...`) are flaky on full-suite runs and pass in isolation. Same flake documented on PRs #403 and #404. Unrelated to packaging or this PR's surface.

## Side-effect: clears up "empty conversation" noise in the Sessions list

A separate user-reported puzzle was that the Sessions tab fills up with "0 msgs" rows tagged `New conversation` after a chat session. Diagnosis traced this to `agent/conversation_loop.py:2606-2628` in the engine: **when the gateway returns 413 Payload Too Large, the engine triggers `_compress_context` as recovery, which closes the running session with `end_reason='compression'` and spawns a fresh session ID**. Every pre-fix oversized-image attempt produced one of these compression-tagged rows.

Post-this-PR, the 413 cascade is gone (Content-Length is set so the gateway can actually decide if it's too big; image compression keeps the body under the cap anyway), so the conversation-compression-as-413-recovery code path stops firing for image attachments. Users who repeatedly attempt big image attachments today see a state.db full of these orphan rows; on the fix they won't.

(Existing rows in state.db from before this lands aren't cleaned up automatically — a follow-up PR for #408 adds an explicit delete button to the Sessions tab for that.)

---

## Maintainer Merge Note

Suggested position: before #422.

Known stacked overlap: #422 also touches prompt attachment rendering. Merging #415 first lets #422 keep the compression tooltip/metadata behavior while adding prompt-image restore/preview behavior.

### DIFF
--- a/src/main/hermes.ts
+++ b/src/main/hermes.ts
@@ -401,8 +401,22 @@ function sendMessageViaApi(
     ...(_resumeSessionId ? { session_id: _resumeSessionId } : {}),
   });
 
+  // Encode the body up-front into a Buffer so we can:
+  //  1. Set `Content-Length` accurately based on byte length (NOT char
+  //     count — JSON.stringify of an image data URL is ASCII so they
+  //     match, but multi-byte chars in user text would diverge).
+  //  2. Disable Node's default `Transfer-Encoding: chunked` framing for
+  //     bodies written via `req.write(body); req.end();`. Chunked
+  //     framing skips the gateway's `body_limit_middleware` (which
+  //     inspects Content-Length only), so an oversized payload that
+  //     should produce a clean 413 "body_too_large" gets the
+  //     misleading 400 "Invalid JSON in request body" via aiohttp's
+  //     client_max_size overflow path. See #405.
+  const bodyBuf = Buffer.from(body, "utf-8");
+
   const headers: Record<string, string> = {
     "Content-Type": "application/json",
+    "Content-Length": String(bodyBuf.length),
     ...getRemoteAuthHeader(),
   };
   // Local API server key (API_SERVER_KEY in the profile's .env /
@@ -479,11 +493,20 @@ function sendMessageViaApi(
       messages: [{ role: "user", content: userContent }],
       stream: false,
     });
+    const probeBodyBuf = Buffer.from(probeBody, "utf-8");
+    // Per-request Content-Length (the outer `headers` object's value
+    // belongs to the streaming request — reusing it here would lie about
+    // this body's size and break the framing the same way the missing
+    // Content-Length did before #405). Spread + override.
+    const probeHeaders = {
+      ...headers,
+      "Content-Length": String(probeBodyBuf.length),
+    };
     const probeUrl = `${getApiUrl()}/v1/chat/completions`;
     const probeMod = probeUrl.startsWith("https") ? https : http;
     const probeReq = probeMod.request(
       probeUrl,
-      { method: "POST", headers },
+      { method: "POST", headers: probeHeaders },
       (res) => {
         let raw = "";
         res.on("data", (d) => {
@@ -512,7 +535,7 @@ function sendMessageViaApi(
         "No response received from the model. Check your model configuration and API key.",
       );
     });
-    probeReq.write(probeBody);
+    probeReq.write(probeBodyBuf);
     probeReq.end();
   }
 
@@ -689,7 +712,7 @@ function sendMessageViaApi(
     );
   });
 
-  req.write(body);
+  req.write(bodyBuf);
   req.end();
 
   return {
--- a/src/renderer/src/components/AttachmentChip.tsx
+++ b/src/renderer/src/components/AttachmentChip.tsx
@@ -16,10 +16,19 @@ export function AttachmentChip({
 }: AttachmentChipProps): React.JSX.Element {
   const isImage = attachment.kind === "image";
 
+  // When the renderer compressed the image down to fit the gateway's
+  // request-body cap (#405), surface the size delta in the tooltip so the
+  // user knows quality changed and isn't surprised by a "compressed"
+  // version appearing in the chat transcript.
+  const tooltip =
+    attachment.originalSize && attachment.originalSize > attachment.size
+      ? `${attachment.name} (${formatSize(attachment.originalSize)} → ${formatSize(attachment.size)}, compressed)`
+      : `${attachment.name} (${formatSize(attachment.size)})`;
+
   return (
     <div
       className={`attachment-chip attachment-chip-${attachment.kind}`}
-      title={`${attachment.name} (${formatSize(attachment.size)})`}
+      title={tooltip}
     >
       {isImage && attachment.dataUrl ? (
         <button
--- a/src/renderer/src/screens/Chat/ChatInput.tsx
+++ b/src/renderer/src/screens/Chat/ChatInput.tsx
@@ -92,6 +92,8 @@ export const ChatInput = forwardRef<ChatInputHandle, ChatInputProps>(
             return t("chat.attachTooMany");
           case "image-too-large":
             return t("chat.attachImageTooLarge", { name: err.filename });
+          case "image-uncompressible":
+            return t("chat.attachImageUncompressible", { name: err.filename });
           case "text-too-large":
             return t("chat.attachTextTooLarge", { name: err.filename });
           case "unsupported-type":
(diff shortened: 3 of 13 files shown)


pmos69:
**Refined to try both encoders, pick smaller** (commit `68f2bc1`).

Previous heuristic was WebP-preferred-with-JPEG-fallback. Better: encode each iteration to **both** WebP and JPEG (PNG only for alpha-bearing images) and pick the smallest blob. ~2× encoding time per iteration but the wire payload is guaranteed optimal regardless of content type — WebP wins on screenshots/photos, JPEG marginally wins on high-frequency noise/grain, and we never pay the loser's cost.

### Final live size sweep (try-both)

| Target | Input MB | Compressed | Output MB | Output MIME    | Error            | Time   |
|--------|----------|------------|-----------|----------------|------------------|--------|
|   1 MB |     0.86 | no         |      0.86 | image/png      | —                |   35ms |
|   3 MB |     2.58 | no         |      2.58 | image/png      | —                |   81ms |
|   5 MB |      4.3 | no         |       4.3 | image/png      | —                |  116ms |
|   7 MB |     6.02 | yes        |      1.26 | image/jpeg ←   | —                |  836ms |
|  10 MB |      8.6 | yes        |       1.8 | image/jpeg ←   | —                | 1138ms |
|  15 MB |    12.89 | yes        |      2.68 | image/jpeg ←   | —                | 1770ms |
|  25 MB |     21.5 | yes        |      4.45 | image/jpeg ←   | —                | 3347ms |
|  60 MB |    51.58 | no         |         — | —              | image-too-large  | 1628ms |

← The synthetic test data is random noise — algorithm correctly selected JPEG since it marginally beats WebP on that pattern. On real photos/screenshots the same algorithm would pick WebP. **The user always gets the smallest possible binary.**

The "image/jpeg" output column here is the algorithm's choice for *this* test data, not a hardcoded format. The candidate list is `[image/webp, image/jpeg]` per iteration; whichever produces the smaller blob wins.

### Timing comparison across the three approaches

| Approach (this PR's history) | 6 MB | 8.6 MB | 12.9 MB | 21.5 MB |
|------------------------------|-------|--------|---------|---------|
| JPEG-only                    | 582ms | 759ms  | 1065ms  | 1853ms  |
| WebP-only                    | 1112ms| 1468ms | 1971ms  | 3352ms  |
| **Try-both (final)**         | 836ms | 1138ms | 1770ms  | 3347ms  |

Try-both is sometimes *faster* than WebP-only because the smaller candidate fits at an earlier quality step, ending the loop sooner. Slowest case (25 MB → 4.5 MB at q=0.85) takes ~3.3s — acceptable for what's effectively a one-shot operation in response to a deliberate user paste.

### Summary of #405 fix surface

| Layer | Change | Effect |
|---|---|---|
| Wire framing | Buffer body + explicit `Content-Length` in `hermes.ts` | Stops Node's chunked-transfer default; gateway middleware now sees CL and returns clean 413 instead of misleading 400 |
| Renderer compression | `compressImageToFit` with try-both WebP/JPEG, PNG for alpha | Oversized images shrink to fit the gateway cap. Smallest possible binary per content type |
| Constants | `MAX_IMAGE_INPUT_BYTES = 50 MB`, `MAX_IMAGE_TARGET_BYTES = 5 MB` | Generous input cap, conservative encoded-size target with headroom inside the 10 MB gateway cap |
| UI | `AttachmentChip` shows "original → compressed" in tooltip when compression ran | Users know quality changed |
| i18n | 7 locales updated | `attachImageUncompressible` for animated GIFs / decode failures; cap bumped |

PR #427, opened 2026-05-28
feat(local): add Atomic Chat local provider preset

yanalialiuk:
## Summary
- add an `Atomic Chat` local preset in setup/provider constants with default base URL `http://localhost:1337/v1`
- include `Atomic Chat` in English locale preset labels so it appears next to LM Studio/Ollama in the Local list
- extend local provider URL detection and tests to treat port `1337` as a local OpenAI-compatible endpoint
- update README local preset references to include Atomic Chat

## Test plan
- [x] Run `npm run test -- detect-provider.test.ts`
- [ ] Open Setup screen and verify `Atomic Chat` appears in Local presets
- [ ] Select `Atomic Chat` and verify Base URL autofills to `http://localhost:1337/v1`

Made with [Cursor](https://cursor.com)

PR #428, opened 2026-05-28
fix(i18n): add Atomic Chat preset keys across locales

yanalialiuk:
## Summary
- add missing `atomicchat` local preset key to `setup.localPresets` for non-English locales (zh-TW, zh-CN, pt-PT, pt-BR, ja, id, es)
- add matching `constants.atomicchat` key in the same locale bundles
- avoid fallback/missing-label behavior in the Local provider presets list for translated UI

## Test plan
- [x] Confirm keys exist in all locale `setup.ts` and `constants.ts` files
- [ ] Manually switch app language and verify `Atomic Chat` appears in Local presets


yanalialiuk:
This PR is a follow-up to #427 and only adds missing i18n keys for the Atomic Chat local preset across non-English locales.

PR #431, opened 2026-05-28
Fix: drop streamed bubble when DB split it into multiple rows (dup message bug)

addhass:
## Summary

Fixes a "same message rendered twice" bug that triggers when an agent turn produces multiple assistant text bubbles separated by a tool call.

## Reproduction

1. Agent emits text → calls a tool → emits more text → finalises turn.
2. Stream finishes; `onChatDone` calls `reconcileStreamedWithDb`.
3. The chat transcript shows the second half of the agent's reply twice — once at the end of the streamed bubble and again as a standalone bubble below it.

Visually it looks like the assistant copy-pasted itself, but the API only generated one response (verified against `agent.log` — single `chat_completion_stream_request` per turn).

## Root cause

`main/index.ts:824` forwards every `chat-chunk` delta with no boundary marker between assistant rows. `useChatIPC.ts:onChatChunk` always appends to the last agent bubble, so a multi-bubble turn collapses into one streamed bubble containing the concatenated text of all rows.

At end-of-stream, `reconcileStreamedWithDb` runs against the DB's canonical split (N assistant rows interleaved with tool_call / tool_result):

- `reconciliationKey` keys on `role + first 200 chars`, so it matches the concatenated streamed bubble against the **first** DB row only.
- The remaining DB rows fall through to `result` as DB-id bubbles.
- The tail-of-stream pass (lines 246–260) checks `consumedIds` and a content-key dedup set built from `result`. The streamed bubble's id is consumed (good), but its full concatenated content **isn't** in `seenBubbleKeys` because no single `result` row matches it — so `mergeDbMetadataIntoStreamed` returned the streamed copy and put it back into `result`, and the tail pass then *also* re-adds the streamed bubble because… wait, no — the streamed bubble was consumed via the FIFO match, so the tail loop's `consumedIds.has(m.id)` check should skip it.

Re-reading the loop: the streamed bubble IS preserved (it carries the original streamed id), so the user sees the long streamed bubble PLUS the second-and-later DB rows that fell through with no streamed counterpart. The second DB row's text equals the tail of the streamed bubble's content → duplicate text on screen.

## Fix

In the tail pass, when a streamed bubble's normalised content contains **two or more** DB-bubble contents (of the same role) as substrings, drop the streamed bubble. The DB split is the canonical view; the concatenated streamed copy is a streaming artifact and the user should see the canonical split instead.

The two-or-more threshold matters: a single substring match is the normal "streamed bubble = single DB row" case the existing FIFO match already handles, and we shouldn't disturb it.

## Test

Added `sessionHistory.test.ts` covering the exact scenario (one streamed bubble = concat of partA + partB, DB has partA / tool_call / tool_result / partB). The test asserts `partB` appears exactly once in the reconciled transcript. Fails on `main`, passes with this change.

Full suite: **673 passed, 3 skipped** (same as `main`).

## Limitations / notes

- The fix is in the renderer only — the underlying gap is that `chat-chunk` has no row boundary marker. A more principled fix would have the main process emit a sentinel between rows (e.g. `chat-bubble-boundary`), letting the renderer start a new bubble at the right place during the stream. Happy to follow up with that as a separate PR if you'd prefer the more invasive fix.
- The dedup is substring-based on normalised whitespace. It will not fire if the streamed bubble drifted from the DB rows in a way the normaliser doesn't collapse (very unusual — both sides see the same agent output).

## Triggering scenario (for the changelog)

Most likely to show up when the agent does a "say something → call a tool → say more" flow inside a single turn. With reasoning models / tool-heavy workflows this is common, which is why the bug is conspicuous.

### DIFF
--- a/src/renderer/src/screens/Chat/sessionHistory.test.ts
+++ b/src/renderer/src/screens/Chat/sessionHistory.test.ts
@@ -1,0 +1,56 @@
+import { describe, it, expect } from "vitest";
+import { reconcileStreamedWithDb } from "./sessionHistory";
+import type { ChatMessage } from "./types";
+
+describe("reconcileStreamedWithDb — multi-bubble turn (dup bug)", () => {
+  it("does not produce duplicate text when one streamed bubble concatenates multiple DB assistant rows", () => {
+    // Scenario: agent's turn produced 2 assistant text bubbles split by a tool call.
+    // Stream concatenated both into a single bubble (because chat-chunk has no
+    // boundary marker). At end-of-stream, DB has 2 separate assistant rows + tool
+    // rows in between. Reconciliation must NOT leave both the concatenated streamed
+    // bubble AND the per-row DB bubbles in the result — that's the visible dup.
+
+    const partA = "First paragraph from before the tool call.";
+    const partB = "Second paragraph after the tool result.";
+
+    const streamed: ChatMessage[] = [
+      { id: "user-1", role: "user", content: "do the thing" },
+      // single concatenated streamed agent bubble
+      { id: "agent-1", role: "agent", content: `${partA}\n\n${partB}` },
+    ];
+
+    const db: ChatMessage[] = [
+      { id: "db-1", role: "user", content: "do the thing" },
+      { id: "db-2", role: "agent", content: partA },
+      {
+        id: "db-tc-3-call_1",
+        kind: "tool_call",
+        role: "agent",
+        callId: "call_1",
+        name: "terminal",
+        args: "{}",
+      },
+      {
+        id: "db-tr-4",
+        kind: "tool_result",
+        role: "agent",
+        callId: "call_1",
+        name: "terminal",
+        content: "ok",
+      },
+      { id: "db-5", role: "agent", content: partB },
+    ];
+
+    const result = reconcileStreamedWithDb(streamed, db);
+
+    // Count occurrences of partB text in the merged transcript.
+    const partBOccurrences = result.filter(
+      (m) =>
+        !("kind" in m) &&
+        typeof (m as { content?: string }).content === "string" &&
+        (m as { content: string }).content.includes(partB),
+    ).length;
+
+    expect(partBOccurrences).toBe(1);
+  });
+});
--- a/src/renderer/src/screens/Chat/sessionHistory.ts
+++ b/src/renderer/src/screens/Chat/sessionHistory.ts
@@ -243,6 +243,27 @@ export function reconcileStreamedWithDb(
     }
   }
 
+  // Collect normalised DB-bubble texts grouped by role so we can detect the
+  // "concatenated streamed bubble vs split DB rows" case below. The desktop's
+  // chat-chunk stream has no boundary marker (main/index.ts forwards every
+  // delta verbatim), so when one agent turn produces multiple text bubbles
+  // split by a tool call, all the deltas land in a single streamed bubble.
+  // The reconciliation pass above keys on the first 200 chars, so it matches
+  // that streamed bubble against the FIRST db row only — and the remaining
+  // db rows fall through here, while the streamed bubble (still carrying the
+  // full concatenated text) also wants to fall through below. That's the
+  // visible "same text twice" bug.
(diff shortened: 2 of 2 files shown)


pmos69 (changes requested):
Thanks for taking this on — this is the right area, but I think this fix still misses one branch of the repro.

**Blocking finding**

- `src/renderer/src/screens/Chat/sessionHistory.ts:267`: the multi-bubble dedup only runs in the tail pass, after `if (consumedIds.has(m.id)) continue`. If the first DB assistant row is long enough to share the first 200 normalized chars with the concatenated streamed bubble, the main DB walk consumes that streamed bubble for the first DB row. At that point `consumedIds` contains the streamed id, so the new substring guard never runs, and the result still contains the concatenated streamed bubble plus the later DB assistant row. That leaves the tail text duplicated.

I confirmed with this additional regression shape:

```ts
const partA = "A".repeat(240);
const partB = "Second paragraph after the tool result.";
const streamed = [
  { id: "user-1", role: "user", content: "do the thing" },
  { id: "agent-1", role: "agent", content: `${partA}\n\n${partB}` },
];
const db = [
  { id: "db-1", role: "user", content: "do the thing" },
  { id: "db-2", role: "agent", content: partA },
  { id: "db-tc-3-call_1", kind: "tool_call", role: "agent", callId: "call_1", name: "terminal", args: "{}" },
  { id: "db-tr-4", kind: "tool_result", role: "agent", callId: "call_1", name: "terminal", content: "ok" },
  { id: "db-5", role: "agent", content: partB },
];
```

On this branch, counting bubbles containing `partB` returns `2`, not `1`.

The guard probably needs to apply before accepting a streamed match in the DB walk, or the reconciliation needs to detect that a matched streamed bubble covers multiple same-role DB bubble segments and prefer the DB split instead of `mergeDbMetadataIntoStreamed` for that row.

Validation I ran:

- `npm test -- --run src/renderer/src/screens/Chat/sessionHistory.test.ts` passes for the current test.
- The temporary long-`partA` regression above fails as described.

pmos69:
Could you make #422 the base/prerequisite for this PR, or otherwise clearly list #422 as required before #431?

This PR overlaps the same `sessionHistory.ts` reconciliation path that #419/#421/#422 already change, and #422 is the top of that stack:

- #419 changes the streamed/DB prefix-suffix reconciliation for cold gateway/session split cases.
- #421 adds normalization for screenshot/image placeholders.
- #422 extends that path to preserve restored prompt images.

Because #431 is written against the older/main reconciliation shape, it will likely conflict or behave differently once the #419/#421/#422 stack lands. Porting it onto #422 should make the final behavior easier to reason about and avoid the maintainer having to resolve overlapping chat-history fixes manually.

There is also a missed edge case in the current #431 logic: if the first DB assistant segment is long enough to match the streamed concatenated assistant bubble during the main DB walk, the streamed bubble is consumed before the new tail-pass duplicate guard runs. A regression test should cover the “long first DB segment + second DB segment contained in one streamed bubble” case after porting.

pmos69:
Thanks for putting this together and for nailing the core failure shape. The issue is indeed that the live `chat-chunk` stream can collapse multiple assistant text rows into one renderer bubble, while the end-of-stream DB refresh returns the canonical split rows around tool call/result rows.

I’m going to close this PR as superseded by #508.

The replacement keeps your intent, but rebases it onto the current post-v0.5.2 reconciliation code and tightens the guardrails:

- it groups DB assistant rows by assistant turn, bounded by user rows, instead of checking against all assistant rows globally;
- it only drops suffix/tail streamed artifacts, so #419’s preserved split-session prefix behavior remains intact;
- it requires two or more same-turn DB assistant segments to cover most of the streamed bubble;
- it adds broader coverage for long-chat repeats and false positives, including an assistant answer that quotes rows from different earlier turns.

I also reproduced the failing shape before the fix, and live-tested the replacement via the #383 CDP harness with a manual scenario that shows the unfixed duplicate counts (2/2) and the patched counts (1/1).

Really appreciate the original diagnosis here - #508 is intended as the current-main version of this fix rather than a disagreement with the direction.

PR #444, opened 2026-05-29
chore(testing): add tester build scripts and guide

0r0b0r011:
## Summary

- Adds tester-facing build/packaging scripts (`scripts/build-for-tester.sh`, `scripts/package-for-tester.sh`, `scripts/quick-test-build.sh`)
- Adds `TESTING-GUIDE.md` with step-by-step instructions for non-developer testers (pre-built releases and local build paths)

This is the first slice of a larger feature branch split. It contains **only** tester build tooling — no app source, dependency, or CI changes.

## Test plan

- [ ] `npm run typecheck`
- [ ] `npm run build`
- [ ] Manually run `./scripts/quick-test-build.sh` (or `./scripts/build-for-tester.sh` / `./scripts/package-for-tester.sh`) and confirm artifacts are produced as documented in `TESTING-GUIDE.md`

## Follow-up PRs (planned)

- Native module CI (optional, `.github/workflows/ci.yml`)
- Embedded terminal
- Vault/credentials
- Files/workspace browser
- Profile wizard/migration
- Dashboard/navigation/UI shell

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,3 @@
 # Auto detect text files and perform LF normalization
 * text=auto
+*.sh text eol=lf
--- a/TESTING-GUIDE.md
+++ b/TESTING-GUIDE.md
@@ -1,0 +1,92 @@
+# Hermes Desktop - Testing Guide (Easy 1-2-3)
+
+> For testers who just want to run the app without any technical setup
+
+---
+
+## Option 1: Download Pre-built Release (Easiest)
+
+### Step 1: Download the Right File for Your Computer
+
+| Your Computer | Download This File | Install Method |
+|--------------|-------------------|----------------|
+| **Mac (Apple Silicon - M1/M2/M3)** | `hermes-desktop-X.X.X-arm64-mac.dmg` | Open DMG, drag to Applications |
+| **Mac (Intel)** | `hermes-desktop-X.X.X-x64-mac.dmg` | Open DMG, drag to Applications |
+| **Windows** | `hermes-desktop-X.X.X-setup.exe` | Double-click, click "Run anyway" if Windows warns |
+| **Linux (Ubuntu/Debian)** | `hermes-desktop-X.X.X.deb` | Double-click to install |
+| **Linux (Fedora/RHEL)** | `hermes-desktop-X.X.X.rpm` | `sudo dnf install ./hermes-desktop-X.X.X.rpm` |
+| **Linux (Any - Portable)** | `hermes-desktop-X.X.X.AppImage` | Double-click to run (no install) |
+
+**Where to download:** https://github.com/fathah/hermes-desktop/releases
+
+### Step 2: Install & Open
+
+- **Mac**: Open the `.dmg`, drag Hermes Agent to Applications, open from Applications folder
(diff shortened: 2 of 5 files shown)


pmos69 (changes requested):
Thanks for putting this together. The goal is useful, but I think this needs another pass before merging because some of the tester-facing paths currently fail or can mislead testers.

Requested changes:

1. The new shell scripts are committed with CRLF endings, which breaks bash parsing/execution.
   I extracted the PR blobs and all three `.sh` files are CRLF-only. `bash -n scripts/build-for-tester.sh` fails at `case "$PLATFORM" in` with `$'in\r'`, and `bash -n scripts/quick-test-build.sh` fails the same way at `case "$OS" in`. Since these files are executable, direct execution can also hit the usual `/bin/bash^M` failure. Please normalize the scripts to LF, and ideally add a repo rule such as `*.sh text eol=lf` so this does not regress.

2. Some documented artifact names do not match the current electron-builder config or published releases.
   The guide lists macOS DMGs as `hermes-desktop-X.X.X-arm64-mac.dmg` and `hermes-desktop-X.X.X-x64-mac.dmg`, but `electron-builder.yml` configures DMGs as `${name}-${version}-${arch}.dmg`, and the current release assets are `hermes-desktop-0.5.1-arm64.dmg` / `hermes-desktop-0.5.1-x64.dmg`. The `.deb` name is also not `hermes-desktop-X.X.X.deb`; current assets use the Debian-style `hermes-desktop_0.5.1_amd64.deb`. Please align the table with the real output names.

3. The tester package references a guide filename that is not actually included.
   `package-for-tester.sh` copies `TESTING-GUIDE.md` to `README.txt`, but the generated `INSTALL.txt` says to read `TESTING-GUIDE.md`. Either include the file under that name or update the generated instructions to point to `README.txt`.

4. The packaging script can silently create an incomplete or stale handoff zip.
   Artifact copies use broad globs plus `|| true`, and `builds-for-testing/` is reused across runs. That means a failed build can still produce a zip with no installer, or a zip can accidentally include artifacts from a previous platform/build. For tester distribution this should fail loudly when no fresh artifact is found, and ideally stage from a clean per-run directory.

5. The security/provenance guidance needs tightening for tester builds.
   The guide currently normalizes bypassing Windows SmartScreen and macOS quarantine, and the script suggests sending zips via email/Dropbox/WeTransfer, but there is no checksum/manifest or trust guidance. For unofficial tester builds, please generate SHA256 sums and tell testers to verify hashes from a trusted source before bypassing OS warnings. Also, the `~/.hermes` troubleshooting advice should be changed from deletion to backup/rename first, since that directory can contain config/state/sessions and possibly sensitive local data.

6. The Windows/bash assumption should be documented or avoided.
   The scripts assume bash, `uname`, `cp`, `zip`, etc. That is fine for Git Bash/WSL/macOS/Linux, but this PR is aimed at simple tester handoff and includes Windows packaging. Please either document the required shell/tooling clearly or expose the flow through npm/PowerShell-friendly commands.

7. Minor doc mismatch: the guide says to create a profile in `Settings -> Agents`, but the current app has a top-level `Profiles` section.

I would be happy with this after the scripts are LF-normalized, the package/filename mismatches are corrected, the packaging script fails safely instead of silently, and the tester-build provenance guidance is made explicit.

PR #438, opened 2026-05-28
feat: Add testing infrastructure and developer tooling

0r0b0r011:
## Summary

This PR adds comprehensive testing infrastructure and developer tooling to make it easier to build and distribute Hermes Desktop for testing.

### New Files Added

**Testing Documentation:**
- `TESTING-GUIDE.md` - Non-technical guide for testers (Easy 1-2-3 setup)

**Build Scripts:**
- `scripts/quick-test-build.sh` - One-command build for current platform
- `scripts/build-for-tester.sh` - Build for specific platform (mac/win/linux)
- `scripts/package-for-tester.sh` - Creates ready-to-send zip package with installer + guide

### How to Use

```bash
# Quick build for current platform
./scripts/quick-test-build.sh

# Build for specific platform
./scripts/build-for-tester.sh mac

# Create complete tester package
./scripts/package-for-tester.sh
```

### What Testers Get

A simple zip file containing:
- Platform-specific installer (.dmg/.exe/.AppImage)
- README.txt with quick install instructions
- Detailed TESTING-GUIDE.md

No terminal, no npm, no git required - just download and double-click!

### CI Updates

- Enhanced CI workflow with native module testing across platforms

---

**Note:** This PR includes commits that add the testing infrastructure on top of the existing feature work.

Made with [Cursor](https://cursor.com)

PR #445, opened 2026-05-29
feat(terminal): add embedded terminal with node-pty and xterm

0r0b0r011:
## Summary

Adds an embedded terminal screen to Hermes Desktop:

- **Main process:** `node-pty` session management (`terminal-create`, `terminal-write`, `terminal-resize`, `terminal-kill`) with PTY output pushed to renderer via `terminal-data` events
- **Preload:** Narrow terminal API surface (`terminalCreate`, `terminalWrite`, `terminalResize`, `terminalKill`, `onTerminalData`)
- **Renderer:** xterm.js UI with fit-on-resize, lazy-loaded on first visit
- **Navigation:** Sidebar entry + `Ctrl/Cmd+T` shortcut; hidden behind `RemoteNotice` in remote-only mode
- **CI:** `native-modules` matrix job rebuilds native deps (`node-pty`) on Ubuntu/macOS/Windows

Part of the split from #438. Follows #444 (tester build tooling).

- **Shell spawning:** Main process spawns the user's default shell (`$SHELL` / `/bin/bash`, or `powershell.exe` on Windows) via `node-pty`. Sessions inherit `process.env` from the Electron main process.
- **Working directory:** Optional `cwd` on create; defaults to `homedir()`. Renderer currently calls `terminalCreate()` with no `cwd`.
- **IPC boundary:** Renderer can write arbitrary bytes to an existing session ID and create new sessions. There is no command allowlist — this is a full interactive shell, not a restricted command runner.
- **Process lifecycle:** `killAllTerminals()` runs on `before-quit`; individual sessions are removed on PTY exit. Renderer cleanup calls `terminalKill` on unmount.
- **Broadcast:** PTY output is sent to **all** `BrowserWindow` instances (fine for single-window app today).

## Test plan

- [x] `npm run typecheck`
- [x] `npm test -- tests/ipc-handlers.test.ts tests/preload-api-surface.test.ts`
- [ ] `npm run build`
- [ ] Manual: open Terminal from sidebar, run `echo hello`, resize pane, `Ctrl/Cmd+T` shortcut
- [ ] Manual: confirm terminal unavailable in remote-only mode
- [ ] Manual: quit app with active terminal — no orphaned shell processes

## Follow-up PRs (planned)

- Vault/credentials
- Files/workspace browser
- Profile wizard/migration
- Dashboard/navigation/UI shell


pmos69 (changes requested):
Thanks for splitting this out and for calling out the shell-spawning model in the PR body. I think the feature direction is useful, but this needs a security/robustness pass before it should merge because it exposes a full local interactive shell through preload IPC.

Requested changes:

1. Enforce the remote-only boundary in the main-process IPC handler, not just in the renderer.
   The Terminal screen is hidden behind `RemoteNotice` in `Layout.tsx`, but `terminal-create` itself still spawns a local shell from `src/main/terminal.ts` whenever renderer code calls `window.hermesAPI.terminalCreate()`. Other local-only surfaces, like Kanban, guard in the main-process implementation with `isRemoteOnlyMode()` and return an unsupported-mode response. Terminal should do the same. This is especially important because this IPC creates a real local `powershell.exe` / `/bin/bash` process.

2. Bind terminal sessions to the creating `webContents` instead of broadcasting/accepting by id globally.
   `terminal-data` is currently sent to every `BrowserWindow` via `BrowserWindow.getAllWindows()`, and `terminal-write` / `terminal-resize` / `terminal-kill` only require a session id. Please store the owner `event.sender.id` when creating the PTY, send output only to that sender, and reject write/resize/kill calls from non-owner senders. Broadcasting is probably harmless in the current single-window app, but it is a weak contract for a shell feature and will become a footgun as soon as additional windows/webContents exist.

3. Add input validation and limits around the terminal IPC surface.
   Right now `cwd`, `id`, `data`, `cols`, and `rows` are passed through directly. Some freedom is expected for a full terminal, but the IPC boundary should still reject obvious bad inputs: invalid/unknown session ids, non-string writes, very large write chunks, non-finite or unreasonable resize dimensions, invalid `cwd`, and unbounded session creation. Please also return structured errors where practical so the renderer can show a useful failure state.

4. Handle spawn, resize, and renderer initialization failures cleanly.
   `pty.spawn()` can fail if the shell is unavailable or `cwd` is invalid, and `pty.resize()` can throw for bad dimensions or closed PTYs. The renderer also awaits `terminalCreate()` without a `try/catch`, so a failure can leave the Terminal screen blank or only log an unhandled promise rejection. Please wrap the main-process operations and add renderer error UI/cleanup for failed terminal startup.

5. Use the maintained xterm packages.
   This PR adds `xterm` and `xterm-addon-fit`, both of which are deprecated in the lockfile in favor of `@xterm/xterm` and `@xterm/addon-fit`. Since this is new code, please use the maintained package names from the start.

6. Add behavior/security tests, not just API-surface tests.
   The added tests verify that the preload and main channel names line up, which is useful, but they do not exercise the risky parts of this feature. Please add targeted tests for remote-only rejection, owner-only write/resize/kill, resize/input validation, create/session limits, spawn failure handling, and cleanup on kill/exit where feasible.

7. Complete locale coverage for the new navigation label.
   `navigation.terminal` is added only for English. Missing locales falling back to English is not merge-blocking by itself, but it is incomplete and easy to fill in with “Terminal” or the appropriate localized label in the existing locale files.

I did not run `npm ci` or a full build locally because this PR adds `node-pty`, which has native install/build scripts. Static review was enough to confirm the issues above. Once the IPC boundary and dependency choices are tightened up, this will be much easier to validate safely.

PR #447, opened 2026-05-29
feat(files): add workspace file browser with path sandbox

0r0b0r011:
## Summary

Adds a sandboxed file browser for local workspace editing:

- **Main process:** `files-list-dir`, `files-read`, `files-write` with path allowlist (`homedir()` + `process.cwd()`)
- **Preload:** `filesListDir`, `filesRead`, `filesWrite`
- **Renderer:** Split-pane Files screen (tree + editor)
- **Tests:** `tests/files-path.test.ts` validates path traversal boundaries

Part of the split from #438. Follows #444–#446.

- **Path traversal:** All paths resolved and checked against `ALLOWED_ROOTS` before read/write
- **Workspace boundary:** Only `homedir()` and `process.cwd()` roots — `/etc`, `/tmp` outside roots are rejected
- **Write model:** Renderer can write any allowed path via IPC — no file-type or size limits
- **Remote mode:** Files screen gated behind `RemoteNotice`

## Test plan

- [x] `npm run typecheck`
- [x] `npm test -- tests/files-path.test.ts tests/ipc-handlers.test.ts tests/preload-api-surface.test.ts`
- [ ] `npm run build`
- [ ] Manual: browse home dir, open/edit/save a text file
- [ ] Manual: confirm `/etc/passwd` or paths outside roots are rejected


pmos69 (changes requested):
Thanks for splitting this out. I think the Files screen is a useful direction, but this PR should not merge in its current shape. The security boundary is the core feature here, and the current boundary is both too broad and bypassable.

Requested changes / findings:

1. The path sandbox is bypassable through symlinks.
   `isPathAllowed()` uses `resolve()`/`normalize()`, which only performs lexical path resolution. It does not resolve symlinks. The actual filesystem operations (`readdirSync`, `statSync`, `readFileSync`, `safeWriteFile`) do follow symlinks, so a symlink inside an allowed root can point outside the allowed root and still pass the check. For example, a link under the user home that points to `/etc` or another sensitive location would lexically look like it is under home, but reads/writes would follow the link outside the root. Please use real path checks: `realpathSync()` for existing list/read targets, and `realpathSync()` on the parent directory for writes where the target may not exist yet.

2. `homedir()` is not an acceptable default writable sandbox.
   The allowlist is `homedir()` plus `process.cwd()`. The user home contains SSH keys, cloud credentials, browser/app config, shell startup files, Hermes config/state, and potentially the vault DB/key files from #446. Exposing read/write access to the whole home directory through renderer IPC is not a workspace sandbox; it is broad local filesystem authority. Please move to an explicit user-selected workspace root, and consider denying sensitive subtrees even inside that root.

3. `process.cwd()` is a non-deterministic and unsafe root.
   In packaged Electron apps the current working directory depends on launch context. It can be the install directory, a system directory, `/`, or something else. It should not silently widen the filesystem allowlist. Please drop it or replace it with an explicit persisted workspace directory chosen by the user.

4. The main-process handlers are not guarded for remote-only mode.
   The UI hides Files behind `RemoteNotice`, but the IPC handlers are still registered and callable. In remote/SSH modes, a user would reasonably expect a file browser to target the remote environment or be unavailable, not silently expose local home files. Please enforce the mode check in the main process, not only in `Layout.tsx`.

5. The IPC surface has no size/type limits.
   `files-read` uses `readFileSync(path, "utf-8")`; a large or binary file can freeze or OOM the Electron main process. `files-write` is also unbounded. Please add file size limits, text/binary detection or explicit text-only constraints, and bounded write sizes.

6. Tests copy-paste the path helper instead of testing production code.
   `tests/files-path.test.ts` duplicates `isPathUnderRoot()` / `isPathAllowed()` instead of importing the implementation. That means the real handler logic can drift from the tests. Please export/test the production helper or test through the registered handlers, and add coverage for symlink escape, remote-mode rejection, size limits, binary/unreadable files, and write behavior.

7. The Files UI has no error handling for IPC failures.
   Directory listing, read, and write calls are all used without a catch/error state. Permission errors, denied paths, binary decode issues, failed writes, and broken directory entries will likely become unhandled rejections and stale UI. Please surface errors in the Files screen and keep the editor state consistent.

8. Directory listing can fail because of one bad entry.
   `files-list-dir` maps over entries and calls `statSync` on each one. A broken symlink, permission-denied child, or transient file can throw and fail the whole directory. Please handle per-entry failures or use `lstat`/safe stat behavior deliberately.

9. Initial directory detection is not cross-platform.
   `Files.tsx` derives the parent directory with `path.split("/")`, which is wrong for Windows backslash paths. The first load can set `cwd` incorrectly, including to a file path, causing the next `filesListDir(cwd)` to fail. The main process should probably return the current/root path explicitly instead of reconstructing it in the renderer.

10. The CSS redefines global app classes.
    The PR adds new global `.screen-header` and `.btn-primary` rules near the end of `main.css`, both of which already exist earlier. Please scope Files-specific styles or reuse existing app classes to avoid unrelated UI regressions.

11. Locale coverage is incomplete.
    `navigation.files` is added only for English. This is not a security blocker, but it is incomplete.

Credit where due: the lexical `..` traversal and sibling-prefix cases are handled correctly. The missing piece is that filesystem containment has to be based on real paths, not only lexical paths, and the chosen root cannot be the entire home directory.

There is also an important cross-PR concern with #446: if the vault PR and this file-browser PR both land, renderer code could use this Files IPC to read/write `.hermes` vault files and key material from the local home directory, undermining the vault threat model. These PRs should be evaluated together from a security-boundary perspective.

Minimum bar before reconsidering: explicit user-selected workspace roots, no `homedir()`/`process.cwd()` broad allowlist, realpath-based containment including write-parent checks, main-process remote-mode enforcement, size/type limits, robust UI errors, and tests that exercise the real production code including symlink escape attempts.

PR #449, opened 2026-05-29
feat(ui): add dashboard, MCP hub, swarm, ecosystem, and color themes

0r0b0r011:
## Summary

Adds the remaining navigation/UI shell screens from the #438 split:

- **Dashboard:** Live stats (sessions, model, gateway, cron jobs) via `dashboard-stats` IPC
- **MCP Hub:** Installed servers + static catalog via `mcp-catalog` IPC
- **Swarm:** Multi-agent worker grid prototype (UI-only conductor toggle)
- **Ecosystem:** Skills count, memory provider links, theme previews
- **ThemeProvider:** Accent color themes (Hermes/Nous/Bronze/Slate/Mono) with Settings picker

**Excluded** (deferred per split plan): `Chat/ToolCallCard.tsx`, `Chat/WorkspacePanel.tsx`, and other chat workspace changes.

Independent PR off `main`. Part of the split from #438.

## Test plan

- [x] `npm run typecheck`
- [x] `npm test -- tests/ipc-handlers.test.ts tests/preload-api-surface.test.ts`
- [ ] `npm run build`
- [ ] Manual: open Dashboard, verify stats refresh
- [ ] Manual: MCP Hub shows installed servers and catalog
- [ ] Manual: change color theme in Settings, confirm accent updates
- [ ] Manual: Swarm/Ecosystem screens render without errors

## Merge order (full split)

1. #444 — Tester build tooling
2. #445 — Embedded terminal
3. #446 — Vault/credentials
4. #447 — Files browser
5. #448 — Profile wizard (depends on #446)
6. **This PR** — Dashboard/UI shell

Made with [Cursor](https://cursor.com)

PR #450, opened 2026-05-29
feat(chat): add workspace panel, tool call cards, and screen polish

0r0b0r011:
## Summary

Final deferred slice from the #438 split — chat workspace UI and minor screen polish:

- **Chat workspace panel:** Side panel showing context folder path and latest tool output during chat
- **Tool call cards:** Collapsible JSON syntax-highlighted cards replacing inline tool-call history rows
- **Memory:** Split-pane edit with live markdown preview
- **Skills:** Origin badges (marketplace/project/built-in) and path display on skill cards
- **Gateway:** "Test connection" button per enabled platform

No IPC, preload, or dependency changes — renderer-only.

## Context

Supersedes the remaining hunks from closed #438. Merge after the feature PRs (#444–#449) or independently — this PR has no dependencies on vault, terminal, or profile wizard.

After all split PRs merge, `main`'s oversized commits (`817bb31`, `edc4cdf`) can be dropped in favor of the merged branches.

## Test plan

- [x] `npm run typecheck`
- [x] `npm test -- tests/sessions-history-items.test.ts src/renderer/src/screens/Skills/Skills.test.tsx`
- [ ] `npm run build`
- [ ] Manual: send chat with tool calls — verify workspace panel and tool call cards
- [ ] Manual: edit memory entry — verify markdown preview pane
- [ ] Manual: Skills tab — verify origin badges and paths
- [ ] Manual: Gateway — enable platform, click "Test connection"

## Full split stack

1. #444 — Tester build tooling
2. #445 — Embedded terminal
3. #446 — Vault/credentials
4. #447 — Files browser
5. #448 — Profile wizard (depends on #446)
6. #449 — Dashboard/UI shell
7. **This PR** — Chat workspace + screen polish

### DIFF
--- a/src/renderer/src/assets/main.css
+++ b/src/renderer/src/assets/main.css
@@ -5959,3 +5959,25 @@ body {
   font-size: 11px;
   color: var(--text-muted);
 }
+
+/* ── Chat workspace panel & tool call cards ────────────────────────────── */
+
+.chat-layout { display: flex; flex: 1; overflow: hidden; }
+.chat-layout .chat-container { flex: 1; min-width: 0; }
+.workspace-panel { width: 280px; border-left: 1px solid var(--border); padding: 12px; overflow-y: auto; background: var(--bg-secondary); }
+.workspace-panel-title { font-size: 13px; font-weight: 600; margin-bottom: 12px; }
+.workspace-panel-section { margin-bottom: 12px; }
+.workspace-panel-label { font-size: 11px; color: var(--text-muted); margin-bottom: 4px; }
+.workspace-panel-path { font-size: 11px; word-break: break-all; }
+.workspace-panel-output { font-size: 11px; white-space: pre-wrap; max-height: 200px; overflow-y: auto; }
+
+.tool-call-card { margin-left: 8px; border: 1px solid var(--border); border-radius: 8px; overflow: hidden; }
+.tool-call-card-header { padding: 8px 12px; cursor: pointer; font-size: 13px; }
+.tool-call-card-body { border-top: 1px solid var(--border); }
+
+.skills-card-path { font-size: 10px; color: var(--text-muted); margin-top: 4px; word-break: break-all; }
+.skill-origin-badge { font-size: 10px; padding: 2px 6px; border-radius: 4px; margin-left: 6px; background: var(--bg-tertiary); color: var(--text-muted); }
+.skill-origin-marketplace { background: rgba(99,102,241,0.2); color: #818cf8; }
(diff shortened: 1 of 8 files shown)


pmos69 (changes requested):
Reviewed statically (didn't execute). This is the cleanest PR of the #438 split — renderer-only, no security issues, and it gets the CSS discipline right. One change request before it lands.

## 🔴 Change request — "Test connection" button doesn't test the connection

`Gateway.tsx`:

```tsx
{platformEnabled[platform.key] && (
  <button ... onClick={async () => {
    const started = await window.hermesAPI.startGateway();
    if (started) await loadConfig();
  }}>Test connection</button>
)}
```

The button is rendered **per enabled platform** and labelled "Test connection", but the handler ignores `platform` entirely and just calls `startGateway()` — the same global action on every row. It never verifies the platform's credentials (e.g. whether the Telegram bot token actually authenticates). So the advertised feature ("Test connection button per enabled platform") isn't implemented — it's an N-times-duplicated "start gateway" button.

Please either:
- wire it to a real per-platform credential/connection check, or
- relabel it ("Start gateway") and render it once rather than per-platform.

## 🟡 Minor — cross-PR CSS coupling

This button uses `.btn-secondary`, which #449 currently restyles (`--bg-elevated` → `--bg-tertiary`, see my review there). Harmless on its own, but if #449's `.btn-secondary` override isn't fixed first, this button inherits that colour change. Flagging so the two are coordinated.

## ✅ What's good (worth calling out)

- **No XSS** despite rendering untrusted content: `ToolCallCard` uses `react-syntax-highlighter` (tokenised React elements, escaped children); `WorkspacePanel` renders tool output via `<pre>{toolOutput}</pre>` (React-escaped); `Memory` preview reuses the existing `AgentMarkdown` (`react-markdown` + `remark-gfm`, no `rehype-raw`, so raw HTML is escaped). No `dangerouslySetInnerHTML` anywhere — reusing the app's existing sanitised markdown renderer rather than rolling a new one is exactly right.
- **CSS is properly namespaced** — every new selector is scoped (`.tool-call-card`, `.workspace-panel-*`, `.memory-preview-*`, `.skill-origin-*`), and it reuses `.btn`/`.btn-secondary`/`.btn-sm` rather than redefining them. This is the model the rest of the split should follow.
- **"No dependency changes" holds** — `react-syntax-highlighter` is already in `package.json`.

Net: fix or relabel the Test-connection button and this is good to go. Everything else (tool-call cards, workspace panel, markdown preview, skill origin badges) is solid.

PR #470, opened 2026-05-30
fix: fall back to configured SSH API port

sysophelper-droid:
Fixes SSH-mode API URL resolution when the in-process tunnel state is unavailable.

In SSH mode, `getApiUrl()` previously threw `SSH tunnel is not active` whenever `getSshTunnelUrl()` returned null. That can happen while the configured local SSH forward is still usable, for example with externally-started tunnels or when an SSH ControlMaster process outlives the child process that created it. The UI then silently fails remote API calls such as `/api/jobs`, making Schedules appear empty.

This change keeps the existing active-tunnel behavior, but falls back to the configured `conn.ssh.localPort` before throwing.

Verification:
- `npm run typecheck`
- `npm run lint` (passes with existing repository warnings only; no warnings in touched files)
- `npm test -- --run tests/remote-mode-url-and-spawn.test.ts`
- `npm test` was also run; it still has pre-existing failures in `tests/hermes-cli-session-id.test.ts` due that test mock missing `getActiveProfileNameSync`, unrelated to this change.

### DIFF
--- a/src/main/cronjobs.ts
+++ b/src/main/cronjobs.ts
@@ -4,7 +4,13 @@ import { join } from "path";
 import { execFile } from "child_process";
 import { HERMES_HOME, HERMES_PYTHON, hermesCliArgs } from "./installer";
 import { profileHome } from "./utils";
-import { isRemoteMode, getApiUrl, getRemoteAuthHeader } from "./hermes";
+import {
+  isRemoteMode,
+  getApiUrl,
+  getRemoteAuthHeader,
+  normaliseRemoteUrl,
+} from "./hermes";
+import { getConnectionConfig } from "./config";
 import { HIDDEN_SUBPROCESS_OPTIONS } from "./process-options";
 
 export interface CronJob {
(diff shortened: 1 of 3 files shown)


pmos69:
I think the fallback is useful for the externally-started tunnel / lost in-process state case, especially for Schedules where `cronjobs.ts` calls `getApiUrl()` directly and does not first run the SSH tunnel setup path that Chat uses.

One concern: `conn.ssh.localPort` is the configured/preferred port, not necessarily the actual active tunnel port. `startSshTunnel()` calls `findFreePort(config.localPort || 18642)` and may choose a different port if the preferred one is occupied; that actual port only lives in `activeConfig`. If `activeConfig` is later unavailable, falling back to `conn.ssh.localPort` can point at a stale port or an unrelated local service.

Also, `getConnectionConfig()` defaults `ssh.localPort` to `18642`, so in production this fallback will almost always return `http://127.0.0.1:18642` in SSH mode when `getSshTunnelUrl()` is null. The new test for "no tunnel state or local port" uses a mocked state that the real config reader does not normally produce.

Could we either:

1. validate the fallback port with `/health` before using it for authenticated API calls, or
2. scope this fallback to the Schedules/Cron path that needs it, rather than changing global `getApiUrl()` behavior?

At minimum, I’d like the test/comment to make clear that this is falling back to the configured/default port and may fail later if that port is not actually the Hermes SSH tunnel.

I ran the targeted test locally in a temp worktree:

```text
npm test -- --run tests/remote-mode-url-and-spawn.test.ts
```

It passes: 1 file, 20 tests.

sysophelper-droid:
Thanks, that concern makes sense — updated.

I changed this so `getApiUrl()` no longer has a global fallback to the configured/default SSH port. It now preserves the original `SSH tunnel is not active` behavior when the in-process tunnel URL is unavailable.

The configured/default port fallback is now scoped to the Schedules/Cron API path only, and it probes `${configuredPort}/health` before sending the actual cron API request. If that health probe fails, cron keeps the original failure path instead of sending authenticated requests to a stale port or unrelated local service.

I also updated the tests to cover both sides:

- global `getApiUrl()` still throws when tunnel state is unavailable
- cron uses the configured/default port only after a successful `/health` probe
- cron does not call `/api/jobs` when the `/health` probe fails

Verification rerun locally:

- `npm test -- --run tests/remote-mode-url-and-spawn.test.ts` passes: 1 file, 22 tests
- `npm run typecheck:node` passes
- `npm run typecheck:web` passes
- `npm run lint` passes with existing warnings only

PR #471, opened 2026-05-30
fix(test): stub missing utils exports in hermes-cli-session-id mock

bionicbutterfly13:
## What

The `vi.mock("../src/main/utils", …)` factory in `tests/hermes-cli-session-id.test.ts` only stubbed `stripAnsi` and `pidIsAliveAs`, but `src/main/hermes.ts` also imports `profileHome` and `getActiveProfileNameSync`. Because a `vi.mock` factory replaces the *entire* module, the un-stubbed exports are `undefined`. When `gatewayPidPaths()` calls `getActiveProfileNameSync()`, Vitest throws *"No 'getActiveProfileNameSync' export is defined on the mock"*, failing all 4 tests in the file.

## Why it matters

This is currently breaking **CI on `main`** — the latest run (Node 22) fails with this exact error. This file is the only one that exercises the gateway-PID path, which is why the sibling `utils` mocks (which stub only `stripAnsi`) don't hit it.

## Change

Add the two missing exports to the mock, mirroring the real module's behavior (`getActiveProfileNameSync` → `"default"`; `profileHome` builds a `profiles/<name>` path for named profiles, so a future named-profile test resolves distinct PID paths). Test-only — no production code touched.

## Verification

- `npm run typecheck` ✅
- `npm test` ✅ — 60 files, 690 passed / 3 skipped (previously 4 failed in this file)

### DIFF
--- a/tests/hermes-cli-session-id.test.ts
+++ b/tests/hermes-cli-session-id.test.ts
@@ -152,6 +152,11 @@ vi.mock("../src/main/ssh-tunnel", () => ({
 vi.mock("../src/main/utils", () => ({
   stripAnsi: (s: string) => s,
   pidIsAliveAs: () => false,
+  getActiveProfileNameSync: () => "default",
+  profileHome: (profile?: unknown) =>
+    profile && profile !== "default"
+      ? `${TEST_HOME}/profiles/${profile}`
+      : TEST_HOME,
 }));
 
 vi.mock("../src/main/models", () => ({


pmos69:
Thanks, this fixes the missing `utils` exports issue, and that part is correct.

I tested the PR branch locally on Windows with:

```text
npm test -- --run tests/hermes-cli-session-id.test.ts
```

The file still fails, but now for the next cross-platform mock issue:

```text
[gateway] Cannot start: Python interpreter not found at /usr/bin/python3
```

The test mock still uses Unix-only paths:

```ts
HERMES_PYTHON: "/usr/bin/python3",
HERMES_REPO: "/dev/null",
```

Current `startGateway()` preflights both paths with `existsSync()`, so on Windows the mocked gateway start returns `false`.

I tested the minimal follow-up locally:

```ts
HERMES_PYTHON: process.execPath,
HERMES_REPO: process.cwd(),
```

With that change, the targeted file passes: 4 tests passed.

Could you include that cross-platform mock adjustment in this PR as well? Then this should be good to merge.

pmos69:
Thanks again for the focused test fix. This has since been superseded by current `main`.

The missing `utils` mock exports are now present in `tests/hermes-cli-session-id.test.ts`, and the follow-up cross-platform mock adjustment is there too: the test now uses `process.execPath` for `HERMES_PYTHON` and a real temp repo path instead of Unix-only `/usr/bin/python3` / `/dev/null`.

This branch now conflicts in the same test file and would reintroduce the older mock shape, so I think we can close it as superseded.

Issue #466, opened 2026-05-30
gateway just not started after upgrade to 0.52.0


Slach:
<img width="694" height="701" alt="Image" src="https://github.com/user-attachments/assets/c1cb055c-f9fa-4dc7-b640-92a14495c2cf" />

manual `hermes gateway run` works

but websocket on 18789 port didn't start

saviliana:
Also office doesn't work as when clicked start, it said:

"
The argument 'stdio' is invalid., Received WriteStream {fdL null, path:'C:\\User\\User\\AppData\\Local\\hermes\\gateway-stderr.log', flags'a', mode:438 fl...
"
The error code wasn't fully generated, log is empty.
In Hermes folder that file is empty, deleted and it would come back if office start button clicked again.
Probably the two is related as office need websocket to run.

poopsoups:
Same here. Rolling back to v0.5.1 has got me working again.

Luxciax:
I am seeing what looks like the same issue on Windows with Hermes Desktop 0.5.2.

Desktop's Gateway button cannot reliably start the gateway/API server, but manually starting the gateway can work.

Environment:

- OS: Windows 11
- Hermes Desktop: 0.5.2
- Install path: `C:\Users\V\AppData\Local\Programs\hermes-desktop`
- Hermes home: `C:\Users\V\AppData\Local\hermes`
- Gateway API server configured in `config.yaml`:

```yaml
platforms:
  api_server:
    enabled: true
    extra:
      port: 8642
      host: "127.0.0.1"
```

Observed behavior:

- Clicking the Desktop Gateway control says/behaves like the gateway cannot be started.
- Another launcher/tool can start the gateway, and manual CLI startup also works.
- `gateway_state.json` can remain stale and report `gateway_state: running` with `api_server: connected`, while in reality:
  - no `hermes gateway run` process is alive
  - `127.0.0.1:8642` is not listening
- This makes Desktop appear out of sync with the real gateway process state.

One concrete failure cause I found:

```text
ERROR gateway.platforms.api_server: [Api_Server] Refusing to start:
API_SERVER_KEY is required for the API server, including loopback-only binds on 127.0.0.1.
```

After adding these to `%LOCALAPPDATA%\hermes\.env`:

```env
API_SERVER_ENABLED=true
API_SERVER_HOST=127.0.0.1
API_SERVER_PORT=8642
API_SERVER_KEY=<random local key>
```

and then manually running:

```powershell
[...]

PR #484, opened 2026-05-31
Fix Windows gateway start stderr fd

Alfianfc:
## Summary

Fix the Messaging Gateway Start action failing on Windows/Electron when stderr is routed to a log file.

`startGateway()` was passing an `fs.createWriteStream(...)` object directly to `child_process.spawn({ stdio })`. In Electron/Node on Windows, the stream's `.fd` can still be `null` at spawn time, so `spawn()` rejects the stdio entry and the desktop Start button cannot launch the gateway. Starting the gateway manually from a terminal still works, which matches the report in #466.

This PR opens `gateway-stderr.log` synchronously with `openSync()` and passes the numeric file descriptor to `spawn()`, then closes the parent fd on process error/close.

## Tests

- `npm test -- tests/hermes-cli-session-id.test.ts`
- `npm run typecheck`
- `npm run build`
- Targeted ESLint error check: `npx eslint src/main/hermes.ts tests/hermes-cli-session-id.test.ts -f json` → `eslint_errors=0`

Note: full `npm run lint` currently emits many pre-existing Prettier CRLF warnings in the checkout; the targeted files have no ESLint errors.

Fixes the Windows stderr fd gateway-start failure reported in #466

### DIFF
--- a/src/main/hermes.ts
+++ b/src/main/hermes.ts
@@ -6,7 +6,8 @@ import {
   appendFileSync,
   unlinkSync,
   mkdirSync,
-  createWriteStream,
+  openSync,
+  closeSync,
 } from "fs";
 import { join } from "path";
 import { homedir } from "os";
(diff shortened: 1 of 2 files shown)


pmos69 (changes requested):
Thanks for jumping on this. The code change looks like the right fix for the specific Windows/Electron regression: v0.5.1 spawned the gateway with `stdio: "ignore"`, while v0.5.2 introduced `createWriteStream(logPath)` as the stderr stdio entry in commit `869056a` / PR #463. On Windows/Electron that stream can still have `fd: null` when `spawn()` validates stdio, which matches the exact `The argument 'stdio' is invalid. Received WriteStream { fd: null, ... }` report. Opening the log file synchronously and passing the numeric fd addresses that root cause.

Please adjust the PR wording before merge: #466 now contains a few related-but-distinct gateway startup reports, including `API_SERVER_KEY` configuration and stale gateway-state detection. This PR fixes the stderr fd startup failure introduced between v0.5.1 and v0.5.2, but it does not cover all of the later #466 discussion. I would avoid `Closes #466` unless the maintainer wants to scope that issue only to the invalid-stdio regression. Safer wording would be something like `Fixes the Windows stderr fd gateway-start failure reported in #466`.

No blocking concern with the implementation itself from static review.

pmos69:
Thanks again for the focused Windows fix. This was the right direction for the original stderr fd failure: passing an already-open numeric fd avoids the Electron/Windows race where a `WriteStream` can still have `fd: null` when `spawn()` validates `stdio`.

Current `main` has since absorbed that production fix in the refactored gateway startup path. `startGatewayDetailed()` now opens `gateway-stderr.log` with `openSync(...)`, passes the numeric fd to `spawn(... stdio ...)`, falls back to `"ignore"` if opening the log fails, and closes the parent fd after spawn.

Because the gateway code has moved a lot since this branch, this PR now conflicts in `src/main/hermes.ts` and `tests/hermes-cli-session-id.test.ts`, and merging it would reapply an older gateway-start shape over the current per-profile implementation.

I'm closing this as superseded by current `main`. The only remaining useful bit might be the explicit regression assertion that gateway spawn receives a numeric stderr fd; if we want that extra coverage, it should be opened as a tiny fresh test-only PR against current `main`.

PR #487, opened 2026-05-31
fix(memory): read character limits from profile config

### LINKED ISSUE #483: Desktop Memory page still hardcodes 2200 / 1375 limits instead of reading config.yaml memory_char_limit / user_char_limit
Body Hermes Desktop appears to read the correct memory files, but the Memory page still uses hardcoded character limits instead of the values configured in ~/.hermes/config.yaml.

Environment:

Hermes Desktop installed app version: 0.4.5
Latest upstream release checked: v0.5.2
macOS
Hermes config:
memory.memory_char_limit: 3200
memory.user_char_limit: 2000
Actual behavior:

Desktop shows:
Agent memory: 2470 / 2200 chars (100%)
User profile: 1873 / 1375 chars (100%)
But the real configured limits are 3200 / 2000
Actual files are:
~/.hermes/memories/MEMORY.md = 2681 chars
~/.hermes/memories/USER.md = 1873 chars
Expected behavior:

The Memory page should read the effective limits from the active profile’s config.yaml
It should display 3200 / 2000 in my case, not 2200 / 1375
Evidence:

Current Hermes Desktop source on main still hardcodes these values:
src/main/memory.ts const MEMORY_CHAR_LIMIT = 2200; const USER_CHAR_LIMIT = 1375;
src/main/ssh-remote.ts const MEMORY_CHAR_LIMIT = 2200; const USER_CHAR_LIMIT = 1375;
The returned memory info also uses these constants as charLimit, so the UI cannot reflect configured values from Hermes
Notes:

This makes the Desktop Memory page inconsistent with actual Hermes runtime behavior
It can incorrectly show memory/profile as “100% full” even when the configured limits were intentionally raised
There may also be a refresh/cache issue for Agent memory count, because Desktop showed 2470 while the current MEMORY.md length is 2681
Suggested fix:

Read memory.memory_char_limit and memory.user_char_limit from the active profile config.yaml
Fall back to 2200 / 1375 only if those config keys are absent

jonasboos:
## Summary
- read `memory.memory_char_limit` and `memory.user_char_limit` from the active profile's `config.yaml`, with the existing 2200 / 1375 defaults as fallback
- use the resolved limits for Memory page payloads and for local add/update/user-profile write validation
- apply the same config-based limits in SSH mode so remote profiles are consistent with local profiles

Fixes #483

## Tests
- `npm test` (clean worktree: 64 files, 716 passed, 3 skipped)
- `npm run typecheck`
- `npx eslint src/main/memory-limits.ts src/main/memory.ts src/main/ssh-remote.ts tests/memory-limits.test.ts`

### DIFF
--- a/src/main/memory-limits.ts
+++ b/src/main/memory-limits.ts
@@ -1,0 +1,27 @@
+import { getYamlPath } from "./yaml-path";
+
+export const DEFAULT_MEMORY_CHAR_LIMIT = 2200;
+export const DEFAULT_USER_CHAR_LIMIT = 1375;
+
+export interface MemoryLimits {
+  memoryCharLimit: number;
+  userCharLimit: number;
+}
+
+function parsePositiveInteger(raw: string | null): number | null {
+  if (!raw) return null;
+  const value = Number(raw.trim());
+  if (!Number.isSafeInteger(value) || value <= 0) return null;
+  return value;
+}
+
+export function parseMemoryLimitsConfig(content: string): MemoryLimits {
+  return {
+    memoryCharLimit:
+      parsePositiveInteger(getYamlPath(content, "memory.memory_char_limit")) ??
+      DEFAULT_MEMORY_CHAR_LIMIT,
+    userCharLimit:
+      parsePositiveInteger(getYamlPath(content, "memory.user_char_limit")) ??
+      DEFAULT_USER_CHAR_LIMIT,
+  };
+}
--- a/src/main/memory.ts
+++ b/src/main/memory.ts
@@ -2,10 +2,9 @@ import { existsSync, readFileSync, statSync } from "fs";
 import { join } from "path";
 import Database from "better-sqlite3";
 import { profileHome, safeWriteFile } from "./utils";
+import { parseMemoryLimitsConfig, type MemoryLimits } from "./memory-limits";
 
 const ENTRY_DELIMITER = "\n§\n";
-const MEMORY_CHAR_LIMIT = 2200;
-const USER_CHAR_LIMIT = 1375;
 
 export interface MemoryEntry {
   index: number;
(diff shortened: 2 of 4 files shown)


pmos69:
Thanks for the focused fix. I checked this locally and the main path looks good to me:

- `npm run typecheck` passes
- `npm test -- --run tests/memory-limits.test.ts` passes
- `npm test -- --run tests/ssh-remote.test.ts tests/memory-limits.test.ts` passes
- `git diff --check main...HEAD` is clean

One small request before merge: could you add an SSH-specific regression test for the new remote behavior?

The PR changes `src/main/ssh-remote.ts` so SSH mode reads `config.yaml` limits too, but the new tests currently cover the parser/local memory path. A focused test that mocks the SSH file reads and verifies one of these would close the loop:

- `sshReadMemory()` returns `memory.charLimit` / `user.charLimit` from the remote profile config
- `sshAddMemoryEntry()` rejects/accepts based on the remote `memory.memory_char_limit`
- `sshWriteUserProfile()` rejects/accepts based on the remote `memory.user_char_limit`
- missing/invalid remote config still falls back to `2200 / 1375`

This does not look like a behavior blocker from my read; it is mainly to make sure the SSH parity claimed by the PR stays covered going forward.

pmos69:
Rechecked this against current `main` and it still looks good to merge.

I verified on a scratch merge with current `origin/main`:

- merge into current `main`: clean
- `git diff --check`: clean
- `npm test -- --run tests/memory-limits.test.ts`: pass, 7 tests
- `npm test -- --run tests/ssh-remote.test.ts tests/memory-limits.test.ts`: pass, 24 tests
- `npm run typecheck:node`: pass

The earlier SSH-specific coverage request is still a useful follow-up, but I don't think it needs to block this PR. The current SSH test suite still passes with the change, and the implementation is small/focused. We can merge this and track the direct SSH memory-limit regression test separately.

PR #496, opened 2026-06-01
feat(sessions): add rename session feature

sysophelper-droid:
Add a pencil icon to the left of the trash icon on each session card that turns the title into an inline editable input (Enter saves, Esc cancels, with check/cancel buttons). Works on both the grouped session list and search-result cards.

The rename writes the database `title` column (not just the JSON cache) so it survives a cache re-sync and is reflected in SSH mode, which reads titles straight from the remote DB:

- renameSession() in sessions.ts updates the DB and syncs the fast-path cache, mirroring deleteSession.
- sshRenameSession() in ssh-remote.ts runs a parameterized UPDATE over SSH, mirroring sshSearchSessions.
- The existing (previously unused) update-session-title IPC channel and updateSessionTitle preload bridge are reused; the handler is now SSH-aware and persists to the DB.

Adds rename strings to all nine locales and full test coverage: local DB rename, SSH remote rename, and renderer interaction (list and search cards).


pmos69:
This looks superseded by #599, which has already been merged.

Current `main` already includes the session rename UI, the `update-session-title` IPC/preload path, local DB/cache persistence, and remote/SSH-dashboard persistence through the current remote sessions bridge. This branch is also now stale/conflicting and follows the older SSH/session implementation shape.

I think we can close this PR as superseded. If there is any extra test coverage here that we still want, we can extract that separately against current `main`.

PR #499, opened 2026-06-01
Fix gateway restart after resume

### LINKED ISSUE #492: 【Bug Report】 hermes update` breaks desktop app gateway state
# Bug Report: `hermes update` breaks desktop app gateway state

## Summary

After running `hermes update`, the Hermes desktop app (Electron) becomes unable to send messages — the UI stays stuck on "sending" indefinitely. The CLI version (`hermes chat`) continues to work fine because it does not depend on the gateway service.

---

## Environment

- **OS:** Windows 11
- **Hermes version:** 0.15.1
- **Model/provider:** stepfun-ai/step-3.7-flash via NVIDIA NIM (https://integrate.api.nvidia.com/v1)
- **Desktop app path:** `C:\Users\Administrator\AppData\Local\hermes\app\hermes-agent.exe`
- **Hermes home:** `C:\Users\Administrator\AppData\Local\hermes\`

## Steps to Reproduce

1. Open Hermes desktop app. Confirm it is working (send a message, receive a response).
2. Run `hermes update` from the CLI.
3. Restart the desktop app (or leave it running).
4. Send a new message from the desktop app UI.

## Expected Behavior

The desktop app should send the message and display the model's response normally, just like before the update.

## Actual Behavior

- The message stays in "sending" state indefinitely.
- No response is displayed.
- No error dialog is shown to the user.
- Closing and reopening the desktop app does not fix it.
- The CLI (`hermes chat`) continues to work correctly.

---

## Root Cause (as observed)

The `hermes update` command triggers a bootstrap process that rewrites/reinstalls the Hermes backend. After the update:

1. The gateway service (`hermes gateway`) ends up in a **stopped / crashed** state.
   - Evidence: `gateway-exit-diag.log` records `gateway.exit_nonzero pid=2520`.
   - Evidence: `hermes status` shows `Gateway Service Status: ✗ stopped`.
2. The Electron desktop app relies on the gateway being active to process messages. With the gateway down, the message is enqueued but never dispatched.
3. The desktop app has no visible timeout or fallback when the gateway is unreachable, so the UI silently blocks.
4. The CLI bypasses the gateway entirely (it runs a local agent loop directly), so it remains functional.

---

## Temporary Fix

Restart the gateway manually:

```bash
hermes gateway restart
```

After this, the desktop app recovers and messages are sent normally.

---

## Proposed Fixes

1. **Update flow should not leave the gateway in a crashed state.**
   - After `hermes update` completes, the bootstrap runner should ensure the gateway is cleanly restarted.
   - If the gateway fails to start post-update, the error should be surfaced to the user instead of being silently swallowed.

2. **Desktop app should detect an unreachable gateway and surface an error.**
   - Show a "gateway unavailable" state in the UI after a timeout.
   - Offer a "Retry" or "Restart Gateway" button.
   - Do not leave the user staring at an infinite "sending" spinner.

3. **`hermes update` should be idempotent and self-healing.**
   - If the gateway is detected as unhealthy after an update, the update command itself should attempt a restart, or at minimum print a clear message: "Gateway restart required. Run `hermes gateway restart`."

---

## Workaround Script

A `fix-hermes.bat` workaround has been created that performs gateway restart + venv regeneration + hermes.exe re-copy. This script will be attached/updated as we refine the exact recovery steps.

---

## Logs (from the reported incident)

- `~/.hermes/logs/gateway-exit-diag.log` — records gateway process exit with non-zero status after update.
- `~/.hermes/logs/agent.log` — shows API calls succeeding on the CLI side during the same window.
- `~/.hermes/logs/errors.log` — was empty during desktop app hang (no error surfaced to user).
- No `desktop.log` was produced during the hang, indicating the Electron main process failed before reaching its logging stage.

pmos69:
## Summary

Fixes the Desktop gateway restart/recovery path so a local gateway that becomes stale after Windows sleep/resume or `hermes update` does not leave Hermes Desktop stuck in an indefinite Sending state.

The previous Desktop restart flow ultimately relied on `hermes gateway restart`. Recent hermes-agent builds can run that command as a long-lived foreground gateway process on Windows, which is why live testing showed a visible `cmd.exe` window with "Hermes Gateway Starting...". This PR switches the GUI restart path to a native Desktop-managed restart: stop the current gateway, wait for `/health` to drop, start the gateway through the existing hidden `startGateway()` path, then wait for `/health` to become ready again.

It also makes the chat send path recover a stale local API cache. If Desktop believes the gateway is ready but the local API request fails with `ECONNREFUSED`, Desktop now recovers the gateway and retries once. If the request was accepted but times out, Desktop recovers the gateway without replaying the user message, avoiding duplicate model/tool execution and surfacing a clear resend-needed error.

Closes #492.

## Details

- Adds `startGatewayWithRecovery()` for local start attempts that fail to become healthy, falling back to the same native restart transaction.
- Makes `restartGateway()` async and health-based, returning `true` only when `/health` is actually ready.
- Requires the old gateway to become unhealthy before spawning a replacement, avoiding false success when the old process never stopped.
- Restores previously tracked gateway state or PID-file state if the restart cannot stop the old gateway, so the UI does not lose track of a still-running gateway.
- Uses the existing hidden subprocess options for the replacement gateway, avoiding the visible Windows terminal window seen with the CLI restart command.
- Adds a `restart-gateway` IPC/preload API and a Restart button on the Gateway screen.
- Keeps restart failures visible in the Gateway screen while refreshing the real running/stopped status.
- Leaves `restartGatewayViaCli()` as a compatible helper/test target, but the GUI IPC path now uses the native restart flow.
- Removes the eager `send-message` IPC gateway start and lets `sendMessage()` perform bounded recovery, so local chat sends do not bypass the health-checked recovery path.

## Validation

Automated:

- `npm test -- tests/hermes-cli-session-id.test.ts` — 7 passed
- `npm test -- tests/gateway-restart.test.ts tests/hermes-cli-session-id.test.ts tests/remote-mode-url-and-spawn.test.ts tests/ipc-handlers.test.ts tests/preload-api-surface.test.ts src/renderer/src/screens/Gateway/Gateway.test.tsx` — 231 passed
- `npm run typecheck:node` — passed
- `npm run typecheck:web` — passed
- `git diff --check` — clean except normal Windows CRLF warnings

Live test on Windows:

- Launched a dedicated Electron dev instance from this branch.
- Verified a normal live chat send returned `OK` while gateway stayed healthy.
- Killed the active local gateway process externally via `gateway.pid`, without using Desktop's Stop button.
- Sent through `window.hermesAPI.sendMessage(...)`; Desktop recovered the gateway and got `OK` in about 12.7s.
- Killed the recovered gateway externally again.
- Sent from the visible Chat UI; Desktop recovered and the UI displayed the new `OK` response.
- Confirmed final `/health` returned `200 {"status":"ok","platform":"hermes-agent"}`.

### DIFF
--- a/src/main/hermes.ts
+++ b/src/main/hermes.ts
@@ -3,6 +3,7 @@ import { randomUUID } from "crypto";
 import {
   existsSync,
   readFileSync,
+  writeFileSync,
   appendFileSync,
   unlinkSync,
   mkdirSync,
@@ -240,22 +241,26 @@ interface ChatHandle {
 
 function isApiServerReady(profile?: string): Promise<boolean> {
   return new Promise((resolve) => {
-    const url = `${getApiUrl(profile)}/health`;
-    const mod = url.startsWith("https") ? https : http;
-    const req = mod.request(
-      url,
-      { method: "GET", timeout: 1500, headers: getRemoteAuthHeader() },
-      (res) => {
-        resolve(res.statusCode === 200);
-        res.resume();
-      },
-    );
-    req.on("error", () => resolve(false));
-    req.on("timeout", () => {
-      req.destroy();
+    try {
+      const url = `${getApiUrl(profile)}/health`;
+      const mod = url.startsWith("https") ? https : http;
+      const req = mod.request(
+        url,
+        { method: "GET", timeout: 1500, headers: getRemoteAuthHeader() },
+        (res) => {
+          resolve(res.statusCode === 200);
+          res.resume();
+        },
+      );
+      req.on("error", () => resolve(false));
+      req.on("timeout", () => {
+        req.destroy();
+        resolve(false);
+      });
+      req.end();
+    } catch {
       resolve(false);
-    });
-    req.end();
+    }
   });
 }
 
@@ -266,11 +271,12 @@ function delay(ms: number): Promise<void> {
 async function waitForApiServerReady(
   timeoutMs = 8000,
   profile?: string,
+  pollMs = 250,
 ): Promise<boolean> {
   const deadline = Date.now() + timeoutMs;
   while (Date.now() < deadline) {
     if (await isApiServerReady(profile)) return true;
-    await delay(250);
+    await delay(pollMs);
   }
   return false;
 }
@@ -801,10 +807,10 @@ function sendMessageViaApi(
     finish(`API request failed: ${err.message}`);
   });
   req.on("timeout", () => {
-    req.destroy();
     finish(
       "API request timed out. Check the SSH tunnel and remote Hermes gateway.",
     );
+    req.destroy();
   });
 
   req.write(bodyBuf);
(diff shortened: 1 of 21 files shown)


pmos69:
@greptileai please re-review the latest PR head. The previous review was on 2fa1a5e; the current head b41eeb1 addresses those findings.

PR #504, opened 2026-06-02
fix: resolve xAI Grok OAuth callback loop and model configurations

saxster:
Resolves redirect/callback loop on xAI OAuth flow, corrects model selection logic to support grok-4.3, and ensures named profiles inherit/merge configurations cleanly from xai-oauth.


pmos69 (changes requested):
Thanks for the work across this set. I looked at #504, #505, and #506 together because they are all from the same old base and overlap in ways that make them hard to review or merge independently.

At the moment all three PRs need changes before merge.

Current verified state:

- #504, #505, and #506 are all `DIRTY` against current `main` and have no status checks reported.
- Local merge-tree checks reproduce conflicts:
  - #504 conflicts in `src/renderer/src/screens/Sessions/Sessions.test.tsx` after recent Sessions work.
  - #505 conflicts in `src/main/sessions.ts`.
  - #506 also conflicts in `src/main/sessions.ts`.
- All three branches are based on the same older base (`81c7349`, around v0.5.4), so recent merged work like the Sessions changes and Spanish docs/i18n changes appears as stale-branch noise when comparing against current `main`.

The biggest issue is that the three PRs are not independent:

1. #505 imports and calls `detectAuthUrl()` from `src/main/hermes-auth.ts`, but #505 does not define that helper. The helper is introduced in #504. That means #505 currently appears to depend on #504 while being presented as an independent DB/process/SSH PR. Please either make #505 explicitly stacked on #504, or remove that OAuth change from #505 and keep it in #504 only.

2. #506 duplicates/supersedes part of #505. Both add/refactor the shared SQLite connection layer (`src/main/db.ts`, `src/main/sessions.ts`, `src/main/session-cache.ts`, and related tests). #505 is the cleaner, narrower vehicle for DB caching/process/SSH robustness. Please remove the DB/session-cache slice from #506 if #505 remains open.

3. #506 is too broad as a single PR. It currently mixes several unrelated changes:
   - safeStorage encryption for `desktop.json.remoteApiKey`
   - async media reads
   - per-session chat cancellation
   - file-permission hardening
   - bundled installer scripts under `resources/`
   - SQLite caching already covered by #505
   - renderer/Layout/Memory/Skills tweaks

   Those should be split into focused PRs. In particular, I would not merge the bundled installer scripts as part of this PR; vendoring large installer scripts into Desktop needs a separate justification and maintenance plan, if it is wanted at all.

Suggested path to make this mergeable:

1. Rebase all three branches onto current `main`.
2. Make #504 narrow and first in the stack: xAI OAuth/model handling only. Keep `detectAuthUrl()` and its tests there. Drop broad formatting/docs/script churn unless directly necessary.
3. Make #505 either independent or explicitly stacked on #504. If independent, remove the `detectAuthUrl()` usage from #505. Keep #505 focused on DB caching, Windows process cleanup, and SSH startup robustness.
4. Rework #506 so it no longer includes DB caching or unrelated UI/refactor/install-script changes. Split the remaining useful pieces into separate PRs, for example:
   - safeStorage encryption for `remoteApiKey`
   - async media reading
   - per-session chat cancellation
   - file permission hardening
5. After rebasing/splitting, please rerun the relevant tests and update the PR bodies with the new scope and test results.

There are useful ideas here, but in the current shape the overlap and stale base make it too risky for the maintainer to merge or review confidently.

saxster:
Closing — opened against the wrong repo by mistake; I'm keeping this work in my own fork (saxster/hermes-desktop). Apologies for the noise!

PR #530, opened 2026-06-03
feat(i18n): add complete Hebrew (he) translation with RTL support

yosef-chai:
## Summary

This PR adds **Hebrew (`he`)** as a fully supported interface language, on par with the existing locales. Every UI namespace is translated, and the app now renders **right-to-left** when Hebrew is active.

## What's included

- **Complete `he/` locale** — all **23 namespaces** translated against the English source of truth (`common`, `navigation`, `chat`, `settings`, `setup`, `tools`, `sessions`, `models`, `providers`, `office`, `errors`, `schedules`, `skills`, `gateway`, `agents`, `soul`, `memory`, `install`, `constants`, `kanban`, `discover`, `welcome`, `diagnose`).
  - All interpolation placeholders (`{{version}}`, `{{count}}`, `{{path}}`, …) are preserved.
  - Inline HTML (`<code>`, `<strong>`) is kept intact.
  - Product, brand, and technical identifiers (`Hermes`, `Claude`, `OpenAI`, `Telegram`, `config.yaml`, env-var names, …) are kept in Latin script, as is convention for Israeli software UIs.
- **Locale registration** — `he` added to `AppLocale` (`types.ts`), `APP_LOCALES` (`config.ts`), and the `resources` map (`index.ts`); native name **"עברית"** added to the language picker in Settings.
- **RTL support** (new) — Hebrew is a right-to-left language and the app previously had no RTL handling:
  - New `RTL_LOCALES` list and `getLocaleDirection()` helper in `config.ts`.
  - `I18nProvider` now sets `<html dir="…" lang="…">` from the active locale (at init and on every locale change), so the entire UI mirrors correctly for RTL languages and reverts to LTR otherwise.
- **Tests** — the shared i18n test suite is extended to cover Hebrew translation lookup and RTL direction resolution.

## Verification

- `npm run typecheck` — passes
- `npm test` — 840 passed, 3 pre-existing skips
- `npm run lint` — 0 errors

## Notes

- The RTL change is additive and behavior-neutral for existing LTR locales (`getLocaleDirection` returns `"ltr"` for every non-Hebrew locale).
- No dependencies, build config, or unrelated files were touched.


pmos69:
Thanks for the very thorough Hebrew localization work. The RTL approach looks good overall: adding `getLocaleDirection()`, registering `he`, adding `עברית` to the language picker, and setting `<html dir/lang>` from `I18nProvider` all make sense.

This branch is now quite far behind current `main`, though, and the English source locale has moved since the PR was opened. I merged the PR onto current `main` locally and compared the Hebrew key tree against `locales/en`; Hebrew is currently missing 207 leaf keys. The largest gaps are in `gateway`, `tools`, `settings`, `office`, and `chat`, with smaller gaps in `constants`, `diagnose`, `kanban`, `navigation`, `sessions`, and `skills`.

Could you please rebase on current `main` and fill in the newly added keys? Once the Hebrew locale matches the current English key tree, this should be much easier to merge.

pmos69:
Thanks for updating this after the previous review. The RTL plumbing still looks good, and the PR now merges cleanly onto current `main`.

However, `main` has moved again since the update. I merged this locally into today’s `main` and compared `locales/he` against the current English source locale. Hebrew is currently missing 77 leaf keys, mostly in `constants`, `agents`, `chat`, and `models`, with smaller gaps in `common`, `gateway`, `navigation`, `sessions`, and `setup`.

Could you please rebase on current `main` again and fill in the newly added Hebrew keys? Once the Hebrew key tree matches `locales/en` again, this should be close.

PR #578, opened 2026-06-06
fix: gracefully disable local Hermes CLI operations in SSH/Remote mode

andreab67:
## Summary
- Added \
equireLocalHermes()\ helper to check if local operations are allowed before attempting to spawn the local Python venv.
- Returns a clear error message instead of throwing a confusing \spawnSync ... ENOENT\ error when the local venv is missing and the app is configured for SSH/Remote mode.
- Applied this check to profiles, kanban, skills, installer (doctor/update/migrate), and hermes (chat/gateway) modules.

## Why
When the user configures the desktop app to connect to a remote hermes instance via SSH, the app should not attempt to run local Hermes CLI commands. Previously, it would try to spawn the local venv's \python.exe\, which doesn't exist, resulting in an unhandled \ENOENT\ error bubbling up to the UI.

## Testing
- Verified that in SSH mode, local operations now return a descriptive error message instead of crashing with ENOENT.
- Local mode continues to work as expected when the venv is properly installed.


fathah:
conflict

pmos69:
Thanks for working through this. I took another pass now that #665 and the later `main` changes have landed.

I don’t think we should try to merge this branch as-is anymore. It is now far behind `main` and conflicts in several places (`package*.json`, `hermes.ts`, `installer.ts`, `kanban.ts`, `profiles.ts`, `skills.ts`). More importantly, #665 changed the shape of the remote/SSH handling quite a bit, so several of the guards/routes here are now either redundant or would need to be placed very carefully.

The main example is Kanban: current `main` now supports SSH mode by dispatching Kanban operations through `sshRunKanban()` before falling back to the local CLI. A blanket `requireLocalHermes()` guard at the top of that path would now block a valid SSH flow. Similarly, profile create/list/delete and skill install/uninstall are already routed at the IPC layer in current `main`, and `startGatewayDetailed()` already refuses local gateway starts in remote/SSH mode.

So I think the best path is to close/re-cut this as a much smaller PR against current `main`, limited to any remaining local-CLI guard gaps. When doing that, please keep the SSH-specific paths before any “local Hermes required” check, so SSH mode can continue to use the remote helpers. I would also drop the dependency bump from this change.

One piece that may still be worth keeping separately is the lazy `modelsFile()` cleanup in `models.ts`, but I’d prefer that as its own small PR if we want to carry it forward.

fathah:
Superseded

fathah:
For the record after close — verified against current `main`:

- **Already guarded**: cronjobs and kanban route through `isRemoteMode()` / SSH branches today, and the skills IPC handlers route SSH to `sshInstallSkill` / `sshListInstalledSkills` et al. Much of what this branch guards landed independently via the SSH transport overhaul (#786) and earlier fixes.
- **The one still-valid finding here** — remote (HTTP) mode falling through to the local CLI for the Skills surface, showing and mutating the wrong machine's skills — is now fixed in #824, which routes remote mode to the dashboard's `/api/skills*` endpoints instead of disabling the surface.
- The 600+-line lockfile churn and the pre-#786 base were the main review obstacles; for future PRs, keeping `package-lock.json` out of fix branches makes review much faster.

Thanks for the report — it's what surfaced the remaining skills gap.

PR #579, opened 2026-06-06
fix: route list-mcp-servers to remote SSH in remote mode

andreab67:
## Summary
- Added \sshListMcpServers()\ in \src/main/ssh-remote.ts\ that reads and parses the \mcp_servers\ section of the remote \config.yaml\ via SSH exec.
- Updated the \list-mcp-servers\ IPC handler in \src/main/index.ts\ to route to \sshListMcpServers()\ when connection mode is \ssh\ — matching the pattern already used for \get-toolsets\, \set-toolset-enabled\, and other remote-aware IPC handlers.

## Why
The \list-mcp-servers\ IPC handler always called the local \listMcpServers()\ which reads \config.yaml\ from the local filesystem. In SSH/Remote mode this returned an empty list because the actual config lives on the remote VPS — so the MCP Servers tab in the desktop app was always empty.

## Testing
- Verified the new SSH function reads and parses the remote \mcp_servers\ block correctly.
- Verified the IPC handler routes correctly based on \getConnectionConfig().mode === 'ssh'\.


fathah:
conflict

pmos69:
Thanks for this. After #665 and the later main changes, I don’t think we should merge this branch as-is.

The original bug this was addressing has largely moved: current `main` now routes MCP server management through `src/main/mcp-servers.ts`, and in SSH mode `listMcpServers()` uses the SSH tunnel/remote gateway API (`/api/mcp/servers`) rather than reading the local config. So the direct `sshListMcpServers()` implementation here is now stale relative to the current architecture.

There are also practical merge blockers: this branch is very far behind `main`, conflicts, includes an unrelated dependency bump, and duplicates MCP parsing logic that now exists in `parseMcpServersFromConfig()`. If we still want older-gateway compatibility, I think that should be a small new PR against current `main`, probably in `mcp-servers.ts`, reusing the shared parser or adding a graceful unsupported/fallback path for `/api/mcp/servers`.

The lazy `modelsFile()` change may still be worth considering separately if we can reproduce the TDZ crash, but I’d keep that out of this MCP routing PR.

fathah:
listMcpServers on main is already mode-aware (mcp-servers.ts:670 routes to /api/mcp/servers), and register.ts imports that one. Rest duplicates [#578](https://github.com/hermes/hermes-desktop/pull/578) + 689-line lockfile churn.

PR #604, opened 2026-06-09
feat(chat): render gateway clarify.request as an inline card

0xr00tf3rr3t:
## Summary
- render a mid-turn `clarify.request` as an inline card in the chat transcript instead of dead-ending the turn with "Hermes One does not yet expose that gateway dialog"
- show choice buttons when the agent offers choices, otherwise an open-ended textarea, plus a "skip — let Hermes decide" control that answers empty so the agent proceeds autonomously rather than hanging
- follow the existing live-stream pattern (`tool_event` / `reasoning`): a new `ClarifyMessage` variant on the `ChatMessage` union and one `chat-clarify-request` / `clarify-respond` IPC channel pair — no new window
- key pending requests by `request_id` in the main process and clear them on turn end so an abandoned turn leaks no resolver; the card is idempotent on `request_id` and flips to a read-only resolved state once answered
- leave `sudo.request` / `secret.request` handling unchanged — that carries its own security-review surface and is a deliberate follow-up

## Testing
- `npm run typecheck`
- `npm run build`
- `npm run lint -- --quiet` (no new problems from this change; one pre-existing error in `src/main/ssh-remote.ts` is untouched here)
- `git diff --check`
- `npm test` — 1044 passing (+5 new `ClarifyCard.test.tsx` cases, no regressions)
- verified live via `npm run dev`: the `clarify.request → inline card → clarify-respond` round-trip works end-to-end for both choice buttons and the open-ended textarea, and the turn continues after answering

## Notes
- The `ClarifyMessage` variant is added the same way `reasoning` / `tool_call` / `tool_result` already extend the `ChatMessage` union, and the IPC channel mirrors the existing `onChatToolEvent` registration.

### DIFF
--- a/src/main/hermes.ts
+++ b/src/main/hermes.ts
@@ -910,6 +910,41 @@ export function extractReasoningDelta(delta: unknown): string {
   return "";
 }
 
+/**
+ * Pending clarify requests, keyed by the gateway `request_id`. When the agent
+ * asks a clarifying question the stream handler registers a resolver here (a
+ * closure over the live gateway client) and surfaces the question to the
+ * renderer. The renderer's answer arrives via the `clarify-respond` IPC handler,
+ * which calls `resolvePendingClarify` to fire the resolver and forward the
+ * answer to the gateway. Entries are one-shot and self-clear on use; the stream
+ * handler also clears any leftover on turn end so an abandoned turn can't leak a
+ * stale resolver.
+ */
+const pendingClarify = new Map<string, (answer: string) => void>();
+
+export function registerPendingClarify(
+  requestId: string,
+  resolver: (answer: string) => void,
+): void {
+  pendingClarify.set(requestId, resolver);
+}
+
+/** Fire and remove the resolver for `requestId`. Returns true if one was waiting. */
+export function resolvePendingClarify(
+  requestId: string,
+  answer: string,
+): boolean {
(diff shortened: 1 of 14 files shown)


0xr00tf3rr3t:
Thanks for the review — both points addressed in 2817c39:

**1. Transcript ordering after reconcile.** Confirmed: a clarify card is renderer-only and has no `reconciliationKey`, so `reconcileStreamedWithDb` flushed it to the suffix and it rendered below any post-answer agent content. Rather than give it a `reconciliationKey` (it has no DB twin to key against), I added a pure post-pass `repositionClarifyCards()` that re-anchors each card immediately after its streamed predecessor, with a safety net so a card is never dropped if that predecessor was deduped away. Covered by two new cases in `reconcile-streamed-with-db.test.ts` (mid-stream card + leading card).

**2. Failed delivery silently resolving.** Fixed: `submit()` no longer resolves in a `finally`. It now flips the card to resolved only on confirmed delivery; if `respondClarify` rejects or returns `false` (no pending request — e.g. the turn already ended), it surfaces an inline error and keeps the controls live for a retry. Two new `ClarifyCard.test.tsx` cases cover the returns-`false` and rejects paths.

Full suite 1048 pass (+4), `npm run typecheck` / `npm run build` green. Diffs kept surgical — no formatting sweep of existing lines.

PR #606, opened 2026-06-09
feat(chat): handle gateway sudo.request / secret.request via hardened modal

0xr00tf3rr3t:
## Summary
- handle the gateway's `sudo.request` / `secret.request` prompts instead of dead-ending the turn with "Hermes One does not yet expose that gateway dialog" (the same dead-end #604 removed for `clarify.request`)
- a sudo password / secret value is **sensitive**, so — unlike the inline clarify card — it is collected in the app's existing **hardened askpass modal** (CSP `default-src 'none'`, `script-src 'none'`, sandboxed, `contextIsolation`, ephemeral data-URL) and **never** rendered into the chat transcript or persisted
- `showPasswordDialog` is parameterized with `title` / `heading` (defaults unchanged, so the installer's existing call is untouched) and exported; `gatewayPrompt.ts` wraps it as `promptSudoPassword()` / `promptSecretValue(envVar, prompt)`
- answers are forwarded with the gateway's exact shapes — `sudo.respond { request_id, password }` and `secret.respond { request_id, value }` — and **Cancel maps to an empty answer**, a safe skip the gateway already handles (`secret.request` → `{ skipped: true }`; sudo -S -p '' lets the terminal prompt fail cleanly), so the turn never wedges

## Why a modal, not an inline card
`clarify.request` (#604) is a non-sensitive question → inline transcript card. A sudo password or secret must never land in scrollback, so this reuses the installer's already-hardened askpass dialog rather than introducing a new surface. Same trust boundary the app already uses to collect the installer's sudo password.

## Testing
- `npm run typecheck`
- `npm run build`
- `npm run lint -- --quiet` (no new problems from this change; one pre-existing error in `src/main/ssh-remote.ts` is untouched here)
- `git diff --check`
- `npm test` — 1052 passing (+4 new `askpass-security.test.ts` cases: modal reuse, no-log/no-persist, exact respond shapes, cancel→skip)

## Note
Stacked on #604 (inline clarify card) — that branch is its base, so the diff shows the clarify commits too until #604 merges. **The only new commit here is the top one** (`feat(chat): handle gateway sudo.request / secret.request via hardened modal`). Happy to rebase onto `main` once #604 lands.


0xr00tf3rr3t:
Addressed Greptile's finding + added security provider integration:

**1. `finished` guard** — added `if (finished) return;` at the top of the `.then()` callback (commit `c07534e`). A turn cancelled while the askpass modal is open no longer dispatches credentials to a dead gateway session.

**2. Vault-first secret resolution** — for `secret.request`, the handler now calls `getSecret(env_var, profile)` from the configured security provider before falling back to the interactive modal. If the vault already holds the key (e.g. it was loaded for the provider at chat-start), the secret is answered silently without prompting the user. `sudo.request` always goes to the modal — no vault lookup applies to passwords.

Rebased onto current main. `typecheck:node` clean, `askpass-security.test.ts` 10/10 passing.

PR #612, opened 2026-06-09
Add AIML API provider preset

### LINKED ISSUE #601: Provider request: AIML API preset for OpenAI-compatible models
﻿Hi! Would you be open to adding AIML API as a provider preset?

AIML API exposes an OpenAI-compatible API, so the integration should be relatively small:

```ts
baseURL: "https://api.aimlapi.com/v1"
apiKey: process.env.AIML_API_KEY
```

Why it may be useful for Hermes Desktop users:

- one OpenAI-compatible endpoint for many models;
- easy configuration alongside existing OpenAI-compatible providers;
- another fallback/provider option for users who already switch between model backends.

I am happy to implement the provider config/docs PR myself if this fits the project direction. A minimal first pass could include:

- AIML API provider preset;
- `AIML_API_KEY` env/config docs;
- one or two model examples;
- a small smoke-test/example path if the repo already has provider tests.

Thanks!

D1m7asis:
## Summary
- add AIML API as a built-in OpenAI-compatible provider preset
- wire `https://api.aimlapi.com/v1` to `AIMLAPI_API_KEY` across provider registry, URL/env-key mapping, setup/settings UI, and URL detection
- keep model selection on existing `/v1/models` auto-discovery rather than adding hardcoded default models

Closes #601

## Tests
- `npm test -- tests/provider-registry.test.ts tests/constants.test.ts tests/detect-provider.test.ts tests/model-discovery.test.ts tests/set-model-config-base-url.test.ts tests/url-key-map.test.ts tests/compat-host-derived-key.test.ts tests/install-gate-providers.test.ts`
- `npm run typecheck`
- `npm run lint` (exits 0; emits existing Windows CRLF Prettier warnings across the checkout)
- `npm test` (fails only in `tests/ssh-remote.test.ts`: the test environment cannot find a `hermes` CLI on the remote PATH or known venv locations)


D1m7asis:
Addressed the i18n gap from the review: AIML API provider UI now uses constants.* keys, and all existing locale constants files include AIML API name/description/preset/API-key/hint entries. Re-ran targeted provider tests, typecheck, and lint.

D1m7asis:
Follow-up: I tested the actual Hermes Desktop send path with AIML API, not just model discovery. The current Hermes Agent build does not know provider id аimlapi natively yet, so Desktop now bridges this preset through the existing custom CLI provider with CUSTOM_BASE_URL and the host-derived AIMLAPI_API_KEY

Verification run
Also added a regression test for the Desktop CLI bridge so AIML API keeps launching as --provider custom while preserving AIMLAPI_API_KEY.

PR #625, opened 2026-06-10
fix: replace prefix/suffix with interleave in reconcileStreamedWithDb

hankkyy:
## Problem

`reconcileStreamedWithDb` uses a prefix/suffix model that incorrectly places unconsumed streamed messages at the top or bottom of the merged transcript, instead of interleaving them at their correct chronological positions.

### Root Cause

The old code finds `firstConsumedIndex` (the earliest streamed message that matched a DB row — almost always 0, the first user message) and splits the streamed array into:

- **prefix**: messages before `firstConsumedIndex` → rarely populated
- **suffix**: messages at/after `firstConsumedIndex` that weren't consumed → **includes mid-conversation messages**

Since `firstConsumedIndex` ≈ 0, nearly ALL unconsumed messages from the middle of the conversation land in the suffix → at the bottom, below later DB-matched messages.

### Symptoms (user reports)

1. Agent messages from previous turns appear BELOW newer user messages instead of being interleaved
2. When the DB misses middle turns (e.g. due to session split or write timing), the ordering appears reversed — newer messages on top, older at bottom

### Scenario

```
Streamed: [turn1-u, turn1-a, turn2-u, turn2-a, turn3-u, turn3-a]
DB:       [turn1-u, turn1-a, turn3-u, turn3-a]  (turn 2 missing)
```

- **Old**: `[turn1-u, turn1-a, turn3-u, turn3-a, turn2-u, turn2-a]` ← turn 2 at bottom (BUG)
- **New**: `[turn1-u, turn1-a, turn2-u, turn2-a, turn3-u, turn3-a]` ← correct interleaved order

## Fix

Replace prefix/suffix accumulation with an interleave that walks the streamed array in order, placing each unconsumed message at its correct chronological position between consumed ones. Remaining DB-only rows (reasoning, late-written splits) are appended at the end.

The dedup logic (`isBubbleDup`, `isCoveredByDbBubbleSplit`, `consumeCanonicalToolMatch`) is preserved and integrated into the interleave pass.

## Test Plan

The `reconcileStreamedWithDb` function is pure — the fix can be verified by running the existing test suite:

```bash
npx vitest run tests/reconcile-streamed-with-db.test.ts
```

Or by reproducing with this scenario:

```js
const streamed = [
  {id:'u-1',role:'user',content:'hello'},
  {id:'a-1',role:'agent',content:'Hi'},
  {id:'u-2',role:'user',content:'how'},
  {id:'a-2',role:'agent',content:'good'},
  {id:'u-3',role:'user',content:'time?'},
  {id:'a-3',role:'agent',content:'3pm'},
];
const db = [
  {id:'db-1',role:'user',content:'hello'},
  {id:'db-2',role:'agent',content:'Hi'},
  // turn 2 missing
  {id:'db-5',role:'user',content:'time?'},
  {id:'db-6',role:'agent',content:'3pm'},
];
// Expected: u-1, a-1, u-2, a-2, u-3, a-3
// NOT: u-1, a-1, u-3, a-3, u-2, a-2
```

Closes: user-reported chat ordering issues where:
- Agent messages appear below instead of interleaved with user messages
- Newer messages appear at top when DB misses middle turns

### DIFF
--- a/src/renderer/src/screens/Chat/sessionHistory.ts
+++ b/src/renderer/src/screens/Chat/sessionHistory.ts
@@ -381,13 +381,20 @@ export function reconcileStreamedWithDb(
   const shouldKeepUnconsumed = (m: ChatMessage): boolean => {
     if (consumedIds.has(m.id)) return false;
     if (consumeCanonicalToolMatch(canonicalToolMatchCounts, m)) return false;
+    // Only bubble messages (user/agent content) need dedup and split checks.
     if (!("kind" in m)) {
       const bubble = m as ChatBubbleMessage;
       const contentKey = `${bubble.role}:${normalizeBubbleContentForMatch(bubble.content || "")}`;
+      // Check for duplicates first.
       if (seenBubbleKeys.has(contentKey)) return false;
-      if (isCoveredByDbBubbleSplit(bubble, dbAssistantSplitSequences)) {
+      // For agent bubbles, also check if covered by a DB split before marking as seen.
+      if (
+        bubble.role === "agent" &&
+        isCoveredByDbBubbleSplit(bubble, dbAssistantSplitSequences)
+      ) {
         return false;
       }
+      // Only mark as seen if we're actually keeping this message.
       seenBubbleKeys.add(contentKey);
     }
     return true;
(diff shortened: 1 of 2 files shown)


fathah:
Conflict

PR #644, opened 2026-06-11
feat(secrets): pluggable secrets-provider module (env + command backends)

0xr00tf3rr3t:
## Summary
- Add an opt-in secrets-provider abstraction so API keys can be resolved from an external secret store at runtime instead of plaintext `~/.hermes/.env`. **This PR is the self-contained core** — the interface + two backends + their tests. It is purely additive: nothing consumes it yet (wiring is a follow-up PR), so it changes no behavior on its own.
- **`env`** backend (default): reads the existing `.env` via `readEnv()`. Its `list()` *is* the `.env` map, so the default path is byte-for-byte unchanged.
- **`command`** backend: runs a user-configured helper that prints the secret on stdout (bare value) or a `KEY=VALUE` dotenv blob (list mode). The requested key is passed as **data** via the `HERMES_SECRET_KEY` env var — never interpolated into the shell string — so a hostile key name is inert (no injection surface).
- Main-process hardening (resolution is synchronous on Electron's main process): 3s timeout + 1 MiB output cap, stderr piped & discarded, structured-only failure logs (never the command string / helper stderr), a 5s TTL cache with a 1s hard spawn-floor (a renderer can't wedge the UI via invalidate+poll), and a cross-key misroute guard (a helper emitting the wrong key's line resolves `null` rather than leaking a credential).

## Why split this out first
A provider abstraction is reviewable on its own: it has a clean interface, full test coverage, and zero blast radius until something consumes it. Keeping it as its own PR keeps each diff small and focused per CONTRIBUTING.

## Platform
POSIX-only (runs the helper via `/bin/sh`). On Windows it degrades to `null` and the user stays on the default `env` provider.

## Testing
- `npm run typecheck` — clean
- `npm run lint` — clean for the changed files
- `npm test` (secrets suite) — 58 passing across 5 files: `parseSecretOutput`, `unquoteDotenvValue`, the command provider (bare/dotenv/quote/CRLF/timeout/injection/failure-logging), the spawn floor, `resolvedSecretMap`. Each security property has a regression test that fails without the guard.


0xr00tf3rr3t:
Thanks for the review — both flagged gaps are real and now fixed in bcf0e33:

1. **`resolvedSecrets()` spawn-floor bypass** — it now routes through `providerListSafe()` instead of calling `provider.list()` directly, so the S1 TTL cache + 1s hard spawn floor cover this path too. Regression test in `spawnRateFloor.test.ts` asserts repeated `resolvedSecrets()` calls spawn the helper exactly once.

2. **`list()` / `get()` whitespace-only disagreement** — `list()` now drops entries whose unquoted value is whitespace-only, matching `get()`/`parseSecretOutput`. So a quoted-blank vault placeholder (`K="  "`) no longer appears as a configured key in `list()` while resolving to null on read. Regression test in `commandProvider.test.ts` asserts `list()` omits the blank entry and agrees with `get()`.

Secrets suite 60 passed; full committed suite 1141 passed / 3 skipped; typecheck + lint clean.

0xr00tf3rr3t:
Good catch — this is a real cross-key leak and it's now fixed (commit `6af26b1`, will land on the next push to this branch).

The S2 guard was matching the **full** trimmed output against the `^`-anchored `ENV_LINE`. As you described, a header comment before a single wrong-key line (`# Generated by keepassxc-cli\nOTHER_KEY=secretvalue`) makes that match return null, the guard never fires, and the whole comment+wrong-key string flows back as the bare value for the wanted key.

Fix is exactly your suggestion: when the comment-stripped `dotenvLines` has exactly one entry, run the guard against that extracted line instead of the full text:

```ts
const s2Target = dotenvLines.length === 1 ? dotenvLines[0] : value;
const envShaped = s2Target.match(ENV_LINE);
```

Two regression tests added:
- `# comment\nOTHER_KEY=secret` for a different wanted key → `null` (both a generic case and a realistic `ANTHROPIC_TOKEN` → `API_SERVER_KEY` misroute).
- `# comment\nWANTED_KEY=value` → still resolves (the fix doesn't break the legitimate header-comment case keepassxc-cli/secret-tool produce).

commandProvider 39 tests; secrets suite 62 passed; typecheck + lint clean.

0xr00tf3rr3t:
Closed the last gap from your review (commit `8921792`): the `get()` single-key path now has a spawn floor too.

`getSecret()` routes through a new `getSecretSafe()` — a per-`(profile,key)` hard floor mirroring `providerListSafe()`'s list() floor — so the per-key resolution the wiring PR introduces can't spawn the helper unbounded on the Electron main process.

One deliberate difference from list(): it caches **no value**. It records only the last-spawn timestamp per key and **degrades to null** inside the floor window, rather than retaining plaintext secrets in a long-lived main-process map (a new at-rest exposure surface the provider design avoids — resolved values are never logged or persisted). list() can serve stale because it already holds the map; get() can't without retaining secrets, so it chooses null over a new exposure surface. The tradeoff is documented inline as an auditable decision. `invalidateProviderListCache()` ("Refresh from vault") also clears the get floor so a refresh isn't swallowed by an open window. Env provider stays a pass-through.

Regression suite `getSpawnFloor.test.ts` (7 cases): one-spawn-then-degrade, no-value-replay, post-floor fresh re-spawn, independent per-key floors, invalidation clears floor, env pass-through, process.env precedence. Tracked secrets suite 77 passed; typecheck:node + eslint clean.

0xr00tf3rr3t:
Fixed (commit `7cd5366`) — good catch, this was a real contract violation. `getSecretSafe()` now wraps its body in `try/catch { return null }`, exactly as your suggestion and matching `providerListSafe()`'s top-level guard. A throwing `getConfigValue()` (malformed config.yaml / first-read I/O error) now degrades that key to null instead of propagating out of `getSecret()` and aborting a per-key gateway-spawn loop.

Regression test added to `getSpawnFloor.test.ts`: a throwing `getConfigValue` mock asserts `getSecret()` does not throw and resolves null. Tracked secrets suite 78 passed; typecheck:node + eslint clean.

PR #650, opened 2026-06-12
feat(secrets): wire the provider into key-resolution paths + vault-aware audit

0xr00tf3rr3t:
## Summary
- Wire the secrets-provider (added in #644, now merged) into the **real key-resolution paths** — this is the follow-up that makes the provider actually do something. Resolution order is preserved everywhere: **`process.env` > `.env` > provider**.
- `getApiServerKey()` gateway-env overlay fills **only** keys absent from both `.env` and (trimmed) `process.env`, so a vault value can never silently override an explicit secret.
- Gateway-spawn env (`buildGatewayEnv` / `tuiGatewayEnv`) resolves through the never-throwing `providerListSafe()`, so a misconfigured vault degrades gracefully instead of aborting a spawn.
- `config-health` audit is vault-aware via `resolvedSecretMap()` — vault-only users no longer get false "API_SERVER_KEY is not set" warnings.
- `invalidate-secrets-cache` IPC busts the `env:` / `apiServerKey:` caches and marks the provider `list()` stale — **no secret read, no unbounded helper spawn** (the 1s spawn-floor survives invalidation).
- Logs emit only provider id + key counts, **never a resolved value**; the migration path masks via `maskKey()`.

## Why lazy provider construction
config.ts statically imports `./secrets`, which closes the cycle `config → secrets → commandProvider → config`. The secrets module therefore must do **zero work at module-init**: it now constructs its providers lazily (`_x ??= new X()`). An earlier eager `new CommandSecretsProvider()` at module scope threw `"is not a constructor"` on load-order-dependent paths — a crash that hid behind `vi.mock("../config")` in unit tests and only surfaced in the unmocked property test. The import-site comment documents the invariant so the trap can't be reintroduced. (Lazy construction was chosen over lazy-importing the barrel because consumers here use `vi.resetModules()`, under which a runtime relative `require` doesn't resolve.)

## Testing
- `npm run typecheck` — `typecheck:node` clean. (`typecheck:web` has one PRE-EXISTING error in `ModelPicker.test.tsx`, unrelated to this PR — present on `main`.)
- `npm run lint` — clean for the changed lines. (3 PRE-EXISTING "unused eslint-disable" warnings in `config-health.ts` are not introduced here.)
- `npm test` — 117 passing across the affected suites, including the **unmocked** `commandProvider.property.test.ts` (which exercises the real import cycle) and `invalidate-secrets-cache.test.ts` (the `vi.resetModules()` path). New coverage: `config-health.test.ts`, `hermes.test.ts`, `tui-gateway-stream.test.ts`, `api-server-key-secrets-provider.test.ts`, `api-server-key-status.test.ts`, `invalidate-secrets-cache.test.ts`.
- `git diff --check` — clean.

## Note
Follow-up to #644 (now merged). This PR is the consumer wiring; a renderer-UI PR for selecting the provider follows. Kept to one cohesive change (provider wiring + the lazy-construction fix that wiring requires) per CONTRIBUTING — the test files are the bulk of the line count.

### DIFF
--- a/README.md
+++ b/README.md
@@ -259,6 +259,87 @@ Hermes files are managed in:
 - `~/.hermes/state.db` — session history database
 - `~/.hermes/cron/jobs.json` — scheduled tasks
 
+## Secrets provider
+
+By default, API keys live in `~/.hermes/.env` (the **env** provider). No
+configuration is needed — this is byte-for-byte the historical behavior, and
+nothing changes for you.
+
+If you'd rather not keep keys in a plaintext `.env`, the opt-in **command**
+provider resolves them by running a helper command you configure. Resolution
+order everywhere is: `process.env` → `.env` → provider → unset.
+
+Per-key helper (the requested key name arrives as `$HERMES_SECRET_KEY`):
+
+```yaml
+# ~/.hermes/config.yaml
+secrets:
+  provider: command
+  command: secret-tool lookup hermes "$HERMES_SECRET_KEY"
+```
+
+Or a helper that dumps a dotenv blob (e.g. a vault that unseals into tmpfs):
+
+```yaml
+secrets:
+  provider: command
+  command: "cat /run/user/1000/hermes-secrets.env"
+```
+
+The helper's stdout may be either a single bare value (per-key helpers) or
+`KEY=VALUE` lines (dotenv dumps); both shapes are auto-detected.
+
+### Vault / secret manager integration (no TPM required)
+
+The `command` provider is **vault-agnostic** — it runs whatever helper you
+configure and reads its stdout. The helper is the only thing that needs to
+talk to your secret store. If you don't have a TPM-sealed keyfile, any of
+these work without code changes to Hermes:
+
+- **KeePassXC (password-only DB, no keyfile):** point `secrets.command` at a
+  small `kpxc-export.sh` script that does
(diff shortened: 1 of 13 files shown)


0xr00tf3rr3t:
Heads-up on the red CI check — it is **pre-existing breakage on `main`, not from this PR**:

- The failing step is **Typecheck** (`typecheck:web`), erroring at `src/renderer/src/screens/Chat/ModelPicker.test.tsx:63` — `Cannot find namespace 'vi'` (a `vi.fn` used as a *type* annotation).
- That file is **not in this PR's diff** — this change touches only `src/main/`, `src/main/secrets/`, `src/shared/i18n/`, `tests/`, and `README.md` (zero `src/renderer/` files), so a renderer typecheck error cannot originate here.
- The broken line landed on `main` in 5393518 ("Add tests for ModelPicker.tsx", 2026-06-10), before this branch's base. #644's CI passed because it was checked against a tree that predated that commit; this PR is just the first CI'd after it landed in the base.

`typecheck:node` (the half this PR actually affects) is clean, as are lint-on-changed-files, gitleaks, and the test suite (117 passing, incl. the unmocked property test that exercises the real import cycle). Happy to rebase once the `ModelPicker.test.tsx` typing is fixed on `main`, but it's out of scope for this PR per CONTRIBUTING (no bundling unrelated fixes).

PR #660, opened 2026-06-13
Attempted to fix ManifestVersion in the build files.

DandelionSprout:
As simple as that, hopefully.


DandelionSprout:
As far as I can tell, the Greptile bot is spot-on or close-to-spot-on on all of the above. That is *very* high praise from me when it comes to Aİ bots.

PR #673, opened 2026-06-15
Security Providers: vault-backed secret management for the desktop (AIR-017→023)

0xr00tf3rr3t:
# Security Providers: vault-backed secret management for the desktop

This PR adds a **Security Providers** capability to the desktop app: choose and test a
secrets provider (KeePassXC/command, etc.), have vault-resolved credentials feed the
gateway, and surface clear config-health diagnostics — all without writing secrets to
plaintext `.env`. It then hardens that feature against a series of real, live-data-found
correctness and credential-routing bugs (the AIR-017→023 series below).

## What it adds

- **Security Providers section in Settings** — pick a provider, test it, see which keys it
  resolves (labeled "Vault Provided"), and Refresh-from-vault.
- **First-run onboarding step** for choosing a secrets provider, including a guided
  KeePassXC vault-create flow (picker + prerequisites, not a dead-end).
- **Vault-aware config diagnostics** — config-health surfaces missing/mismatched
  credentials and never blocks Send on a credential the gateway can actually resolve.
- **Off-main-thread TPM seal + vault db-create** (AIR-016) so the UI never freezes during
  a multi-second crypto operation.

## Correctness / credential-routing hardening (found by live-data testing)

Each of these was found by exercising the feature against a **real vault + real OAuth
token**, not mocks — and each is fixed with a RED-proven regression test:

- **AIR-017** — `secretsProviderStatus` listed vault keys *plus all of process.env* (130
  badges, not 6). Now lists vault-only keys.
- **AIR-018** — Setup didn't detect an Anthropic credential stored as the vault OAuth
  token (`CLAUDE_CODE_OAUTH_TOKEN`); added alias detection + a "use vault / enter a key"
  toggle.
- **AIR-019** — Setup could persist an empty `model.default` (→ `model: ""` → API 400).
  Now never writes an empty model, and a blank field falls back to a model discovered via
  the provider's `/v1/models`.
- **AIR-020** — the env-key-mismatch auto-fix would offer to copy an *unrelated* service's
  token (e.g. `MATRIX_ACCESS_TOKEN`) into `ANTHROPIC_API_KEY` — a credential-bleed.
  Restricted to known aliases.
- **AIR-021** — a populated OAuth-token alias was flagged as a "wrong name" mismatch with a
  one-click "copy into ANTHROPIC_API_KEY" fix that would send the OAuth token as
  `x-api-key` → 401. A populated accepted alias is now treated as **satisfied** (no banner,
  no copy).
- **AIR-022** — the pre-send chat-readiness gate blocked Send with "Missing
  ANTHROPIC_API_KEY" for vault OAuth users. Now accepts the alias.
- **AIR-023** — the credential-name list the desktop forwards from the security provider to
  the gateway (`KNOWN_API_KEYS`) had drifted out of sync with the gateway provider plugins'
  `env_vars`, dropping every OAuth/Bearer name and per-vendor alias (anthropic, gemini,
  copilot, zai, kimi, dashscope, xai, nvidia, …). Extended to mirror the plugins, with a
  **drift-guard test**.

## Tests

Full suite green except 3 pre-existing `reconcile-streamed-with-db` failures that are
unrelated to this work (present on the base). Each AIR fix ships a behavior-contract test
(not a snapshot), and the credential-name set has a drift-guard so it can't silently fall
out of sync with the provider plugins again.

## Notes for reviewers

- The primary `buildGatewayEnv` path already overlayed the full provider secret set; the
  AIR-023 fix brings the CLI-fallback path to parity.
- Credential-name equivalence (canonical key ↔ OAuth/Bearer/alias names) is now consistent
  across all five gates (install, config-health warn, config-health mismatch, Setup detect,
  chat-readiness) plus the env-forward layer.


0xr00tf3rr3t:
Thanks for the review — all three findings addressed:

**P1 — Triplicate `KEY_ALIASES` (no shared source)** ✓ Fixed in `7694532`. The alias map is now a single source of truth in `src/shared/url-key-map.ts` (`KEY_ALIASES` + `aliasesForEnvKey()`) — already the shared module for credential-key names, imported by both main and renderer. `config-health.ts`, `validation.ts`, and `Setup.tsx` all import it; no local copies remain. Added `src/shared/url-key-map.test.ts` as a drift guard, and RED-proved it: removing `CLAUDE_CODE_OAUTH_TOKEN` from the one shared map now reds 5 tests across the shared guard + the validation (AIR-022) and Setup (AIR-018/018b) gates, confirming every gate consumes the single source.

**P1 — `detectExistingVault` hardcodes `keepassxc-cli`** ✓ Fixed in `ab45317`. It now resolves the snap-aware CLI name via `resolveKeepassxcCli()` (matching `createVault`), falling back to the apt name only for display when no CLI is installed yet. On a snap-only system the suggested read command now uses `keepassxc.cli` and works.

**P2/security — `hasBinary` / `keepassxcIsSnap` unquoted shell interpolation** ✓ Fixed in `ab45317`. Both now `shellQuote()` the interpolated binary name before it reaches `/bin/sh -c`, closing the injection surface for any future dynamic caller (all current callers pass literals, so no behavior change today).

Full suite: 1549 passed, 0 failed. The branch was also merged up to current `main` (conflicts resolved, gates RED-proved intact through the merge), so the PR is mergeable.

0xr00tf3rr3t:
**AIR-016 consistency gap — fixed** (`f35c407`).

Good catch. `commandWriteSecret` / `commandDeleteSecret` were still synchronous (`execFileSync`) while `createVault` / `sealKeyFileToTpm` in this PR were made async — so a vault write could block the Electron main thread up to 5s on an already-async IPC path. Now non-blocking.

One implementation note for reviewers: the obvious fix — `promisify(execFile)` — is **wrong** here. The async `execFile` does **not** honor the `input` (stdin) option the sync version used to deliver the secret value; it silently drops it, which would break value delivery with no type error. I caught this against a real `/bin/sh` before it shipped. The fix instead uses `child_process.spawn` and writes the value to `child.stdin` explicitly.

Every security invariant of the original is preserved (and now regression-tested):
- value on **stdin only** — never argv / shell string / env
- key name via `HERMES_SECRET_KEY` env only, validated `^[A-Za-z_]\w*$` before any spawn
- hard timeout kills a hung helper; output capped at 1 MiB
- stderr piped + discarded (it can echo the value) — never inherited
- coarse, secret-free error reasons; single Promise resolution (idempotent settle guard); EPIPE-on-early-exit handled

The two IPC handlers now `await` the result with a defensive `.catch` returning a coarse error. Added spawn-based tests plus timeout / helper-not-found branch coverage — full secrets suite green (162/162), typecheck + lint clean.

Re: the `checkRuntimeEnvKeyMismatch` → `[]` dead-code note (the `UI_RUNTIME_ENVKEY_MISMATCH` issue code + auto-fix handler are now unreachable) — acknowledged, tracking separately to keep this PR scoped to the blocking-write fix.

0xr00tf3rr3t:
Rebased onto current main (31 commits behind → 0). All conflicts resolved:

**Conflicts resolved:**
- `src/preload/index.d.ts` — duplicate `invalidateSecretsCache` entry (upstream added it; our commits added it too — deduplicated)
- `src/renderer/src/screens/Gateway/Gateway.test.tsx` — inline mock object replaced with `createHermesAPIMock()` helper (our version taken)
- `src/renderer/src/screens/Settings/Settings.tsx` — 10s polling replaces `onConnectionConfigChanged` subscription (our version); Refresh-from-vault button added (our version)
- `src/main/index.ts` — additive imports merged (both `type ConnectionConfig` and `secretsProviderStatus`)
- `src/main/config-health.test.ts` — `vi.hoisted` harness conflict resolved; rebase artefacts (`FAKE_VAULT`, stale `ReturnType` reference) cleaned up

**Verification:** `npm run typecheck:node` clean (0 errors). `npm test` — 1602 passed, 3 skipped (pre-existing), 0 failures.

PR #709, opened 2026-06-17
feat: drag-and-drop attachment reordering

hankkyy:
Drag-and-drop reorder attachments in the composer strip. Dragged item fades, drop target shows dashed outline.

### DIFF
--- a/src/renderer/src/assets/main.css
+++ b/src/renderer/src/assets/main.css
@@ -3925,6 +3925,26 @@ body {
   color: var(--text-secondary);
 }
 
+/* Drag-and-drop attachment reorder wrapper */
+.attachment-drag-wrapper {
+  cursor: grab;
+  transition: opacity 0.15s ease;
+}
+
+.attachment-drag-wrapper:active {
+  cursor: grabbing;
+}
+
+.attachment-dragging {
+  opacity: 0.4;
+}
+
+.attachment-drag-over {
+  outline: 2px dashed var(--border-focus, var(--accent));
+  outline-offset: 2px;
+  border-radius: var(--radius-sm);
+}
+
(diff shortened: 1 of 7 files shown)


hankkyy:
## Problem

1. **Context gauge stuck**: % and token count never update after first turn.
2. **Cache hit % nonsensical**: Shows "100%" (307.5k read / 14k used = 2196% → clamped).
3. **Session mixing**: Opening two agent sessions causes messages to leak between tabs.

### Root Cause

1. `contextTokens` used `||` which treats `0` as falsy — stale value kept forever.
2. `cacheHitPct = cacheReadTokens / used` where `used`=latest turn but cache tokens are cumulative.
3. `useChatIPC` and `useDashboardChatTransport` both registered IPC listeners/DB polling simultaneously, causing reconciliation races across sessions.

### Changes

| File | Change |
|------|--------|
| `useChatIPC.ts` L300 | `||` → `??` for contextTokens |
| `useDashboardChatTransport.ts` L886 | Same |
| `useDashboardChatTransport.ts` L596 | Remove `|| undefined` wrapper |
| `ContextGauge.tsx` | Add promptTokens; calc cache % against cumulative |
| `Chat.tsx` L535 | Pass usage.promptTokens; `enabled: !dashboardChatEnabled` |
| `useChatIPC.ts` L47,L78 | Add `enabled` param; skip registrations when disabled |
| `Layout.tsx` L437 | Check for existing sessionId in setRuns updater |

## Verification

- [ ] `npm run typecheck` passes
- [ ] `npx vitest run` passes
- [ ] Context gauge updates each turn; cache % ≤ 100%
- [ ] Two agent sessions open simultaneously — no message leaking when switching tabs

PR #737, opened 2026-06-21
fix: use per-provider env key for custom providers in Models dialog

### LINKED ISSUE #681: Custom providers share a single CUSTOM_API_KEY — per-provider API keys are ignored at runtime
## Problem

When adding multiple OpenAI-compatible custom providers (e.g. micu, cctq, xiaomi), all models end up sharing a single `CUSTOM_API_KEY` environment variable. Per-provider API keys configured in `config.yaml` (`custom_providers[].api_key` or `providers.<name>.api_key`) are not used at runtime.

### 1. `models.ts` — `loadCustomProviders()` hardcodes `provider: "custom"`

```typescript
current = {
    name: m ? m[1].trim() : "Custom",
    provider: "custom",   // ← all custom_providers get "custom", not their actual name
    model: "",
    baseUrl: "",
};
```

This means every custom provider entry in `models.json` has `provider: "custom"`, regardless of the provider's `name` field.

### 2. `Models.tsx` — `handleSave()` only writes key when `formProvider === "custom"`

```typescript
if (formApiKey.trim() && formProvider === "custom") {
    const envKey = expectedEnvKeyForUrl(formBaseUrl.trim());
    await window.hermesAPI.setEnv(envKey, formApiKey.trim());
}
```

Two issues:
- Only triggers for `provider: "custom"` — selecting a named provider from the dropdown skips key persistence entirely
- `expectedEnvKeyForUrl()` maps unknown hosts to `CUSTOM_API_KEY` — all custom providers overwrite the same variable

### 3. `url-key-map.ts` — `expectedEnvKeyForUrl()` fallback

For URLs not in `URL_KEY_MAP` (micuapi.ai, cctq.ai, etc.), the function returns `CUSTOM_API_KEY`. So when the GUI writes the key, all custom providers share one variable.

### 4. `runtime_provider.py` — bare `"custom"` skips `custom_providers` lookup

When `model.provider: "custom"` with a `base_url`, the runtime takes the "direct-alias" path which only checks:
- credential pool
- `OPENAI_API_KEY` (if openai.com URL)
- `_host_derived_api_key(base_url)` (vendor-derived env var)

It does **not** check `custom_providers[].api_key` or `providers.<name>.api_key`. The `_get_named_custom_provider()` lookup (which reads those fields) only runs when the provider is named (e.g. `"custom:micu"` or `"xiaomi"`), not bare `"custom"`.

### 5. `providers:` dict is never read by the desktop's model list

The `providers:` dict format (new-style in config.yaml) is checked by the runtime's `_get_named_custom_provider()`, but the desktop's `loadCustomProviders()` and `seedDefaults()` only read the legacy `custom_providers:` list. So models from `providers:` never appear in the GUI.

## Impact

Users with multiple custom providers (different base URLs, different API keys) cannot use them correctly:
- All models resolve to the same `CUSTOM_API_KEY` at runtime
- Switching between models in the GUI doesn't switch API keys
- The `providers:` dict format (which the runtime supports) is invisible to the GUI

## Suggested Fix

1. **`loadCustomProviders()`**: Use the entry's `name` as the `provider` value (e.g. `"custom:micu"`) instead of hardcoding `"custom"`
2. **`handleSave()`**: Write API key to .env for all custom-like providers, not just `formProvider === "custom"`. Use a per-provider env var name (e.g. `CUSTOM_PROVIDER_<NAME>_KEY`)
3. **`seedDefaults()`**: Also read from `providers:` dict in addition to `custom_providers:` list
4. **`expectedEnvKeyForUrl()`**: Consider returning a per-provider key for unknown URLs, or accept a provider name parameter to generate `CUSTOM_PROVIDER_<NAME>_KEY`

## Workaround

Currently users must manually set host-derived env vars in `.env` (e.g. `MICUAPI_API_KEY`, `CCTQ_API_KEY`) that match what `runtime_provider.py`'s `_host_derived_api_key()` derives from the base URL. This is fragile and undocumented.

## Environment

- hermes-desktop version: latest (main branch)
- OS: Windows 10
- Config: `custom_providers:` with 3 entries (micu, cctq, xiaomi), each with distinct `base_url` and `api_key`

365diascollaboration-prog:
## Problem

When adding multiple custom providers with unrecognised base URLs, they all resolve to the shared `CUSTOM_API_KEY` env var via `expectedEnvKeyForUrl()`. The second provider's key silently overwrites the first, making it impossible to run two custom providers simultaneously. Closes #681.

## Root cause

`expectedEnvKeyForUrl()` returns `CUSTOM_API_KEY` as a fallback for any URL not in `URL_KEY_MAP`. The save path in `handleSave()` and the read-back path in `openEditModal()` both use this function, so all unknown-URL providers share one bucket.

Note: `seedDefaults()` in `models.ts` already writes `CUSTOM_PROVIDER_<NAME>_KEY` correctly for providers seeded from `config.yaml`. This fix brings the GUI dialog into alignment with that existing behavior.

## Fix

Introduce `customProviderEnvKey(name, baseUrl)` — a thin wrapper over `expectedEnvKeyForUrl` that falls back to `CUSTOM_PROVIDER_<NAME>_KEY` (same naming as `seedDefaults`) instead of the shared `CUSTOM_API_KEY` when the URL is unknown.

Three call-site changes, all in `Models.tsx`:
- `handleSave()` — writes the per-provider key
- `openEditModal()` — reads the per-provider key, falls back to `CUSTOM_API_KEY` for keys written by older versions

## Testing

1. Add two custom providers with different unrecognised base URLs and distinct API keys
2. Verify each model persists its own key in `.env` (`CUSTOM_PROVIDER_<NAME>_KEY`)
3. Re-open each model's edit dialog and confirm the correct key is shown

### DIFF
--- a/lat.md/provider-setup.md
+++ b/lat.md/provider-setup.md
@@ -36,6 +36,12 @@ Custom providers in the Models library ([[src/renderer/src/screens/Models/Models
 
 Renaming a custom provider or deleting it changes or removes its derived key; `handleSave` and `handleDelete` call the new [[src/main/config.ts#deleteEnvValue]] (through `hermesAPI.deleteEnv` → the `delete-env` IPC handler, mirroring `setEnv`/`set-env`) to remove the old `CUSTOM_PROVIDER_<OLDNAME>_KEY` line so it doesn't linger in `.env`. [[src/main/ssh-remote.ts#sshDeleteEnvValue]] is the remote-mode equivalent, selected by the same `conn.mode === "ssh"` branch `set-env` already uses.
 
+### Cleanup never deletes a key another entry still resolves to
+
+Before either cleanup path deletes a key, [[src/renderer/src/screens/Models/Models.tsx#envKeyUsedByOtherModel]] checks whether another entry still resolves to it, and skips the delete if so.
+
+Two entries can legitimately derive the *same* key: two custom providers on the same known vendor host both fall back to that vendor's shared key (e.g. `GROQ_API_KEY`), or two unknown-host providers with the same display name derive the same per-name key. `envKeyUsedByOtherModel` scans the in-memory `models` list, excluding the entry being edited/deleted, for any other entry resolving to that key — without this guard, renaming or removing one entry would silently wipe the vendor key a sibling entry still depends on.
+
 ## Provider icons
 
 Each card's logo is resolved by [[src/renderer/src/components/common/BrandLogo.tsx]] from the provider id, falling back to a generic robot for unknown ids.
--- a/lat.md/provider-setup.md
+++ b/lat.md/provider-setup.md
@@ -28,6 +28,14 @@ For compatible/custom endpoints, an inline **API Key** field appears under Base 
 
 Ids the agent can't resolve by id are listed in `OPENAI_COMPATIBLE_BASE_URLS` ([[src/renderer/src/constants.ts]]) — openai, perplexity, and every `LOCAL_PRESETS` chip (local servers + remote endpoints like groq, deepseek, atlascloud, mistral, …). This map MUST contain every preset id, or selecting that chip mis-routes; a test in `tests/constants.test.ts` enforces it. Selecting one autofills its base URL and shows the base-URL field; on save it is persisted as `provider: custom` + `base_url`, which the gateway accepts and uses to host-derive the API key (`runtime_provider._host_derived_api_key`, e.g. `api.groq.com` → `GROQ_API_KEY`). `displayProviderFromConfig` reverse-maps a stored `custom` + known base URL back to the brand id so the dropdown re-selects it on load. Native providers (the gateway hardcodes their base URL) clear the field instead.
 
+## Models library: per-provider keys for unknown URLs
+
+Custom providers in the Models library ([[src/renderer/src/screens/Models/Models.tsx]]) that point at an unrecognized base URL each get their own env key instead of sharing one bucket, and that key is cleaned up when the provider is renamed or deleted.
+
+`expectedEnvKeyForUrl` falls back to the shared `CUSTOM_API_KEY` for any host not in [[src/shared/url-key-map.ts]], which meant every unknown-URL custom provider overwrote the same env var. [[src/renderer/src/screens/Models/Models.tsx#customProviderEnvKey]] derives `CUSTOM_PROVIDER_<NAME>_KEY` from the provider's display name in that case, used by `openEditModal` (read, with a `CUSTOM_API_KEY` fallback for entries saved before this existed) and `handleSave` (write). [[src/main/config.ts#customEndpointKeyResolvable]] mirrors the same derivation — via a call-time `require("./models")` to avoid a static import cycle — so the config-health audit checks every model sharing that base URL, not just the first `readModels()` match.
+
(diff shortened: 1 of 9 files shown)


fathah:
Thanks for the fix. The direction makes sense, but I think this currently only fixes the Models dialog storage/display side, not the full runtime path.

For known custom-compatible hosts, this should still work because we already resolve URL-specific env vars like `GROQ_API_KEY`, `DEEPSEEK_API_KEY`, etc.

The issue is unknown custom base URLs. This PR saves the key as:

`CUSTOM_PROVIDER_<DISPLAY_NAME>_KEY`

But the active chat/gateway path does not consistently read that key. In the normal active config path, unknown custom URLs still rely on `model.api_key`, `CUSTOM_API_KEY`, `OPENAI_API_KEY`, or host-derived keys. So a user can add a custom model, enter an API key, see it saved in the dialog, but chat may still run without that key and fail upstream.

Can we either:

1. make the runtime/readiness/config-health path read the same `CUSTOM_PROVIDER_<DISPLAY_NAME>_KEY`, or
2. keep writing a fallback key the runtime already reads, such as `CUSTOM_API_KEY`, or
3. write/update `model.key_env` / `model.api_key` when that custom model becomes active?

Without one of those, this may regress unknown custom providers even though the dialog looks correct.

365diascollaboration-prog:
Thanks for the detailed review, fathah.

You're right that the original PR only covered the dialog side. I've pushed a second commit that extends customEndpointKeyResolvable() in config.ts to also resolve CUSTOM_PROVIDER_<NAME>_KEY:

- Looks up the model by aseUrl from models.json (the same lookup hermes.ts already does in the gateway path)
- Adds the per-provider key to the candidates set before both the env check and the vault check

This way the config-health path and the runtime gateway path stay in sync — no false MODEL_KEY_MISSING warning for a provider whose key was saved under CUSTOM_PROVIDER_<NAME>_KEY.

The gateway runtime at hermes.ts:startGateway already had the CUSTOM_PROVIDER_<NAME>_KEY fallback (lines 2370–2383), so chat itself should resolve the key correctly once it's written. The gap was only in the health-check side, which this commit closes.

365diascollaboration-prog:
Pushed a follow-up commit addressing all three review points:

1. **Import ordering** — moved `import type { ModelRegistry, ... }` above `customProviderEnvKey()`, so all imports are grouped at the top of the file as suggested.
2. **Orphaned key on rename/delete** — added `deleteEnvValue()` (local) / `sshDeleteEnvValue()` (remote) plus a `delete-env` IPC handler and `hermesAPI.deleteEnv`, mirroring the existing `setEnv`/`set-env` path. `handleSave` now deletes the old `CUSTOM_PROVIDER_<OLDNAME>_KEY` when a rename changes the derived key, and `handleDelete` deletes it when the provider is removed.
3. **Single-match `find()` in the health check** — `customEndpointKeyResolvable` now uses `filter()` over all models sharing a base URL instead of only the first match, so two differently-named custom providers on the same unknown host both resolve correctly.

Added `tests/delete-env-value.test.ts` covering the new `deleteEnvValue()` behavior. Existing suites (`Models.test.tsx`, `config-health.test.ts`) still pass, and `tsc --noEmit` / `eslint` are clean on all touched files.

365diascollaboration-prog:
Closing as obsolete after the provider-owned-models refactor merged in #848. The custom-provider key separation is now handled by customProviderEnvKey() in shared/url-key-map.ts. Will open a smaller follow-up if the config-health audit still misses the per-provider key.

Issue #757, opened 2026-06-22
Bug: Message Text Truncation in Streaming Responses


SpiritLiberation:
---

### Environment
- **Project:** fathah/hermes-desktop (Electron + React + TypeScript)
- **Version:** 0.6.3537
- **Platform:** Linux / Wayland / Electron

### Summary
When the AI assistant generates a long response (especially with interleaved text and tool calls), message text is randomly truncated — some characters/chunks are lost from the middle of the displayed message. Restarting the app shows the full text because it's loaded from the session database.

### Root Cause

**`useDashboardChatTransport.ts:839-841` — React state → ref sync effect causes stale reads during rapid delta events.**

```typescript
// THIS was the bug:
useEffect(() => {
  messagesRef.current = messages;
}, [messages]);
```

The event processing flow:
1. `handleGatewayEvent` runs synchronously for each WebSocket event
2. It reads `messagesRef.current`, applies the event (e.g. appends a delta chunk), writes the result back to the ref, then calls `setMessages(nextMessages)`
3. React batches the state update. Before the re-render commits, the `messages` prop still holds the OLD array
4. The `useEffect` fires, and resets `messagesRef.current = messages` — back to the OLD array

When a second `message.delta` event arrives before React finishes re-rendering, it reads the stale `messagesRef.current` — a state where the first delta was never applied. Both deltas are applied to the same starting point, and one chunk silently overwrites the other.

**Evidence from debug logging:**

```
[DIAG] appendAssistantDelta: existingLen=26 chunkLen=2 newLen=28 chunkStart="后续"
[DIAG] appendAssistantDelta: existingLen=26 chunkLen=2 newLen=28 chunkStart="文本"
```

Both `"后续"` and `"文本"` were appended from the same `existingLen=26` — the second delta should have seen `existingLen=28` if the ref was current. The `useEffect` reset it back to 26 between the two events.

### Fix

**File:** `src/renderer/src/screens/Chat/hooks/useDashboardChatTransport.ts`

**Before:**
```typescript
useEffect(() => {
  messagesRef.current = messages;
}, [messages]);
```

**After:**
```typescript
useEffect(() => {
  // Only sync from React state to ref when the state has MORE messages
  // than the ref (e.g. user sent a message via pushUser, which only
  // calls setMessages and not the ref). When the ref is ahead (during
  // rapid delta events), do NOT sync — that would reset the ref to a
  // stale state.
  if (messages.length > messagesRef.current.length) {
[...]

PR #780, opened 2026-06-26
fix(stream): resolve message text truncation in streaming responses

SpiritLiberation:
Bug: Message Text Truncation in Streaming Responses https://github.com/fathah/hermes-desktop/issues/757


laofoot:
Confirmed — this is the right fix, and it's the most fundamental of the truncation reports (the mid-stream chunk drop, which matches the CJK examples in #757 better than the completion-suffix path).

For what it's worth: we've been running this exact length-guard in production for a couple of weeks. I verified it on real Chromium — without the guard, a delta arriving between an intermediate same-length commit and the next render reads the stale `existingLen` and one chunk silently overwrites the other; with `if (messages.length > messagesRef.current.length)` the ref stays authoritative through rapid deltas while still syncing on `pushUser`/resume. Holds up.

One suggestion to get CI green and lock it in — the race is subtle (it needs an **intermediate stale, same-length** commit, which a naive test misses), so a regression test is worth adding. This drops into the existing `useDashboardChatTransport.test.tsx` (reuses its `Harness` + `dashboardMock`):

```tsx
it("preserves streamed deltas when an intermediate stale commit resets the messages ref (#757)", async () => {
  dashboardMock.request.mockImplementation(async (method: string) => {
    if (method === "session.create") return { session_id: "live-1", stored_session_id: "s1" };
    if (method === "model.options") return { model: "m", provider: "p", providers: [] };
    return {};
  });
  const api: HarnessApi = {};
  render(<Harness api={api} />);
  await act(async () => { await api.send?.("turn"); });

  const delta = (text: string): void => {
    dashboardMock.onEvent?.({ payload: { text }, session_id: "live-1", type: "message.delta" });
  };
  await act(async () => { delta("Hello"); });
  await act(async () => { delta("World"); });

  // Intermediate stale, SAME-LENGTH commit: the unguarded
  // `messagesRef.current = messages` would reset the ref backward to this,
  // so the next delta appends onto stale content and drops "World" (#757).
  await act(async () => {
    api.setMessages?.((prev) =>
      prev.map((m) => (m.role === "agent" ? { ...m, content: "Hello" } : m)));
  });
  await act(async () => { delta("!"); });

  const streamed = api.messages
    ?.filter((m): m is ChatBubbleMessage => "content" in m && m.role === "agent")
    .at(-1);
  expect(streamed?.content).toBe("HelloWorld!");
});
```

Without the guard this asserts `"Hello!"` (red); with it, `"HelloWorld!"` (green). Happy to push it as a commit to your branch if that's easier — your call, it's your PR. 🙏

SpiritLiberation:
> Confirmed — this is the right fix, and it's the most fundamental of the truncation reports (the mid-stream chunk drop, which matches the CJK examples in #757 better than the completion-suffix path).已确认的&mdash；这是正确的修复，它是截断报告中最基本的（中流块丢失，它与<；b0></b0>；中的CJK示例匹配得比补全后缀路径更好）。
> 
> For what it's worth: we've been running this exact length-guard in production for a couple of weeks. I verified it on real Chromium — without the guard, a delta arriving between an intermediate same-length commit and the next render reads the stale `existingLen` and one chunk silently overwrites the other; with `if (messages.length > messagesRef.current.length)` the ref stays authoritative through rapid deltas while still syncing on `pushUser`/resume. Holds up.为了它的价值：我们已经在生产中运行这个精确的长度守卫几个星期了。我在真实的Chromium上验证了这一点，在没有保护的情况下，在中间相同长度的提交和下一个渲染之间到达的delta读取了过期的数据，一个块静默地覆盖了另一个块；与ref保持权威通过快速三角洲，同时仍然同步/恢复。持有。
> 
> One suggestion to get CI green and lock it in — the race is subtle (it needs an **intermediate stale, same-length** commit, which a naive test misses), so a regression test is worth adding. This drops into the existing `useDashboardChatTransport.test.tsx` (reuses its `Harness` + `dashboardMock`):一个建议是将CI设置为绿色，并将其锁定在竞争中，这是很微妙的（它需要一个中间陈旧的、相同长度的提交，这是幼稚的测试错过的），因此值得添加一个回归测试。这落入现有的（重用其）：
> 
> ```tsx
> it("preserves streamed deltas when an intermediate stale commit resets the messages ref (#757)", async () => {
>   dashboardMock.request.mockImplementation(async (method: string) => {
>     if (method === "session.create") return { session_id: "live-1", stored_session_id: "s1" };
>     if (method === "model.options") return { model: "m", provider: "p", providers: [] };
>     return {};
>   });
>   const api: HarnessApi = {};
>   render(<Harness api={api} />);
>   await act(async () => { await api.send?.("turn"); });
> 
>   const delta = (text: string): void => {
>     dashboardMock.onEvent?.({ payload: { text }, session_id: "live-1", type: "message.delta" });
>   };
>   await act(async () => { delta("Hello"); });
>   await act(async () => { delta("World"); });
> 
>   // Intermediate stale, SAME-LENGTH commit: the unguarded
>   // `messagesRef.current = messages` would reset the ref backward to this,
>   // so the next delta appends onto stale content and drops "World" (#757).
>   await act(async () => {
>     api.setMessages?.((prev) =>
>       prev.map((m) => (m.role === "agent" ? { ...m, content: "Hello" } : m)));
>   });
>   await act(async () => { delta("!"); });
> 
>   const streamed = api.messages
>     ?.filter((m): m is ChatBubbleMessage => "content" in m && m.role === "agent")
>     .at(-1);
>   expect(streamed?.content).toBe("HelloWorld!");
> });
> ```
> 
> Without the guard this asserts `"Hello!"` (red); with it, `"HelloWorld!"` (green). Happy to push it as a commit to your branch if that's easier — your call, it's your PR. 🙏如果没有卫兵，这条路是肯定的（红色）；用它，（绿色）。很高兴将其作为提交推送到您的分支，如果这更容易&mdash；您的呼叫，这是您的PR.🙏

Of course, I'd be happy to. Your commit is welcome.

PR #800, opened 2026-07-02
fix(chat): read gateway usage keys so context gauge shows exact tokens

jaeeyoungkim:
## What does this PR do?

The context gauge at the bottom-right of the chat never appears for **local /
dashboard-transport sessions running against the built-in Hermes gateway**,
regardless of provider.

### Root cause

`useDashboardChatTransport`'s `usageFromPayload` only read
`usage.prompt_tokens` / `usage.promptTokens`. But the Hermes gateway
(`_get_usage` in `tui_gateway/server.py`) emits usage on `message.complete`
with **snake-case, non-`_tokens`** keys:

```
model, input, output, reasoning, prompt, completion, total, calls
# + context_used, context_max, context_percent  (when the context compressor is active)
```

So `prompt_tokens`/`promptTokens` are both absent → `promptTokens = 0` →
`contextTokens = undefined` → `ChatInput`'s `contextUsage.used > 0` guard is
never satisfied → the gauge silently does not render.

(The legacy IPC path in `useChatIPC` works because the Electron main process
already normalizes `u.input ?? u.prompt ?? u.prompt_tokens` before forwarding.
The dashboard transport parses `event.payload.usage` directly and skipped that
normalization.)

### Fix

Read every spelling in `usageFromPayload` — gateway snake-case
(`input`/`output`/`prompt`/`completion`/`total`), legacy OpenAI `*_tokens`, and
camelCase — and prefer `context_used` for the gauge value.

`context_used` is the compressor's live context-window occupancy
(`last_prompt_tokens`): a **snapshot of the current turn**, which is exactly
what the gauge documents it wants (see the `contextTokens` doc comment in
`types.ts` — "current context-window occupancy, NOT summed across turns").

### Relationship to #789

This is an alternative to #789. #789 falls back to a `chars / 4` **estimate**
when tokens are missing. This PR instead reads the **exact** counts the gateway
already reports, so no estimation is needed for the common case (built-in
gateway, any provider). The two are complementary — an estimate could still be
layered on for truly usage-less third-party providers — but the gauge-blank
symptom that #789 targets is, for the built-in gateway, a key-mismatch bug that
this PR fixes at the source.

## How to Test

1. Run a local session against the built-in gateway with any provider
   (e.g. `anthropic` / `claude-opus-4-8`).
2. Send a message — the context gauge now appears bottom-right and tracks the
   real context-window occupancy reported by the gateway.
3. Legacy OpenAI-style providers that return `prompt_tokens` continue to work
   unchanged.

`npm test` (adds `usageFromPayload.test.ts`, 7 cases covering all key
spellings + the no-usable-data → null path), `npm run typecheck`, and
`npm run lint` all pass.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

### DIFF
--- a/src/renderer/src/screens/Chat/Chat.tsx
+++ b/src/renderer/src/screens/Chat/Chat.tsx
@@ -901,10 +901,15 @@ function Chat({
   );
 
   // Context-gauge data: the latest turn's prompt tokens vs the model's window.
+  // Denominator priority: gateway-reported context_max (authoritative — knows
+  // the actual model config) > provider /models catalogue > static heuristic.
   const contextUsage: ContextUsage | null = usage?.contextTokens
     ? {
         used: usage.contextTokens,
-        window: realContextWindow ?? contextWindowForModel(chatCurrentModel),
+        window:
+          usage.contextWindowTokens ??
+          realContextWindow ??
+          contextWindowForModel(chatCurrentModel),
         cacheReadTokens: usage.cacheReadTokens,
         cacheWriteTokens: usage.cacheWriteTokens,
       }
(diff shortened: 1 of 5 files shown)


jaeeyoungkim:
image

It has been confirmed that the session context window is appearing correctly.

PR #805, opened 2026-07-02
Fix duplicate slash alias crash

giattijunior:
Fix duplicate slash alias crash when agent catalog includes `/compact`

## Summary

The renderer crashed when the backend command catalog exposed `/compact` both as:

- a standalone command in `pairs`
- an alias mapping `compact -> compress` in `canon`

`createSlashCatalog()` registered the command first, then `registerAlias()` threw `Duplicate slash command alias: /compact`, which broke the Chat screen on connection.

## Fix

Ignore duplicate alias registrations when the canonical command is already present instead of throwing.

## Validation

- Added a regression test for `pairs: /compact, /compress` plus `canon: compact -> compress`
- Ran:

```bash
npm test -- src/renderer/src/screens/Chat/slash/commandCatalog.test.ts
```

Result: `1 passed`, `4 passed`

### DIFF
--- a/src/renderer/src/screens/Chat/slash/commandCatalog.test.ts
+++ b/src/renderer/src/screens/Chat/slash/commandCatalog.test.ts
@@ -53,4 +53,22 @@ describe("slash command catalog", () => {
       }),
     ).toThrow("Duplicate slash command: /inspect");
   });
+
+  it("ignores alias entries that duplicate an already-registered command", () => {
+    const upstream = agentCommandsFromCatalog({
+      pairs: [
+        ["/compact", "Compact and summarize the conversation"],
+        ["/compress", "Compress conversation with optional focus topic"],
+      ],
+      canon: { compact: "compress" },
+    });
+
+    const catalog = createSlashCatalog({
+      agentCommands: upstream.commands,
+      aliases: upstream.aliases,
+    });
+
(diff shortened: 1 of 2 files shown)


fathah:
Thanks for tracking this down and for the regression test — the diagnosis of the `/compact` pairs-vs-canon collision was exactly right.

This has since been fixed on `main` by #808, which resolves the same collision one layer earlier: `agentCommandsFromCatalog` now drops a `canon` alias whose name is already a first-class `pairs` command, so the reconciled catalog is self-consistent before it ever reaches `createSlashCatalog`. Your repro scenario (`pairs: /compact, /compress` + `canon: compact → compress`) is covered by a regression test there.

We deliberately kept the fix out of `registerAlias`: its throw-on-duplicate is intentional validation for genuine desktop-authoring conflicts (two in-repo commands claiming the same name), and silently returning would mask those at startup. The design rule we've settled on — also applied in #818 for the related backend-command-vs-desktop-alias case — is that backend catalog data is untrusted runtime input and gets sanitized at the reconciliation layer, while `createSlashCatalog`'s throw stays strict for in-repo authoring mistakes.

Closing as superseded by #808. Thanks again for the report and the clean repro — it helped confirm the fix covers this exact case.

Issue #748, opened 2026-06-21
Bug Report: Chat Input Lag with Long Conversations，Hermes Desktop v0.6.35（Fixed, please refer to it.）


SpiritLiberation:
## Bug Report: Chat Input Lag with Long Conversations

**Environment**: Hermes Desktop (fathah/hermes-desktop), Electron 39, Linux (AOSC / Wayland)

### Problem
When a conversation contains many messages, typing in the chat input (`<textarea>`) becomes progressively slower. Each keystroke takes up to **2.6 seconds** to respond. New/empty sessions are instant. GPU flags (`--disable-gpu`, `--disable-gpu-compositing`, `--in-process-gpu`) have no effect. DevTools shows empty JS flamegraph during the lag — bottleneck is in Chromium's C++ layout engine.

**Root cause**: Browser style/layout recalculation traverses the **entire DOM tree** on every keystroke. With many messages rendered, this becomes O(n) in DOM size.

### Fix (3 CSS changes in `src/renderer/src/assets/main.css`)

**1. Convert `.chat-messages` from flex to block flow**

`content-visibility: auto` conflicts with flex item layout — `display: flex; flex-direction: column` on the scroll container causes incorrect `scrollHeight` when children have `content-visibility: auto`, breaking scrollbar and auto-scroll.

```css
.chat-messages {
  flex: 1;
  overflow-y: auto;
  padding: 24px;
  /* REMOVED: display: flex; flex-direction: column; gap: 16px; */
}
```

**2. Add `content-visibility: auto` to each message**

Skipping off-screen message layout eliminates the per-keystroke O(n) layout cost:

```css
.chat-message {
  display: flex;
  gap: 12px;
  width: 100%;          /* block layout needs explicit width */
  max-width: 85%;
  animation: messageIn 0.2s ease;
  margin-bottom: 16px;  /* replaces removed flex gap */
  content-visibility: auto;
  contain-intrinsic-size: auto 3rem;
}
```

**3. Fix user message alignment**

`.chat-message-user` used `align-self: flex-end` which only works in flex layout — replaced with:

```css
.chat-message-user {
  flex-direction: row-reverse;
  margin-left: auto;
}
```

### Notes for Upstream
- The original `display: flex; flex-direction: column` on `.chat-messages` was needed for `align-self` on children — but this prevents using `content-visibility: auto` for performance.
[...]

PR #807, opened 2026-07-03
fix(chat): render image lightbox via portal so paint containment can't clip it

laofoot:
**Bug:** in a long conversation, clicking a chat image (or an image attachment chip) opens the zoom lightbox **clipped to the message row's box** instead of fullscreen.

**Root cause:** `.chat-message` rows use `content-visibility: auto` (the #748 input-lag fix), which implies **paint containment** — anything rendered inside the row can't paint outside its box. The lightbox overlay was rendered inline inside the row, so containment clips it. Same mechanism the CSS already works around for the hover timestamp (`main.css` comment: "the paint containment that content-visibility implies does not clip it") — the lightbox path was missed.

**Fix:** render both lightboxes (`MediaImage`, `AttachmentChip`) through `createPortal(…, document.body)`, escaping the contained subtree. Overlay behavior (click-outside to close, save action, stopPropagation on the action bar) unchanged.

Typecheck + full vitest suite pass. Repro: any conversation with `content-visibility` active → click an image → before: clipped strip; after: fullscreen overlay.


laofoot:
Addressed the Escape-key gap from the review: added a shared `useLightboxClose` hook (document keydown bound only while the overlay is open) and wired it into both lightboxes — `de08184`. Typecheck + component tests pass.

Issue #813, opened 2026-07-03
Startup failed: Duplicate slash command alias: /compact


aEgoist:
<img width="1086" height="843" alt="Image" src="https://github.com/user-attachments/assets/e1d44c2a-d55c-472f-8c99-d2e22f5202c1" />

uptimized:
I am also getting this after updating the engine (app was already up to date). After restart this pops up within few seconds and if you click the button it just reappears instantly.

AmirF194:
The `/compact` variant of this crash was fixed in #808 (closing #802 / #804) and shipped in **v0.7.3**, which released shortly after this issue was filed. The reconciler now drops a `canon` alias whose name is already a first-class `pairs` command, so `/compact` being both a standalone command and an alias of `/compress` no longer throws.

If you're still hitting `Duplicate slash command alias: /compact` on **v0.7.3 or later**, please reopen with the engine (hermes-agent) version, that would mean a different collision path and we'd want the exact catalog.

While verifying the fix I did find a *separate* residual crash in the same class, a backend command whose name equals a desktop **alias** (rather than a command name) still throws on startup. Filed as #817 with a fix in #818.

fathah:
Fixed

Issue #804, opened 2026-07-02
Duplicate slash command alias: /compact crashes app on agent connect


real-habibul:
## Bug Description

The desktop app throws `Error: Duplicate slash command alias: /compact` when connecting to the Hermes Agent backend.

## Root Cause

The bug is in `createSlashCatalog()` in the renderer. When the agent backend sends its command catalog:

1. The agent's Python `COMMANDS\dict (hermes_cli/commands.py) includes aliases as flat entries:
   - `/compact` → `Compact and summarize... (alias for /compress)`
   - `/compress` → `Compress conversation context...`

2. The agent also sends `compact` as an alias mapping for `/compress`

3. `createSlashCatalog()` receives both:
   - `agentCommands` array includes `/compact` as a standalone command (from the catalog pairs)
   - `aliases` object includes `compact → compress`

4. `registerAlias()` (line ~129240 in compiled renderer) hits:
   ```js
   if (byName.has(aliasKey) || aliases.has(aliasKey)) {
     throw new Error('Duplicate slash command alias: /' + aliasKey);
   }
   ```

## Affected Code

`out/renderer/assets/index-DvMTJoSZ.js` (or equivalent source):

```js
// registerAlias function
function registerAlias(byName, aliases, alias, target) {
  const aliasKey = normalizeName(alias);
  if (byName.has(aliasKey) || aliases.has(aliasKey)) {
    throw new Error(`Duplicate slash command alias: /${aliasKey}`); // <-- crashes here
  }
}
```

## Suggested Fix

Change `registerAlias` to skip duplicates gracefully instead of throwing:

```js
if (byName.has(aliasKey) || aliases.has(aliasKey)) {
  return; // skip — command already registered
}
```

Alternatively, deduplicate in `agentCommandsFromCatalog()` by filtering out catalog entries whose normalized name also appears in the agent's alias mapping.

## Environment

- Hermes Desktop: v0.7.2 (latest)
- Hermes Agent: v0.18.0
- OS: Windows 11
- Provider: Xiaomi MiMo

## Steps to Reproduce

1. Open Hermes Desktop
2. Connect to a running Hermes Agent backend
3. The error appears immediately on connection

## Workaround

Hot-patched `app.asar` to replace the throw with a return statement.

Issue #802, opened 2026-07-02
Slash-command catalog merge throws on any name collision, crashes app startup


Dropicx:
## Description

The desktop app's slash-command catalog merge logic (`createSlashCatalog` / `registerAlias`, in the bundled renderer JS) throws `Duplicate slash command alias: /<name>` whenever a name appears simultaneously in the backend's `pairs` list and its `canon` alias map. This crashes the app during the initial "getting started" flow, since the catalog is built as soon as the agent command catalog is fetched.

## Steps to reproduce

1. Run current `hermes-agent` main (which includes [PR #57029](https://github.com/NousResearch/hermes-agent/pull/57029), merged 2026-07-02), pairing it with Hermes Desktop v0.7.2.
2. Launch Hermes Desktop and go through onboarding / open a chat.
3. The command catalog fetch resolves, `createSlashCatalog` runs, and throws.

## Root cause

- `hermes-agent`'s `tui_gateway/server.py` `commands.catalog` RPC returns `pairs` (flat name/description list) and a separate `canon` alias map.
- A pre-existing, TUI-only entry in `_TUI_EXTRA` — `("/compact", "Toggle compact display mode", "TUI")` — lands in `pairs` as a standalone command.
- PR #57029 added `/compact` as a first-class alias of `/compress` on the backend (`aliases=("compact",)` in `hermes_cli/commands.py`), which populates `canon["/compact"] = "/compress"`.
- The desktop's `agentCommandsFromCatalog()` builds real commands from every `pairs` entry, then separately tries to register every `canon` alias — but only validates that the alias's *target* isn't already claimed, not that the *alias name itself* isn't already claimed by a `pairs` entry. Result: `/compact` gets registered twice (once as a standalone command from `pairs`, once as an alias from `canon`), and the second registration throws.

## Suggested fix

In the merge logic, either:
- Skip/log instead of throwing when a `canon` alias name collides with an existing `pairs`-derived command name, preferring one source deterministically, or
- Exclude any `canon` alias whose key already exists in the `pairs`-derived command set before attempting registration.

For reference, NousResearch's newer (unbundled) desktop source no longer uses this throw-on-collision merge approach at all — see `apps/desktop/src/lib/desktop-slash-commands.ts` in `NousResearch/hermes-agent`, which hardcodes commands like `/compact` as known terminal-only entries instead of dynamically reconciling two catalogs.

## Environment
- Hermes Desktop v0.7.2 (Linux, .deb install)
[...]

PR #818, opened 2026-07-03
fix(chat): drop backend commands that collide with a desktop alias

### LINKED ISSUE #817: Backend command whose name equals a desktop alias crashes startup (Duplicate slash command alias)
### Summary

The slash-command catalog reconciliation checks backend agent commands against desktop command **names** but not desktop **aliases**. A backend command whose name equals a desktop *alias* survives the merge, gets registered as an agent command, and then makes the desktop command's own alias registration throw `Duplicate slash command alias`, crashing the renderer on startup.

This is the same startup-crash class as #802 / #804 / #813; those covered the `pairs`-vs-`canon` (`/compact`) path, but the desktop-alias path is still open.

### Root cause

In `Chat.tsx`'s `slashCatalog` useMemo, three filters exclude collisions with desktop entries, but they are asymmetric:

- `desktopTargetAliases` filter → excludes desktop **names + aliases** ✅
- `aliases` map filter → excludes desktop **names + aliases** ✅
- `agentCommands` filter → excludes desktop **names only** ❌

So a backend command named like a desktop alias is not dropped. Today the only desktop alias is `commands` (declared by `help`), so a backend `/commands` command triggers it:

```
Error: Duplicate slash command alias: /commands
```

`createSlashCatalog` registers the backend `commands` command, then `help` tries to register its `commands` alias → duplicate → throw → app fails to start.

### Reproduction

Driving the exact reconciliation with a backend catalog that exposes `/commands`:

```ts
reconcileSlashCatalog({
  catalog: { pairs: [["/commands", "List agent commands"]], canon: {} },
  desktopCommands: [help /* aliases: ["commands"] */],
  fallbackAgentCommands: [],
});
// throws: Duplicate slash command alias: /commands
```

### Expected

The backend catalog is untrusted runtime data — a collision there must never crash the app. A backend command that collides with a desktop name **or** alias should be dropped, and desktop authoring should win deterministically (`/commands` → the desktop `help` command). `createSlashCatalog`'s throw is still valuable for genuine desktop-vs-desktop authoring conflicts and should stay.

### Fix

PR incoming: add the missing desktop-alias check to the agent-command filter (extracting the reconciliation into a pure, unit-tested `reconcileSlashCatalog` helper).

AmirF194:
Fixes #817

### Problem

The slash-catalog reconciliation filtered backend agent commands against desktop command **names** but not desktop **aliases** — an asymmetry with its two sibling filters (desktop-target aliases and the alias map both already exclude desktop aliases).

A backend command whose name equals a desktop alias therefore survived into the merged catalog and was registered as an agent command; when the desktop command later registered that same name as its alias, `createSlashCatalog` threw `Duplicate slash command alias` and the app crashed on startup.

Today the only trigger is a backend `/commands` command, since desktop `help` aliases `commands` — but it's the same startup-crash class as the `/compact` collisions in #802 / #804 / #813.

### Fix

- Extract the reconciliation out of `Chat.tsx`'s `useMemo` into a pure, exported `reconcileSlashCatalog` so the merge is unit-testable (mirrors the tested-helper approach from #808).
- In that helper, exclude any backend command whose name matches a desktop **name OR alias**.

Desktop authoring still wins deterministically (`/commands` resolves to the desktop `help` command), and `createSlashCatalog`'s throw-on-collision guard for genuine desktop-vs-desktop authoring conflicts is left intact.

### Tests

Added `reconcileSlashCatalog` specs in `commandCatalog.test.ts`:
- backend `/commands` colliding with `help`'s `commands` alias no longer throws, and resolves to the desktop command
- desktop wins over a same-named backend command
- non-colliding backend commands are still exposed

`vitest run src/renderer/src/screens/Chat/slash/` → 23 passed. `tsc --noEmit` clean. `lat check` passes.


fathah (approved):
Reviewed and verified locally — approving.

**Confirmed the bug is live on `main`:** replicated the current `Chat.tsx` merge logic with a backend catalog exposing `/commands` against the desktop `help` command (which aliases `commands`) — it throws `Duplicate slash command alias: /commands` exactly as described in #813. So this is not covered by #808, which fixed the backend-internal pairs-vs-canon collision; this PR closes the remaining backend-command-vs-desktop-alias axis.

**Verification performed:**
- The `reconcileSlashCatalog` extraction is a faithful move of the `Chat.tsx` `useMemo` logic — the `desktopTargetAliases` and alias-map filters are byte-identical; the only behavioral change is the added `!desktopAliases.has(command.name)` guard, which is the fix.
- Mutation-tested the regression test: removing just the alias check makes it fail with the exact production error; with the fix, all 23 slash tests pass.
- Mixed-case squatting is covered — `normalizeName` lowercases backend names before the filter.
- The `catalog: null` fallback path now also gets alias protection — a strict improvement over main.
- `createSlashCatalog`'s throw-on-collision stays intact for genuine desktop-vs-desktop authoring conflicts, consistent with the layering established in #808.
- `tsc` clean (the one `i18n/index.ts` error is pre-existing on main), eslint clean on touched files (the 3 `exhaustive-deps` warnings in `Chat.tsx` are pre-existing, outside this diff), `lat check` passes, and the `lat.md` update accurately documents the new reconciliation layer.

Nice work keeping the fix at the reconciliation boundary and making the merge unit-testable.

PR #838, opened 2026-07-09
Fix setup credentials for active profile

lucyonegit:
When a named profile needs first-run setup, saving a provider key previously wrote to the default workspace. Setup now passes the profile returned by the install check to both credential and model-config writes, including custom endpoint credentials.

Only the latest completed install check can select the setup profile, preserving the startup generation guard when checks overlap.

Validation: 2,175 tests pass (3 skipped), including setup coverage for named and default profiles; typecheck, lint, and lat check pass. Integrated current main and checked the installer’s profile-aware readiness contract against the existing profile APIs.

### DIFF
--- a/lat.md/provider-setup.md
+++ b/lat.md/provider-setup.md
@@ -4,6 +4,8 @@ The first-run screen where the user picks an AI provider and enters credentials 
 
 The provider list is data-driven from `PROVIDERS.setup` in [[src/renderer/src/constants.ts]]. Each entry carries an `envKey`, `configProvider`, `baseUrl`, and `needsKey`; selecting a card drives which form fields show (API key, or the Local server/base-URL flow).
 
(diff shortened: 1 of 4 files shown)


fathah (commented):
Thanks @lucyonegit for catching the mismatch between profile-aware install checks and default-profile setup writes. I reviewed the credential/model IPC signatures, installer behavior, and Greptile’s review, and integrated current main.

I also moved setup-profile state updates behind the startup generation guard, so a stale install check cannot configure the wrong profile. Named and default setup paths have regression coverage. All 2,175 local tests pass (3 skipped), with typecheck, lint, and lat check passing. Waiting for refreshed GitHub checks before merging.

fathah (approved):
Approved the final head: active-profile credential and model writes are consistent, the startup generation guard is preserved, all 2,175 local tests and checks pass, GitHub CI passes, and refreshed Greptile is 5/5. Thanks for the fix.

PR #853, opened 2026-07-15
feat: support OAuth-authenticated remote dashboards

itsJai42:
## Summary

- detect OAuth-authenticated remote dashboards through public status metadata
- keep OAuth cookies inside a persistent Electron session partition
- mint single-use WebSocket tickets without exposing credentials to renderer code
- relay insecure non-loopback dashboard WebSockets through a one-shot capability URL
- suppress preserved bearer keys whenever a Remote connection resolves to OAuth
- add browser sign-in, sign-out, and bounded session-state UI
- route direct-Remote session REST calls through OAuth cookies
- keep startup status probes public and ignore stale token state after OAuth selection

## Why

Hermes Desktop supports token-authenticated remote dashboards, but Hermes One browser OAuth was not wired into the Desktop connection flow. OAuth remotes therefore could be reachable while still being rejected or probed with stale token state.

Related context: NousResearch/hermes-agent#38873. That issue is closed, but documents the broader remote-mode state and fallback failure that led to this Desktop-side investigation.

## Hermes Agent compatibility

- Existing token-authenticated dashboards remain supported.
- OAuth mode requires an Agent dashboard that exposes public authentication metadata, browser login/session cookies, and single-use WebSocket ticket minting.
- Older Agents without OAuth metadata continue through token mode; Desktop does not synthesize OAuth support or fall back to local Agent state.
- OAuth access and refresh cookies remain Electron-owned and are never returned through the preload bridge.

## Verification

- `npm ci` — Node 22
- `npm run typecheck` — passed
- `npm test` — 1,737 passed, 3 skipped
- `npm run build` — passed
- changed-file ESLint — 0 errors; 3 pre-existing React hook warnings in `Chat.tsx`
- full `npm run lint` — same 9 errors already present on `main`; warnings decrease from 127 to 112
- `npm run dev` — Electron main/preload built, renderer served, Electron app started
- `lat check` — passed

## Follow-up stack

This is 1/9. Later fork branches add the authenticated management client, profiles/gateway/tools, providers, diagnostics, attachments, discovery, memory, and Kanban. They will be submitted sequentially after each predecessor merges so every upstream PR remains focused.


itsJai42:
Addressed the latest stale-bearer finding in `7cb0a47`.

`getRemoteAuthHeader()` now treats a stored Remote API key as inactive whenever `remoteAuthMode === "oauth"`, so fallback and auxiliary main-process requests cannot attach credentials from the previous token principal. Token-mode and SSH headers remain unchanged.

Added a focused regression test that first reproduces `Authorization: Bearer stale-token` in OAuth mode, then verifies suppression plus token-mode preservation. Current PR verification: 1,737 tests passed, 3 skipped; typechecks, production build, changed-file ESLint, and `lat check` passed.

PR #870, opened 2026-07-22
Fix dev blank screen and remote cron/dashboard API routing

dataguybobby:
### Dev blank screen (`npm run dev`)
Running Hermes Desktop locally could show a white window because the renderer failed during Vite module evaluation:

- `@react-three/drei` (~3.5MB) pre-bundle races with Electron startup — Electron loads before the dep optimizer finishes.
- Corrupt `.vite` cache (`.map` present without `.js`) after interrupted re-optimizes.
- Stale optimize-dep URLs return 504 on first load.
- Opening the Office tab discovered `SkeletonUtils.js`, triggering mid-session re-optimize that invalidated drei mid-request.

Chat startup also eagerly imported the Office 3D stack, coupling startup to drei.

### Remote cron and gateway readiness
Remote/dashboard mode had two related failures:

- Cron HTTP calls hit gateway `/api/jobs`, which returns **405** when the active transport is the dashboard (`/api/cron/jobs`).
- `isApiServerReady` treated dashboard SPA HTML responses on `/health` as a healthy gateway, causing false positives during remote connection setup.

### Dev loading
- Lazy-load the Office tab in `Layout` and Office3D in `Office` so Chat mounts without drei.
- Pre-bundle `@react-three/drei`, fiber, three, troika, and `SkeletonUtils` via `optimizeDeps.include`.
- `scripts/ensure-vite-deps.mjs` clears corrupt drei cache entries before `npm run dev`.
- `scripts/vite-wait-for-drei.ts` blocks the first HTML response until drei pre-bundle completes.
- Bounded dev recovery: `vite:preloadError` reload in the renderer; main process reloads once if `#root` is still empty after 4s.
- Added `npm run dev:clean` for stale cache resets.

### Remote API routing
- `listCronJobs` and mutations prefer dashboard `/api/cron/jobs`, falling back to gateway `/api/jobs` on 404/405.
- Map dashboard `/trigger` to gateway `/run` on fallback.
- `isApiServerReady` rejects `text/html` `/health` responses.
- Remote connection tests prefer `/api/status` when chat transport is not legacy.

## Test plan
- [x] `npm test` — 1822 passed
- [ ] `npm run dev` — Chat tab loads on first launch (no blank screen)
- [ ] Open Office tab — 3D scene loads without dynamic import failures
- [ ] Remote mode — cron jobs list/create/delete against dashboard URL
- [ ] Remote mode — legacy gateway-only host still works via `/api/jobs` fallback
- [ ] Remote connection test — dashboard HTML `/health` no longer marks gateway ready


fathah (changes requested):
Thanks @dataguybobby for the detailed reproductions and the dev-startup work. I checked out the branch and attempted integration with current main. It has eight conflicts, mostly around session-title persistence already landed in #865; retaining that older implementation would lose the profile, provenance, and rename-serialization fixes now on main.

I am holding this PR because the new dashboard cron path still has two functional gaps:

- `remoteFetchCron` calls the gateway `remoteFetch` helper. In browser-authenticated Remote mode, `getRemoteAuthHeader` intentionally returns no header and plain Node fetch cannot use the desktop’s isolated OAuth session. Changing the route alone therefore leaves authenticated dashboard cron requests failing. Use the existing dashboard request abstraction, which handles OAuth sessions and token authentication.
- The new dashboard calls drop the supplied profile. `createCronJob(..., profile)` POSTs to `/api/cron/jobs` without its profile query, and list/mutation paths do the same. The unified dashboard owns multiple profiles; its create route defaults to the default workspace, so creating a named-profile schedule can modify the wrong workspace. Carry the selected profile through every dashboard request and test it.

Greptile’s 5/5 review missed these transport boundaries. Please separate the remaining dev-startup/API changes from the superseded rename code and cover cookie-authenticated Remote plus named-profile dashboard operations before merging. No code from this branch has been merged.

Issue #749, opened 2026-06-22
Hermes Desktop Bug Fix Summary（Fixed, please refer to it.）


SpiritLiberation:
### Issue 1: AI reply bubble text truncated (missing trailing content)

**Root cause:** In `src/main/tui-gateway-stream.ts`, the `gatewayCompletionSuffix` function compares streamed text against the final `message.complete` payload using strict `finalText.startsWith(streamedText)`. Any whitespace discrepancy between the accumulated stream chunks and the final text (e.g. trailing space, interior whitespace shifts) causes `startsWith` to return false, which makes the function return an empty string — silently discarding the remaining content that was never streamed.

**Fix:** Added multi-tier fuzzy matching — `trimEnd()`-based `startsWith` fallback, `includes`-based substring detection with `slice`, and whitespace-normalized equality check. Ensures trailing content is preserved even when minor formatting differences exist between the stream and final text.

### Issue 2: DB refresh never replaces incomplete streamed content with complete DB content

**Root cause:** In `src/renderer/src/screens/Chat/sessionHistory.ts`, `mergeDbMetadataIntoStreamed` (used during reconciliation at stream end) only merges attachment metadata from the DB row — it never overwrites the `content` field. If the streamed message is incomplete due to issue 1 or a race condition, the truncated content is permanently kept.

**Fix:** When DB content is longer than the streamed version, strip structural wrapping (legacy `<file>` tags, `[screenshot]`/`[image]` markers) from both sides and compare. Only replace with the DB content when it contains substantive text the stream missed. Skips cases where the stream already resolved the content correctly and the DB only added redundant markup.

### Issue 3: Session resume scrolls to wrong position (short of the latest message)

**Root cause:** In `src/renderer/src/screens/Chat/hooks/useChatScroll.ts`, `scrollIntoView({ behavior: "smooth" })` fails when loading a session from history with many messages. The CSS rule `content-visibility: auto` on `.chat-message` causes off-screen messages to occupy only 3rem (`contain-intrinsic-size: auto 3rem`) as a placeholder. The browser underestimates the total `scrollHeight`, `scrollIntoView` lands at the wrong position, and the bottom keeps shifting down as the user scrolls and real heights are computed.

**Fix (three-part):**
1. Replaced `scrollIntoView` with `container.scrollTop = container.scrollHeight` for direct programmatic control
2. Added a `ResizeObserver` on the container that force-scrolls to bottom each time `content-visibility: auto` reveals another batch of true message heights
[...]

PR #896, opened 2026-08-07
fix(chat): merge DB content into a streamed message when it is more complete

AmirF194:
## Root cause

`reconcileStreamedWithDb` matches a streamed message to its `state.db` row by the first 200 normalized characters (`reconciliationKey`), so a message the gateway streamed short (a dropped chunk, or the still-open `gatewayCompletionSuffix` truncation) still matches its DB row once the stream ends. `mergeDbMetadataIntoStreamed` then merges the DB row's `timestamp` and `attachments` into the streamed copy, but never `content`, so the truncated text ships to the UI forever even though the complete answer sits in `state.db` right next to it.

## Invariant and fix

The merge should keep the streamed message's React identity but carry the more complete of the two texts. `mergeDbMetadataIntoStreamed` now adopts the DB row's `content` only when its normalized form (`normalizeBubbleContentForMatch`) is strictly longer than the streamed copy's. Comparing normalized text, not raw length, matters: a legacy `<file>` wrapper makes a DB row raw-longer even when the modern `attachments` field already covers the same data, and that case must not overwrite already-correct streamed text (pinned by the existing "matches legacy text-file DB wrappers" test).

## Scope

This covers Bug 2 of the three bundled in #749 (the DB-reconcile content drop). Bug 1 (`gatewayCompletionSuffix`) and Bug 3 (`useChatScroll` underestimate) are untouched by this diff; #757 already tracks the streaming-side truncation cause separately.

## Verification

- New regression test (`tests/reconcile-streamed-with-db.test.ts`) fails on `main` and passes on this branch: a streamed bubble truncated past the 200-char match window is replaced with the DB's complete text.
- Added a companion test confirming the legacy-wrapper case is unaffected (content stays as the streamed, shorter value).
- Full suite: `npm test` (187 files / 1897 tests) and `npm run typecheck` both pass in a clean `node:22-slim` container.
- Not checked: the live gateway/streaming path end to end (no running Hermes backend in this environment); verification is at the pure-function level `reconcileStreamedWithDb` is designed for.

Refs #749

### DIFF
--- a/src/renderer/src/screens/Chat/sessionHistory.ts
+++ b/src/renderer/src/screens/Chat/sessionHistory.ts
@@ -296,13 +296,23 @@ function mergeDbMetadataIntoStreamed(
     !!d.attachments &&
     d.attachments.length > 0 &&
     (!s.attachments || s.attachments.length === 0);
+  // The DB row can hold more complete text than a stream cut short mid-answer.
+  // Compare normalized text, not raw length: a legacy `<file>` wrapper makes
+  // the DB row raw-longer even when `attachments` already covers it.
+  const content =
+    normalizeBubbleContentForMatch(d.content).length >
+    normalizeBubbleContentForMatch(s.content).length
+      ? d.content
+      : undefined;
   if (
     needsAttachments ||
+    content !== undefined ||
     (timestamp !== undefined && timestamp !== s.timestamp)
   ) {
     return {
       ...s,
       ...(needsAttachments ? { attachments: d.attachments } : {}),
+      ...(content !== undefined ? { content } : {}),
       ...(timestamp !== undefined ? { timestamp } : {}),
     };
   }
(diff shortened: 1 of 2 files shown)


AmirF194:
Checking in after a week, no rush. The greptile bot review left no blocking notes, and CI is clean if you want to take a look when you get a chance.

AmirF194:
Restricted content adoption to agent bubbles (pushed 4777489). I tried to reproduce the trailing `<file>` wrapper leak you flagged and couldn't: normalizeBubbleContentForMatch strips that wrapper from both sides before the length comparison, so it never reads as longer there. Scoped the fix to agent role anyway, since that's the actual intent and it closes off any other DB-side augmentation of a user row, not just this one.

AmirF194:
Closing to keep the queue clean. Happy to reopen if there's interest.

PR #904, opened 2026-08-12
Require manual approval for dangerous desktop commands

Sushanth012:
Dangerous commands now pause for an explicit decision in Dashboard and local gateway WebSocket chat. The approval card shows the command, offers the gateway's allowed choices, and requires a second confirmation before granting permanent permission.

Responses target the gateway's request ID and runtime session. The main-process bridge also binds each opaque UI request to its renderer and chat run. Expired requests, lost acknowledgments, duplicate clicks, disconnects, and cancellation cannot approve another queued command or replay the original prompt. Pending cards become unavailable when their turn ends.

The current Hermes Agent Runs approval endpoint ignores request IDs and resolves the queue head. Runs that request approval therefore stop safely with the original connection's credentials and direct users to Dashboard chat; ordinary Runs streaming remains supported. Manual Runs approval support needs an upstream request-addressed contract first.

Validated against the sibling Hermes Agent source at 503d863fcd2cbfc0be5a6d6c536fae2e98aa4204. Updated against current main while preserving connection-scoped run routing.

Validation: 2,150 tests passed, 3 skipped; TypeScript checks, ESLint, and `lat check` passed. Added transport-level regressions for request correlation, stale requests, missing IDs, disconnects, prompt acknowledgment failure, and unsupported Runs approvals.

### DIFF
--- a/lat.md/chat-commands.md
+++ b/lat.md/chat-commands.md
@@ -30,7 +30,6 @@ Because no global loading state is set, the slash branch shows its own feedback:
 
 ## Transport connection lifecycle
 
-
 Every dashboard turn first connects a JSON-RPC WebSocket to the gateway; that handshake must be time-bounded or a stalled socket wedges the whole transport with no error and no fallback (issue #718).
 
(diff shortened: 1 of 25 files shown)


fathah (commented):
Thanks @Sushanth012 for replacing silent command approval with an explicit user decision. This is an important security improvement.

I reviewed the source against current main and the sibling Hermes Agent implementation, including Greptile's 5/5 review. I found and pushed fixes for issues that review missed:

- Upstream WebSocket responses support `request_id`; FIFO-only responses could approve a different queued command after a timeout or lost acknowledgment. Responses now use the gateway-issued ID and fail closed on an unresolved acknowledgment.
- The upstream Runs approval handler ignores request IDs. Those turns now stop safely instead of approving through that endpoint or replaying via chat completions.
- Prompt acknowledgment failures and connection loss after an approval request no longer replay the prompt. Completion and cancellation invalidate pending cards, and run cleanup preserves connection scoping.

I reproduced the FIFO issue using the actual upstream resolver with synthetic queue entries: a stale response approved the next entry without an ID; the same response with the original ID resolved zero entries and left the next command untouched.

Local validation passed: 2,150 tests (3 skipped), type checking, ESLint, and `lat check`. Added regression coverage for these failure paths. Final merge is gated on the updated GitHub checks.

fathah (approved):
Approved after the requested fixes. The reviewed head passes CI, all 2,150 local tests, type checking, lint, and lat check. Greptile also re-reviewed this head at 5/5. Thanks for the security improvement.

PR #906, opened 2026-08-12
Add MiniMax speech synthesis

octo-patch:
Reason: Add MiniMax speech synthesis to the desktop TTS toolset with current models, regional endpoints, and response handling.

## Summary

- add a context-isolated speech synthesis boundary with region-bound credentials
- support the current speech model catalog and synchronous hex-audio responses
- expose speech model and endpoint selection and a listen action for agent messages

## Checks

- `npx vitest run tests/constants.test.ts tests/minimax-speech.test.ts tests/preload-api-surface.test.ts src/renderer/src/screens/Chat/MessageRow.test.tsx`
- `npm run typecheck:node`
- `npm run typecheck:web`
- `npm exec --yes --package lat.md -- lat check`


fathah (changes requested):
Thanks @octo-patch for the speech integration. I checked the synthesis boundary against Hermes Agent’s MiniMax region/key conventions and reviewed Greptile’s feedback. The fixed regional endpoints and credential separation are useful, but this is not ready to merge.

I agree with Greptile’s two blocking UI findings: every message row owns an independent Audio instance, so clicking Listen on a second message leaves the first speaking; and all synthesis/playback failures return silently to idle, including missing credentials. A shared playback owner and actionable error feedback need regression coverage.

There is also no timeout or cancellation signal on the main-process fetch. Stop only increments a renderer counter; a stalled response can leave an IPC request and paid synthesis running indefinitely while further requests are started. Add a bounded request lifecycle, coordinate playback, and validate stop/unmount/error behavior before merging. The selected PCM format also needs a playable container/decoder rather than passing raw PCM to HTMLAudioElement as audio/L16.

Holding this feature until those lifecycle and playback paths are covered. No code from this PR has been merged.

PR #908, opened 2026-08-17
fix: Merge pull request #80 from MillsAirCode/fix/office-scroll-overflow

Ricardo-M-L:
This PR addresses: Merge pull request #80 from MillsAirCode/fix/office-scroll-overflow

### DIFF
--- a/src/main/session-cache.ts
+++ b/src/main/session-cache.ts
@@ -76,83 +76,91 @@ function getDb(): Database.Database | null {
 
 // Sync from hermes DB to local cache — only fetches new/updated sessions
 export function syncSessionCache(): CachedSession[] {
-  const cache = readCache();
-  const db = getDb();
-  if (!db) return cache.sessions;
-
+  _cacheLock = true;
   try {
-    // Fetch sessions newer than last sync, or all if first sync
-    const rows = db
-      .prepare(
-        `SELECT s.id, s.started_at, s.source, s.message_count, s.model, s.title
-         FROM sessions s
-         WHERE s.started_at > ?
-         ORDER BY s.started_at DESC`,
-      )
-      .all(cache.lastSync > 0 ? cache.lastSync - 300 : 0) as Array<{
-      id: string;
-      started_at: number;
-      source: string;
-      message_count: number;
-      model: string;
-      title: string | null;
-    }>;
-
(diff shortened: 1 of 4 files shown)


src/main/session-cache.ts
```diff
@@ -165,15 +173,23 @@ export function listCachedSessions(
   return cache.sessions.slice(offset, offset + limit);
 }
 
+let _cacheLock = false;
```
AmirF194, line 176: Static review against `7af232a`, not executed.

`_cacheLock` (session-cache.ts:176) can never be observed as true. `syncSessionCache` and `updateSessionTitle` are both fully synchronous: no `await`, `async`, or `.then` anywhere in this file, and the DB (better-sqlite3, sync API) and file I/O (`readFileSync`/`safeWriteFile`) calls it makes are synchronous too. Since neither function yields control mid-execution, a call can only start after any prior call to either has already run to completion and reset the lock, so the guard at line 183 (`if (_cacheLock) return;`) can never see `true` and `updateSessionTitle` can never silently drop a write the way the guard implies it might.

Not wrong, just dead: two extra state writes and a branch that reads as protecting against a race that a single-threaded, all-synchronous file already rules out. If this is meant to guard a future async version, a one-line comment saying so would save the next reader from re-deriving this; as written it just suggests a hazard that is not there.

fathah (changes requested):
Thanks @Ricardo-M-L for the lifecycle cleanup work. I reviewed the actual changes and Greptile's feedback against current main.

The installer mounted guards are already present on main, and the gateway screen now has a tracked status timer and a different platform-management flow. The remaining cache lock does not address a reachable race: both cache operations are synchronous with no await/yield point, while a process-local boolean would not coordinate writes from another process. The added smoke tests also use the real configured Hermes home instead of an isolated fixture and have unused imports flagged by lint.

I am holding this PR as largely superseded, rather than bringing the old cache implementation back through its conflicts. Any remaining cache concurrency bug needs an isolated reproduction and a fix at the actual shared-write boundary. The PR title/body should also describe the session-cache/lifecycle changes rather than the unrelated office-scroll merge.

src/main/session-cache.ts
AmirF194: Agreed, thanks for confirming. No yield point anywhere in this file is exactly why the lock can never see contention.

PR #909, opened 2026-08-17
security: pin installer script by SHA-256 + escape provider value in Memory hint

azoz778:
Rebased onto current `main` (clears the earlier conflict). Scope is now trimmed to only what upstream has **not** already addressed.

The original branch also hardened main-window navigation/sandbox/webview and the askpass dialog. Since then, `main` has independently implemented equivalent-or-stronger protections for both, so those two changes are **dropped** here as redundant:
- `src/main/app/start.ts` sandboxes the renderer and routes `will-navigate` / `will-attach-webview` through shared guards (`isAllowedAppNavigationUrl`, `hardenWebviewPreferences`).
- `src/main/askpass.ts` now runs the dialog with `contextIsolation: true` + `sandbox: true` + a compiled preload and a strict CSP.

### What remains (two fixes upstream has not made)

**1. `src/main/installer.ts` — pin + checksum the install script (supply-chain).**
The macOS/Linux install path still pipes a mutable `main` branch of `scripts/install.sh` straight into `bash`. This pins it to a reviewed commit and verifies its SHA-256 before executing (portable `sha256sum` / `shasum -a 256`, aborts on mismatch). To update the pin: bump `PINNED_INSTALL_SHA` and `EXPECTED_INSTALL_SHA256`.
> Note: the Windows PowerShell path (`install.ps1`) still fetches `main`; left as-is to keep this PR's scope unchanged — a maintainer may want to pin that similarly.

**2. `src/renderer/src/screens/Memory/MemoryProviders.tsx` — escape a value going into `dangerouslySetInnerHTML` (XSS sink).**
The active-provider name is interpolated into a translation string rendered via `dangerouslySetInnerHTML`. Escape the runtime value first; the translation markup itself stays trusted.

Both changed files are type-clean (the only `tsc` errors in this tree are pre-existing missing-dependency errors in unrelated Office3D/wallet/skills files).

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/src/main/installer.ts
+++ b/src/main/installer.ts
@@ -937,10 +937,30 @@ export async function runInstall(
       // then run the official install script. Electron apps launched from Finder
       // don't inherit the terminal environment.
       const shellProfile = getShellProfile(home);
+
+      // [SECURITY 2026-07-03] Pin the installer to a specific commit and verify
+      // its SHA-256 before executing, instead of piping a mutable `main` branch
+      // straight into bash. To update: bump PINNED_INSTALL_SHA to the reviewed
+      // commit and set EXPECTED_INSTALL_SHA256 to the sha256 of scripts/install.sh
+      // at that commit ( curl -fsSL <raw-url> | sha256sum ).
+      const PINNED_INSTALL_SHA = "528159f7aa6a0c234c61685f552cb45f22849c52";
+      const EXPECTED_INSTALL_SHA256 =
+        "a93c65b01ea392e179cf872e182bd01a2b65c0c15f17833e9f9569033ef10e07";
+      const pinnedInstallUrl = `https://raw.githubusercontent.com/NousResearch/hermes-agent/${PINNED_INSTALL_SHA}/scripts/install.sh`;
+      const verifiedInstall = [
+        "set -e",
+        'TMP="$(mktemp)"',
+        `curl -fsSL ${pinnedInstallUrl} -o "$TMP"`,
+        // portable checksum: sha256sum (Linux) or shasum -a 256 (macOS)
+        `ACTUAL="$( (sha256sum "$TMP" 2>/dev/null || shasum -a 256 "$TMP") | awk '{print $1}' )"`,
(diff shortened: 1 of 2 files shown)


AmirF194 (commented):
I read this against `d58fc502` in a clean checkout, not just the diff.

The escape fix at `src/renderer/src/screens/Memory/MemoryProviders.tsx:73-78` is correct, and I confirmed the pinned installer hash: fetching `scripts/install.sh` at commit `528159f7aa6a0c234c61685f552cb45f22849c52` and running `sha256sum` on it matches `EXPECTED_INSTALL_SHA256` exactly.

The same pattern survives untouched at `src/renderer/src/components/settings/DataPane.tsx:82-86`:

```tsx
dangerouslySetInnerHTML={{
  __html: t("settings.migrationDesc", {
    path: openclawPath || "",
  }),
}}
```

`openclawPath` is set in `useSettingsData.ts:241-243` from `window.hermesAPI.checkOpenClaw()` and interpolated into a translated string the same way `currentProvider` was here, then rendered raw. It is normally a fixed install-directory path, so the practical exploit window is narrow: an OS username or path segment containing HTML metacharacters. But it is the exact pattern this PR's own comment describes as the risk. Escaping it with the same helper (worth lifting to a shared util, since there are now two call sites) would close it too.

I did not check `ConnectionPane.tsx:421` or `WorktreePanel.tsx:44`, both `dangerouslySetInnerHTML` as well: the former interpolates nothing, the latter renders a bundled icon string, neither takes a runtime value the way these two do.

fathah (approved):
Thanks @azoz778 for the contribution and for improving Hermes One.

Reviewed the bootstrap against the current Hermes Agent source and refreshed the commit/checksum pin. Download or verification failures now reject installation even when old binaries exist; temporary files are cleaned on all tested exits. Added executable checksum tests and translated provider-name HTML regressions.

Reviewed commit: `7531780b88a1d558af230886d31f13c6ff29f19d`.

AmirF194 (commented):
Thanks for confirming. The refreshed installer hash still matches what I verified against 528159f.

Issue #799, opened 2026-07-02
Remote mode incompatible with hermes-agent 0.17.0 — dashboard/sessions transport (X-Hermes-Session-Token, /api/profiles/sessions, WS) spams 401s and breaks auto chat


alexjfarling:
Related: #667 (same underlying cause — the app assumes the gateway serves the dashboard transport, which only exists on the separate Hermes Dashboard component, not on the gateway). #667 is the SSH-tunnel/WebSocket face of it; this report is the Remote-mode face, adding the sessions-poll 401 flood and the header/endpoint mismatch details.

App: Hermes One 0.7.2 (macOS, Apple Silicon) · Gateway: hermes_agent 0.17.0 · Mode: Remote, over Tailscale to http://<host>:8642

Summary: With the default auto chat transport, Hermes One can't chat with a hermes-agent 0.17.0 gateway and floods it with ~1 401 per minute. The API key is correct (verified byte-identical to the gateway's API_SERVER_KEY). Root cause is the dashboard/sessions transport using an auth header and endpoints this gateway version doesn't implement.

Details / evidence (verified with curl against the gateway):

GET /api/sessions — the app sends the key in header X-Hermes-Session-Token (see remoteRequestJson in out/main/index.js), but the api_server only accepts Authorization: Bearer:
Authorization: Bearer <key> → 200
X-Hermes-Session-Token: <key> → 401 ("rejected invalid API key")
The session list first calls GET /api/profiles/sessions?...&profile=all → 404 on 0.17.0 (endpoint absent), then falls back to GET /api/sessions → 401 (header mismatch above).
sync-session-cache polls this ~1/min unconditionally in remote mode, so it keeps 401-spamming even when the dashboard is known-unavailable.
Chat in auto surfaces a 404 Not Found transport error (dashboard WS//api/ws endpoints absent on the gateway — cf. #667). Chat over POST /v1/chat/completions + Authorization: Bearer works fine (200).
Workaround: set chat transport to Legacy (remoteChatTransport: "legacy" in ~/.hermes/desktop.json) — chat works, but the sessions-sidebar 401 poll continues (no setting disables it). This matches #667's note that v0.5.8 (HTTP /v1/chat/completions) worked while v0.6.1+ (WebSocket) broke.

Suggested fixes:

For dashboard/sessions requests, fall back to Authorization: Bearer when X-Hermes-Session-Token returns 401 (or negotiate based on server capability).
[...]

Issue #667, opened 2026-06-14
SSH Tunnel mode: chat messages not sent in v0.6.1 Windows


hondoaus:
## Environment
- Hermes Desktop: v0.6.1 (Windows, fresh install)
- Connection mode: SSH Tunnel
- Backend: Hermes Agent v0.16.0 on remote Linux VPS
- SSH key auth (ed25519)

## Bug
Chat messages are never sent through the SSH tunnel in v0.6.1. The UI shows indefinite loading but no request reaches the backend.

## Root Cause (Confirmed)

**Hermes One v0.6.1 uses WebSocket for chat transport**, connecting to:
```
ws://127.0.0.1:{port}/api/ws?token={token}
```

**The Hermes Agent gateway** (`hermes gateway run`, port 8642) only provides HTTP endpoints - NO `/api/ws` WebSocket endpoint exists.

The WebSocket endpoint `/api/ws` exists in the **Hermes Dashboard** (`hermes dashboard`, port 9119), a separate component. Confirmed by:
- Starting dashboard on 9119: `curl http://127.0.0.1:9119/api/ws` returns `{"detail":"Unauthorized"}` (endpoint exists)
- But Hermes One connects to tunnel port (18642→8642), not 9119

**This means:**
- v0.5.8 worked (used HTTP `/v1/chat/completions` through tunnel)
- v0.6.1 broke (switched to WebSocket `/api/ws` which doesn't exist on gateway)
- Other users with "working" SSH tunnel may be running `hermes dashboard` alongside gateway, or using Remote mode

## Evidence
- Source code: `connectWebSocket("ws://127.0.0.1:${port}/api/ws")` in main/index.js
- `/api/ws` returns 404 on port 8642 (gateway) and 8643 (TUI gateway)
- Zero outbound connections from Hermes One when sending chat (netstat monitoring over 30+ seconds)
- Tunnel works: `curl http://127.0.0.1:18642/health` → 200 OK
- Backend works: `curl -X POST http://127.0.0.1:18642/v1/chat/completions` with auth → 200 OK
- Gateway restart did not fix the issue

## Repro
1. Install Hermes Desktop v0.6.1
2. Configure SSH Tunnel mode
3. Verify "SSH Tunnel Connected"
4. Send a chat message to any profile → indefinite loading

## Workaround
Remote mode (direct HTTP) works for basic chat but lacks profile switching and session history.

## Additional info
- Related upstream issues: #358, #577
[...]

PR #912, opened 2026-08-21
fix(remote): send Authorization: Bearer on dashboard/session requests

AmirF194:
`remoteRequestJson` (`src/main/remote-sessions.ts`) only sends `X-Hermes-Session-Token` for
the dashboard/sessions REST calls used in Remote mode with an API-key connection. Some
gateway versions accept `Authorization: Bearer <key>` on these endpoints but reject
`X-Hermes-Session-Token`, so every dashboard request 401s and `sync-session-cache` retries
it roughly once a minute. The sibling OAuth path (`requestRemoteOAuthJson` in
`src/main/remote-oauth.ts`) already sends `Authorization: Bearer` and works fine, which is
the pattern this PR aligns the API-key path with.

Fix: send both headers, `Authorization: Bearer <key>` alongside the existing
`X-Hermes-Session-Token`, so a gateway that only accepts one of the two now succeeds either
way.

**Verified:**
- New regression test (`tests/remote-sessions.test.ts`) spins up a real local `http.Server`
  and asserts the outgoing request carries `Authorization: Bearer <token>`. It fails against
  unmodified `main` and passes on this branch.
- Full suite (`npm test`, clean `node:22-slim` container): 187 files / 1896 passed, 6
  pre-existing skips, no new failures.
- `npm run typecheck` and `eslint` on the two changed files are clean.
- Not verified: an actual round trip against a real `hermes-agent` gateway (none available
  here). Three separate reporters on this issue independently curl-tested the equivalent
  live request and confirmed `Authorization: Bearer` succeeds where
  `X-Hermes-Session-Token` 401s on their gateways.

Refs #799: this issue also asks for the sessions-poll to back off when the dashboard is
genuinely unavailable, and for `auto` transport to detect gateway capability instead of
surfacing a 404 (the WS/`#667` side of the same root cause). Both are separate, more
judgment-laden changes and out of scope for this minimal header fix.

### DIFF
--- a/src/main/remote-sessions.ts
+++ b/src/main/remote-sessions.ts
@@ -102,6 +102,7 @@ export function remoteRequestJson<T>(
         method: options.method ?? "GET",
         headers: {
           "Content-Type": "application/json",
+          Authorization: `Bearer ${token}`,
           "X-Hermes-Session-Token": token,
           ...(body ? { "Content-Length": Buffer.byteLength(body) } : {}),
         },
--- a/tests/remote-sessions.test.ts
+++ b/tests/remote-sessions.test.ts
@@ -19,6 +19,7 @@ interface RecordedRequest {
   method: string;
   url: string;
   token: string;
+  authorization: string;
   body: string;
 }
 
@@ -39,6 +40,7 @@ describe("remote session REST bridge", () => {
           method: req.method || "GET",
           url: req.url || "",
           token: String(req.headers["x-hermes-session-token"] || ""),
+          authorization: String(req.headers["authorization"] || ""),
           body,
         });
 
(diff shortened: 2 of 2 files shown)


AmirF194:
Checking in on this when you get a chance, no rush. CI is green and greptile's review found no issues, happy to rebase or adjust scope if that helps.

fathah (approved):
Thanks @AmirF194 for the contribution and for improving Hermes One.

Cross-checked the upstream authentication contract and corrected the Basic-proxy regression: the dedicated session-token header is retained, and URL credentials can continue supplying Basic Authorization. Actual HTTP-server tests cover Basic-plus-token, Bearer-only, and legacy token-header authentication.

Reviewed commit: `8b44d488b01ad75b5bc3aa24e0c989a539585b2c`.

AmirF194:
Good catch on keeping the session-token header alongside Basic auth, that combination is exactly the case my fix missed.

Issue #913, opened 2026-08-22
Migration banner renders the detected legacy-install path unescaped via dangerouslySetInnerHTML


AmirF194:
DataPane.tsx builds the migration banner text with `t("settings.migrationDesc", { path: openclawPath || "" })` and renders it through `dangerouslySetInnerHTML`. The app's i18next instance is configured with `interpolation: { escapeValue: false }` (matches React's usual assumption that JSX escapes text), but this call site bypasses that by going through `dangerouslySetInnerHTML`, so `path` reaches the DOM without any HTML escaping.

`openclawPath` comes from `window.hermesAPI.checkOpenClaw().path`, so in normal use it is a local filesystem path, not remote-attacker input. Still, anything that can influence that path (a crafted OpenClaw install location, a symlink, or a future caller of the same helper) gets arbitrary markup execution in the settings window.

Repro (Node 22, `i18next@25.6.0`, the version this repo pins): initialize i18next with the exact `migrationDesc` string from `src/shared/i18n/locales/en/settings.ts` and the repo's own `escapeValue: false` config, then interpolate a `path` containing `"><img src=x onerror=...>`. The output string contains the payload verbatim, unescaped:

```js
const i18next = require("i18next");
const inst = i18next.createInstance();
await inst.init({
  lng: "en", fallbackLng: "en",
  resources: { en: { translation: { settings: {
    migrationDesc: "Found OpenClaw at <code>{{path}}</code>. You can migrate your configuration, API keys, sessions, and skills to Hermes.",
  } } } },
  interpolation: { escapeValue: false },
});
console.log(inst.t("settings.migrationDesc", { path: '"><img src=x onerror=alert(1)>' }));
// -> Found OpenClaw at <code>"><img src=x onerror=alert(1)></code>. ...
```

Expected: the interpolated path is HTML-escaped before it reaches `dangerouslySetInnerHTML`, same as the fix already proposed for the identical pattern in `MemoryProviders.tsx` (open PR #909). Actual: it is not, and DataPane.tsx is untouched by that PR.

Happy to send a PR for this if useful.

PR #922, opened 2026-08-26
fix(security): escape OpenClaw path in migration banner to prevent XSS (#913)

aniruddhaadak80:
## What changed

Escape the runtime `openclawPath` value before it is interpolated into the `settings.migrationDesc` translation string and rendered via `dangerouslySetInnerHTML` in `DataPane`.

The translation markup itself (`<code>{{path}}</code>`) is trusted; only the filesystem path — which comes from disk — must be escaped.

## Why

`i18next` is configured globally with `escapeValue: false` (`src/shared/i18n/index.ts:602`), meaning interpolated values land in the returned string raw. Because `DataPane` renders this string via `dangerouslySetInnerHTML`, an attacker who can influence the discovered OpenClaw path (e.g. a directory named ``) gets arbitrary HTML injected into the Electron renderer process.

## Root cause

No escaping is applied to the runtime `openclawPath` value before it reaches `dangerouslySetInnerHTML` via the i18next interpolation pipeline.

## User impact

Medium-severity stored XSS in the Electron renderer: a malicious directory name on disk becomes executable HTML the next time the user opens Settings → Data. Because the renderer runs in an Electron context with `nodeIntegration` capabilities, this is higher-impact than a typical browser XSS.

## Validation

- `npm run typecheck` — clean (both `typecheck:node` and `typecheck:web`)
- `npx vitest run DataPane.test.tsx` — 2/2 pass:
  - renders a normal path inside the trusted `<code>` wrapper
  - never injects HTML from the discovered OpenClaw path (verifies `innerHTML` contains `&lt;img` not `` wrapper survives as real DOM)
- **Fail-without-fix proof:** reverting only the interpolation line causes the "never injects HTML" test to fail (exit 1), confirming the test locks the vulnerability.

## Follow-up

Open PR #909 fixes the same pattern in `MemoryProviders.tsx`. This PR scopes exclusively to `DataPane.tsx` per the small/focused contribution policy. Once both land, the `escapeHtml` helper could be extracted to a shared utility — but that is a maintainer decision.

### DIFF
--- a/src/renderer/src/components/settings/DataPane.test.tsx
+++ b/src/renderer/src/components/settings/DataPane.test.tsx
@@ -1,0 +1,70 @@
+import { render } from "@testing-library/react";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import DataPane from "./DataPane";
+
+// Mimic i18next configured with escapeValue: false (see src/shared/i18n/index.ts):
+// interpolated values land in the returned string raw, so any runtime value a
+// pane passes through t() ends up unescaped in markup rendered via
+// dangerouslySetInnerHTML.
+const rawTemplate =
+  "Found OpenClaw at <code>{{path}}</code>. You can migrate your configuration.";
+
+vi.mock("../useI18n", () => ({
+  useI18n: () => ({
+    t: (key: string, opts?: Record<string, string>) =>
+      key === "settings.migrationDesc"
+        ? rawTemplate.replace("{{path}}", opts?.path ?? "")
+        : key,
+  }),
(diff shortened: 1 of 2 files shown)


aniruddhaadak80:
Closing as superseded: upstream 0144305 (2026-09-02, Fix migration path HTML injection) landed the same fix for #913 after this branch (08-26) - escapeHtml applied at the openclawPath sink, plus locale-parametrized migration banner tests covering it. A rebase confirms zero remaining delta on this branch. Thanks for landing it!

PR #926, opened 2026-08-28
fix(chat): accept named provider vs normalized custom in model verification

### LINKED ISSUE #925: False "dashboard did not switch" error after profile switch: named provider (omlx) vs normalized live provider (custom)
## Summary

After switching to a profile whose model uses a **named provider from `providers:` in `config.yaml`** (e.g. `omlx`, `lmstudio`), sending a message in Hermes One can pop a false error like:

```
Hermes dashboard did not switch to omlx/Qwen3.8-Whittle-MoE-27B-MLX-4bit;
live model is custom/Qwen3.8-Whittle-MoE-27B-MLX-4bit;
/model output: ✓ Model switched: ... Provider: omlx ...;
custom inventory: no custom providers listed
```

The switch **actually succeeded** (`/model` shows `Provider: omlx`, the model answers normally, `GET /v1/models` on the local server lists the model). The failure is only in the desktop-side verification: Hermes Agent normalizes every named `providers:` entry down to the bare billing namespace `"custom"` when reporting the live model, while Hermes One requested the model under its configured name (`omlx`). The string comparison then reports a mismatch.

## Where

- Throw site: `src/renderer/src/screens/Chat/hooks/useDashboardChatTransport.ts` — `ensureSelectedModel()` / `switchAndValidate()` error at the `Hermes dashboard did not switch to ...` template.
- Root cause: `dashboardModelMatches()` in the same file. Its provider-alias tolerance only covers one direction:

```ts
// Named custom providers can be reported by Hermes Agent as custom:<slug>
// while Hermes One's older model config still treats them as custom rows.
return provider === "custom" && liveProvider.startsWith("custom:");
```

It accepts `custom → custom:<slug>`, but **not** the equally common `omlx → custom` (requested name → normalized live name).

Related: `modelOptionsSummary()` only enumerates slugs starting with `custom:`, so the error footer prints `custom inventory: no custom providers listed` even when `providers.omlx` exists in config — the inventory the validator could use for the mapping is invisible to the message, and arguably to the check itself.

## Reproduce

1. Profile `config.yaml` (secrets omitted):

```yaml
model:
  provider: "custom"
  default: "Qwen3.8-Whittle-MoE-27B-MLX-4bit"
  base_url: "http://127.0.0.1:9999/v1"
providers:
  omlx:
    base_url: "http://127.0.0.1:9999/v1"
    # api_key omitted
```

2. Start the local server so `/v1/models` lists `Qwen3.8-Whittle-MoE-27B-MLX-4bit`.
3. Switch to this profile in Hermes One and pick the model (menu shows it under `omlx`).
4. Send a message.

**Actual:** error dialog "did not switch to omlx/...; live model is custom/..." (blocking the send).
**Expected:** send proceeds — the two identities describe the same endpoint.

Environment: Hermes One.app 0.7.6 (macOS), source at `main` (3aaadb0) still contains the logic.

## Suggested fix

In `dashboardModelMatches()`, when the live provider is in the normalized `custom` namespace, treat the request as satisfied if the requested provider name exists in `live.providers` (row slug `provider` or `custom:<provider>`) **and** that row lists the requested model. This is the symmetric counterpart of the existing `custom → custom:<slug>` tolerance and cannot mask a genuinely wrong model (the model-name equality check still applies).

I have a branch with this fix + unit tests ready to open a PR: `dashboardModelMatches` named-custom equivalence (regression tests: accepts `omlx → custom`, accepts `custom:<slug>` spelling, still rejects different model / absent provider / provider row that doesn't list the model).

## Workaround (no code change)

Avoid selecting the named `omlx/...` picker entry after a profile switch; instead pick the generic custom endpoint row (or run `/model <name>` without `--provider`), so requested and live identity are both `custom` and the check passes trivially.

windyleave:
Fixes #925

## Problem
After switching to a profile that uses a named `providers:` entry (e.g. `omlx`, `lmstudio`), sending a message can throw a false error:

> Hermes dashboard did not switch to omlx/Qwen3.8-Whittle-MoE-27B-MLX-4bit; live model is custom/...

Hermes Agent normalizes every named custom provider down to the bare billing class `custom` in `model.options`, while Hermes One requests the model under its configured entry name. `dashboardModelMatches()` only tolerated the `custom → custom:<slug>` direction, so `omlx → custom` (same endpoint, same model) failed the string comparison and blocked the send — even though `/model` confirms the switch succeeded.

## Approach
Add the symmetric tolerance in `dashboardModelMatches()`: when the live provider is in the normalized `custom` namespace, accept the request if the requested provider appears in the live inventory (as `<provider>` or `custom:<provider>`) **and** that row lists the requested model. The model-name equality check still applies, so a genuinely wrong model or absent provider is still rejected.

## Tests
5 unit tests on `dashboardModelMatches` in `useDashboardChatTransport.test.tsx`: named→custom accepted, `custom:<slug>` spelling accepted, wrong model rejected, absent provider rejected, provider row not listing the model rejected. Verified they fail without the fix (sabotage run). Full file green (18 passed); `tsc --noEmit -p tsconfig.web.json` clean; eslint clean.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/renderer/src/screens/Chat/hooks/useDashboardChatTransport.ts
+++ b/src/renderer/src/screens/Chat/hooks/useDashboardChatTransport.ts
@@ -575,7 +575,25 @@ export function dashboardModelMatches(
 
   // Named custom providers can be reported by Hermes Agent as custom:<slug>
   // while Hermes One's older model config still treats them as custom rows.
-  return provider === "custom" && liveProvider.startsWith("custom:");
+  if (provider === "custom" && liveProvider.startsWith("custom:")) return true;
+
+  // The reverse direction: Hermes Agent normalizes every named provider from
+  // config.yaml `providers:` (e.g. `omlx`, `lmstudio`) down to the bare
+  // billing namespace "custom" when it reports the live model, while the
+  // desktop requested the entry by its configured name. If the requested
+  // provider exists in the live inventory as a row that lists the requested
+  // model, the two identities describe the same endpoint and the string
+  // mismatch is a namespace artifact — not a failed switch. Without this,
+  // sending after a profile switch throws a false "did not switch" error.
+  const requestedRow =
+    liveProvider === "custom" || liveProvider.startsWith("custom:")
+      ? (live?.providers ?? []).find(
+          (row) =>
+            row.slug?.toLowerCase() === provider ||
+            row.slug?.toLowerCase() === `custom:${provider}`,
+        )
+      : undefined;
+  return !!requestedRow && modelIsListedByProvider(requestedRow, model);
 }
 
 function asRecord(value: unknown): Record<string, unknown> {
--- a/src/renderer/src/screens/Chat/hooks/useDashboardChatTransport.test.tsx
+++ b/src/renderer/src/screens/Chat/hooks/useDashboardChatTransport.test.tsx
@@ -10,7 +10,10 @@ import {
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import type { Mock } from "vitest";
 import type { DashboardRpcEvent } from "../dashboardGatewayClient";
-import { useDashboardChatTransport } from "./useDashboardChatTransport";
+import {
+  dashboardModelMatches,
+  useDashboardChatTransport,
+} from "./useDashboardChatTransport";
 import type { ActiveTurn, ChatMessage, UsageState } from "../types";
 
 type SetUsageMock = Mock<(value: SetStateAction<UsageState | null>) => void>;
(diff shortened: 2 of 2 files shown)


src/renderer/src/screens/Chat/hooks/useDashboardChatTransport.ts
```diff
@@ -575,7 +575,25 @@ export function dashboardModelMatches(
...
+
+  // The reverse direction: Hermes Agent normalizes every named provider from
+  // config.yaml `providers:` (e.g. `omlx`, `lmstudio`) down to the bare
+  // billing namespace "custom" when it reports the live model, while the
+  // desktop requested the entry by its configured name. If the requested
+  // provider exists in the live inventory as a row that lists the requested
+  // model, the two identities describe the same endpoint and the string
+  // mismatch is a namespace artifact — not a failed switch. Without this,
+  // sending after a profile switch throws a false "did not switch" error.
+  const requestedRow =
```
AmirF194, line 588: I ran the exact `dashboardModelMatches`/`modelIsListedByProvider` logic added in this PR (head `07215ef2e5b912a95f9035765970e1299c4f0cf6`) standalone in a clean Docker container, with only TS type annotations stripped, nothing else changed.

The new fallback at lines 588-596 finds `requestedRow` by matching `slug` alone and never checks `row.is_current`, even though this same file already uses `is_current` a few lines up (544, 553) to pick the actually-active provider row. If two provider rows happen to list a model with the same name, which is common for local runtimes serving the same downloaded weights under two different entries, the function reports a successful switch to the requested provider even when a different row is the one actually live.

Repro (constructed from the same shapes your own new tests use):

```
live = {
  model: "Qwen3.8-Whittle-MoE-27B-MLX-4bit",
  provider: "custom:lm",   // lm is what is ACTUALLY live
  providers: [
    { slug: "omlx", models: [MODEL], is_current: false, base_url: "http://127.0.0.1:9999/v1" },
    { slug: "custom:lm", models: [MODEL], is_current: true, base_url: "http://127.0.0.1:1234/v1" },
  ],
}
dashboardModelMatches("omlx", MODEL, live) => true
```

That result feeds straight into `switchAndValidate` at line 1425: when it comes back `true`, the "did not switch" error is skipped and `appliedModelRef.current` is set to the requested `omlx` key, so the send proceeds silently against whatever is actually live (`custom:lm` here) while the app believes it switched to `omlx`.

Checking `row.is_current` in the find condition, in addition to the slug match, would close this: only accept the fallback when the matched row is the one the server reports as current, the same signal `modelOptionsSummary` already trusts nearby.

I did not test this against a live Hermes Agent instance, only the extracted matching function.

PR #927, opened 2026-08-29
fix: preserve completed schedule state

Sushanth012:
## Problem

Hermes disables jobs after they complete. Schedule normalization currently checks the disabled flag before the explicit `completed` state, so completed jobs appear paused and expose the wrong UI actions.

## Changes

- give the terminal `completed` state precedence over disabled/paused normalization
- add a remote API regression test
- document the schedule-state normalization rule in `lat.md`

## Verification

- `npm test -- tests/cronjobs-ssh.test.ts`
- `npm run typecheck:node`
- `npm exec --yes --package=lat.md -- lat check`


fathah (approved):
Thanks @Sushanth012 for the contribution and for improving Hermes One.

Completed jobs now remain terminal and disabled across local and SSH paths, matching upstream scheduling behavior. Added real jobs-file and SSH coverage so completed jobs cannot leak into active-only lists.

Reviewed commit: `36305f45a3b0bc6b125b84618bf155f5815f962c`.

PR #931, opened 2026-09-03
fix(settings): escape migration path before rendering HTML

### LINKED ISSUE #913: Migration banner renders the detected legacy-install path unescaped via dangerouslySetInnerHTML

Sushanth012:
## Summary
- HTML-escape the detected OpenClaw install path before interpolating it into the migration banner's trusted localized markup
- preserve the existing localized `<code>` presentation and exact visible filesystem path
- add a renderer regression test using a markup-bearing path
- document the runtime-value boundary in `lat.md`

## Why
`openclawPath` comes from the filesystem and was interpolated with i18next's global `escapeValue: false` setting before the result reached `dangerouslySetInnerHTML`. A crafted install path could therefore create arbitrary markup in the settings renderer. Escaping the five HTML-significant characters at this boundary keeps translation-owned markup intact while ensuring the runtime path remains text.

## Testing
- `npx vitest run src/renderer/src/components/settings/DataPane.test.tsx`
- `npm run typecheck:web`
- `npx prettier --check src/renderer/src/components/settings/DataPane.tsx src/renderer/src/components/settings/DataPane.test.tsx lat.md/sidebar-navigation.md`
- `npx eslint src/renderer/src/components/settings/DataPane.tsx src/renderer/src/components/settings/DataPane.test.tsx`
- `npx lat.md check`

Closes #913


AmirF194 (commented):
I compared this PR against the other open PRs touching the same issue before reading it in isolation.

#922, opened 2026-08-26 (8 days before this one), makes the same fix to this same file: identical `escapeHtml()` body, identical call-site change at the `path` interpolation. Diffing the two patches for `DataPane.tsx`, the only difference is a header comment:

```
$ diff <(gh api repos/fathah/hermes-desktop/pulls/922/files --paginate -q '.[]|select(.filename=="src/renderer/src/components/settings/DataPane.tsx")|.patch') \
       <(gh api repos/fathah/hermes-desktop/pulls/931/files --paginate -q '.[]|select(.filename=="src/renderer/src/components/settings/DataPane.tsx")|.patch')
1c1
< @@ -2,6 +2,19 @@ import { Download, Upload } from "lucide-react";
---
> @@ -2,6 +2,15 @@ import { Download, Upload } from "lucide-react";
5,8d4
< +// Escape a runtime value before interpolating it into a translation string
< +// that is rendered via dangerouslySetInnerHTML ...
```

#922 also already has a fail-without-fix regression proof and a clean typecheck, and its own body already names a third PR, #909, which applies the identical pattern to `MemoryProviders.tsx:61-65` (same sink, `t()` result into `dangerouslySetInnerHTML`, sourced from the `memory.provider` config value there instead of the migration path this PR fixes). I checked that sink is real: `getConfigValue` in `src/main/config.ts` only strips YAML quotes, no HTML escaping, and with i18next's global `escapeValue: false` a crafted `memory.provider` value reaches the DOM unescaped, same as the bug this PR fixes.

Given #922 is earlier and further along, and #909 covers the sibling sink, it may be worth closing this one as a duplicate rather than reviewing three PRs for what is one fix plus one sibling fix.

fathah (approved):
Thanks @Sushanth012 for the contribution and for improving Hermes One.

The migration path is escaped at the HTML rendering boundary while filesystem operations retain the raw path. Added real-locale/path regressions, including markup and literal entities, and verified the installer discovery path.

Reviewed commit: `7dbff9d0258dd49e893926557624a7cec8bab3c7`.

PR #942, opened 2026-09-07
fix: honor adopted home across inherited environment restart

JaimeMarques:
## Summary

- remember the exact inherited `HERMES_HOME` replaced by the explicit “use existing installation” choice
- reuse the selected installation when that same inherited value is present after restart
- revalidate the persisted installation before it can shadow the inherited environment
- preserve normal environment precedence for legacy overrides and any different later `HERMES_HOME`

## Problem

When the desktop app is launched from a process that already defines `HERMES_HOME`, selecting a different existing installation persists the choice but the inherited environment value wins again after restart. The app returns to onboarding and the selection appears ineffective.

The override now records a one-way fingerprint of only the environment value it intentionally replaces. That exact value is shadowed on subsequent launches while the adopted installation remains complete; all other precedence behavior remains unchanged.

## Testing

- `npm test -- tests/installer-home-override.test.ts tests/installer-target.test.ts src/renderer/src/screens/Install/Install.test.tsx`
- `LC_ALL=C.UTF-8 LANG=C.UTF-8 npm test`
- `npm run typecheck`
- `npm run lint`
- `npx prettier --check src/main/installer.ts tests/installer-home-override.test.ts lat.md/onboarding.md`
- `npx --yes lat.md check`
- `npm run build`

### DIFF
--- a/lat.md/onboarding.md
+++ b/lat.md/onboarding.md
@@ -38,9 +38,9 @@ The progress view (`wide`) shows a step + percent header with a progress bar, th
 
 An adopted install wins over the exact inherited `HERMES_HOME` active during selection, preventing restart loops while preserving a different later environment override.
 
-[[src/main/installer.ts#setHermesHomeOverride]] records a one-way fingerprint of that shadowed environment value beside the selected home. On restart, a matching fingerprint activates the saved selection; absent, legacy, and later-different environment values keep the existing precedence.
+[[src/main/installer.ts#setHermesHomeOverride]] records a one-way fingerprint of that shadowed environment value beside the selected home. On restart, a matching fingerprint activates the saved selection only while it remains a desktop-compatible install; absent, legacy, incomplete, and later-different values keep the existing precedence.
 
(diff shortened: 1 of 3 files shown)


fathah (commented):
Thanks @MrTheSoulz for the contribution and for improving Hermes One.

Reviewed adopted-home validation and inherited environment precedence against the upstream Windows environment setup. The fix breaks the repeated adoption loop while allowing a changed inherited environment to take precedence. Local validation: 48 focused tests passed.

Reviewed commit: `c0546c7de5ba55e05be8fa9822af0dd03d513a2b`.

fathah (approved):
GitHub CI now passes on the reviewed commit: dependency audit, both TypeScript checks, the full test suite, and lint. Greptile also reports a successful review. The source findings and local validation are detailed in my earlier review. I integrated current main and resolved the onboarding documentation conflict by preserving both sections. All 44 installer regression tests, typechecks, lint, and lat check pass locally after integration. Approved for merge.

Reviewed commit: `37aa3e8da2a60531ab0b5ffaf4ac4fc731628842`.

PR #943, opened 2026-09-07
fix(config): preserve CRLF platform settings on Windows

### LINKED ISSUE #930: [Bug] Opening Hermes Desktop on Windows corrupts CRLF config.yaml: indentation-blind `enabled:` regex + CRLF off-by-one in setPlatformEnabled/readPlatformOverride deletes 18 bytes inside discord.voice_fx and restarts the gateway
## Summary

On Windows, with a CRLF `config.yaml`, simply **opening Hermes Desktop** can corrupt `config.yaml` and force-restart the gateway. The damage is always identical: exactly 18 bytes are deleted from inside the `discord:` block (`  voice_fx:\r\n    e`), leaving a dangling `nabled: false` line. YAML parsing then fails with `mapping values are not allowed here` (always at column 20), and hermes-agent silently falls back to `.env` defaults — every user override, the fallback chain and all `mcp_servers` are ignored while chat keeps "working".

It happened 5 times on my machine between 2026-08-04 and 2026-08-26 before I found the cause. Root cause is two regexes in `src/main/config.ts` that match `enabled:` at **any** indentation, combined with a CRLF off-by-one in the offset accounting. Both are still present on `main` as of `9756351e23` (2026-09-02).

## Environment

- Hermes Desktop **v0.7.6** (Windows 11 x64, `app.asar` built 2026-07-22)
- hermes-agent 0.20.5 (git install)
- `config.yaml` with CRLF line endings, containing the default `discord.voice_fx` block

## Root cause (`src/main/config.ts`)

**1. `readPlatformOverride()` is indentation-blind** (line ~1883):

```ts
const m = line.match(/^[ \t]+enabled:[ \t]*(true|false)\b/);
```

`[ \t]+` matches any indentation, so while scanning the `discord:` block it hits the **nested** `discord.voice_fx.enabled: false` (4 spaces) and reports `discord.enabled === false`. The renderer then sees "Discord disabled" while `DISCORD_BOT_TOKEN` is set, and "corrects" it by calling `set-platform-enabled('discord', true)`.

**2. `setPlatformEnabled()` has the same blind regex plus a CRLF off-by-one** (lines ~1942 and ~1979):

```ts
const enabledLineRe = new RegExp(`^([ \\t]+enabled:[ \\t]*)(true|false)\\b([ \\t]*)$`, "m");
...
const restLines = rest.split(/\r?\n/);
for (const line of restLines) {
  const lineLen = line.length + 1; // include trailing \n   <-- CRLF lines consume 2 chars
```

`split(/\r?\n/)` consumes 2 characters per CRLF line, but the offset only advances by `line.length + 1`, so the computed offset drifts 1 byte per line. `enabledLineRe` matches the nested `    enabled: false` under `voice_fx` (the 14th child line of the default `discord:` block → 13 bytes short). The `enabled=true` branch then slices out `[existingEnabledLineStart, +18)`, which lands on `  voice_fx:\r\n    e` instead of the intended line. Result:

```yaml
discord:
  ...
  max_attachment_bytes: 33554432
nabled: false          # <- was "  voice_fx:\r\n    enabled: false"
    ambient_enabled: true
```

The same IPC handler then calls `restartGateway()`, which is why the corruption always coincided with the gateway being killed.

I reproduced this byte-for-byte: replaying the two functions against a known-good config produced a file identical to the corrupted one (18,440 bytes, same 18-byte hole).

## Trigger conditions (all three required)

1. `config.yaml` uses CRLF line endings
2. the top-level `discord:` block has **no direct** `enabled:` child
3. the block contains a **nested** `enabled:` key — `voice_fx.enabled` is in the default config, so this is the out-of-the-box state on Windows

Only `discord` is affected with the default config (`slack`/`mattermost`/`matrix` have no nested `enabled`). Each config file is corrupted once; after hermes-agent rewrites the file (config migration / `save_config` restores `voice_fx`), the next Desktop launch corrupts it again.

## Steps to reproduce

1. On Windows, have `DISCORD_BOT_TOKEN` set in `.env` and a CRLF `config.yaml` containing the default `discord:` block with `voice_fx:` → `enabled: false` and no direct `discord.enabled`.
2. Note `config.yaml` size/sha256 and mtime.
3. Launch Hermes Desktop and wait ~1–2 minutes.
4. `config.yaml` shrinks by 18 bytes, `hermes gateway status` / `errors.log` show `mapping values are not allowed here ... column 20`, and the gateway has been restarted.

## Suggested fix

- In both functions, only match **direct children** of the platform block: derive the child indentation from the first indented line (or use `[ \t]{2}`) instead of `[ \t]+`.
- In `setPlatformEnabled`, keep offsets consistent with CRLF: e.g. `const lineLen = line.length + (rest.startsWith("\r\n", subBlockEndOffset + line.length) ? 2 : 1);`, or split on `\n` and strip `\r` from the matched line only.
- Fixing `readPlatformOverride` is the important half: once the read side stops misreporting Discord as disabled, the destructive "correction" write is never triggered.

As a stopgap I patched both regexes in my local `app.asar` to `[ \t]{2}`; since then Desktop launches leave `config.yaml` byte-identical (sha256 and mtime unchanged over multiple launches).

Quine-rq:
## Summary

- scope platform `enabled` overrides to direct children, so nested settings such as `discord.voice_fx.enabled` cannot disable the whole platform
- preserve the existing LF or CRLF style when inserting, replacing, or removing an override
- avoid no-op rewrites and document the config ownership contract

## Root cause

`readPlatformOverride` accepted an `enabled` key at any indentation depth. `setPlatformEnabled` then split CRLF content with `/\r?\n/` but advanced offsets by `line.length + 1`, undercounting each Windows line ending by one byte. During startup reconciliation this could splice into the following YAML node and restart the gateway with a corrupted config.

## Verification

- `npm test` — 198 files passed; 1,961 tests passed and 3 skipped
- `npm run lint`
- `npm run typecheck`
- `npx --yes lat.md check`
- 6 focused regression cases cover nested keys, byte-identical no-op behavior, direct-key removal/replacement/insertion, and empty flow blocks using CRLF fixtures

Windows manual app verification is still pending; the deterministic CRLF fixtures reproduce the reported corruption path and now pass.

Closes #930

### DIFF
--- a/lat.md/main-process.md
+++ b/lat.md/main-process.md
@@ -60,6 +60,14 @@ Renderer IPC handlers are isolated from app bootstrap so the registry can be spl
 
 Wallet and token-balance handlers sit in the same registry: `list-wallets`, `create-wallet`, `import-wallet`, `rename-wallet`, `delete-wallet` (backed by [[wallet-token-balances#Wallet Store]]) and `get-token-balances` (backed by [[wallet-token-balances#Token Balances]]).
 
+## Platform enable overrides
+
+Messaging platform availability combines credentials from `.env` with an optional direct-child `<platform>.enabled` override in `config.yaml`.
+
+[[src/main/config.ts#getPlatformEnabled]] ignores nested `enabled` keys, such as `discord.voice_fx.enabled`, so feature-specific settings cannot disable the whole platform.
+
+[[src/main/config.ts#setPlatformEnabled]] edits only the platform block's direct `enabled` child. It preserves unrelated YAML and the file's existing LF or CRLF line endings, and leaves the file byte-identical when the requested enabled state is already the default. This matters at application startup, where merely reading and reconciling a configured platform must not rewrite or corrupt a shared Windows `config.yaml`.
+
 ## Voice transcription IPC
 
 Speech-to-text IPC sends recorded desktop audio through the Hermes API server, not through the active chat model endpoint.
--- a/src/main/config.ts
+++ b/src/main/config.ts
@@ -956,7 +956,8 @@ function readTopLevelBlock(
   while (cursor < content.length) {
     const lineEnd = content.indexOf("\n", cursor);
     const lineEndExclusive = lineEnd === -1 ? content.length : lineEnd;
-    const line = content.slice(cursor, lineEndExclusive);
+    const rawLine = content.slice(cursor, lineEndExclusive);
+    const line = rawLine.endsWith("\r") ? rawLine.slice(0, -1) : rawLine;
 
     // Stop at a non-indented, non-empty line (= next top-level key).
     if (line.trim() !== "" && !/^\s/.test(line)) break;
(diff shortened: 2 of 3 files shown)


fathah (commented):

Reviewed the CRLF-preserving edits and platform enable handling against upstream configuration behavior. The implementation updates the direct platform flag without touching similarly named nested settings. Local validation: 82 focused tests passed, with 3 skipped.

Reviewed commit: `2d34591f26a4bb3a56d206f0c0c53fc5765d8094`.

fathah (approved):

Reviewed commit: `2d34591f26a4bb3a56d206f0c0c53fc5765d8094`.

PR #946, opened 2026-09-07
fix(context): honor remote model context lengths

### LINKED ISSUE #918: Context gauge: /deepseek/i fallback still 131072, and SSH mode ignores server's model.context_length (1M only shows as 131.1k)
## Summary

The context window gauge still shows **131.1k (131072)** for DeepSeek models, despite DeepSeek V4 having a **1M** window. This is present on `main` as of 2026-08-24 and in the latest release.

Root cause (two parts):

1. **The catch-all heuristic is still 131072.** In `src/renderer/src/screens/Chat/contextWindows.ts`:
   ```ts
   // DeepSeek's API models (deepseek-chat / deepseek-reasoner, V3.x) advertise ...
   [/deepseek/i, 131072],
   ```
   #588/#591 added exact mappings for some model ids (e.g. deepseek-chat/reasoner/agnes), but **any DeepSeek model id without an exact mapping still falls through to `/deepseek/i → 131072`**. Models like `deepseek-v4-flash-vision-exp` (widely used as a day-1 model id) hit this fallback.

2. **In SSH Tunnel mode, the server's `model.context_length` override is never consulted.** Gauge resolution in `Chat.tsx` / `ContextGauge.tsx` is:
   `usage.contextWindowTokens → getModelContextWindow(...) → contextWindowForModel(...)`.
   - `getModelContextWindow` (`src/main/model-discovery.ts:624`) checks `getModelContextLengthOverride(profile)` first, but that reads the **local** config; the SSH branch (`sshGetModelConfig`, `src/main/ssh-remote.ts:1325`) returns only `provider/model/base_url` — **not `model.context_length`**.
   - For subscription/OAuth providers (e.g. `opencode-go`) it then returns `null` immediately (non-discoverable), so the app lands on the heuristic.

## Repro

- Remote host `config.yaml`:
  ```yaml
  model:
    default: "deepseek-v4-flash-vision-exp"
    provider: "opencode-go"
    context_length: 1000000
  ```
- Hermes One connected via SSH Tunnel, chat transport legacy (or dashboard) — chat works fine.
- Chat footer / context popover shows: `25.1k / 131.1k tokens used` (i.e. 131072 as the window), and "19% used" — both wrong; the server actually runs with 1000000.

Same symptom in Remote (dashboard) mode when the default model is a DeepSeek V4 id via a subscription provider.

## Suggested fix

1. Bump the DeepSeek catch-all in `contextWindows.ts` to `1048576` (DeepSeek V3/V4 API models are 1M), and/or
2. Extend `sshGetModelConfig` (SSH mode) to also read `model.context_length` and feed it into the existing override path (`getModelContextLengthOverride`) — this is exactly what #588's reporter suggested ("read context_length from ~/.hermes/config.yaml") but was never wired for SSH mode.

References: #588 (closed, same bug for 65536), #591 (partial fix, exact-id mappings only).

Quine-rq:
## Summary

- Resolve the context gauge from the active SSH or Remote model config before using local provider discovery.
- Read `model.context_length` from legacy SSH config in one remote file operation.
- Expose and normalize the active context length through Dashboard model responses.
- Add a 1M heuristic for DeepSeek V4 IDs while retaining the 128K fallback for older DeepSeek aliases.
- Guard delayed remote responses so an old model's context length cannot leak after a model switch.

Fixes #918

## Why

The context-window IPC handler always called the local resolver. In SSH and Remote modes, that resolver could only see the desktop's local config, so a server-side `model.context_length: 1000000` was ignored. Subscription providers then skipped `/models` discovery and the renderer displayed the generic 131072-token fallback.

The fix keeps local behavior unchanged and reads the override from the connection that owns the active model:

- SSH legacy reads the selected remote profile directly.
- SSH Dashboard reads the Dashboard response and falls back to the SSH config when an older Dashboard omits the field.
- Remote Dashboard accepts the active row or legacy model-options field.
- Missing, invalid, or mismatched overrides retain the existing discovery/heuristic fallback.

DeepSeek's current V4 models are documented with a 1M context window in the [official model details](https://api-docs.deepseek.com/quick_start/pricing/). The specific V4 rule precedes the older 128K DeepSeek fallback.

## Validation

- `npm test` — 199 files passed; 1,961 tests passed, 3 skipped.
- Targeted transport, resolver, compatibility-patch, and heuristic tests — 25 passed.
- `npm run lint`
- `npm run typecheck`
- Prettier check for all changed source, test, and documentation files.
- `npx --yes lat.md check`
- `git diff --check`

The SSH regression uses a mocked remote file transport and verifies the named-profile path, positive override, invalid override, and single-read behavior. Dashboard tests cover both the active library row and legacy options response; no live host or credentials are required.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation-only change

## Checklist

- [x] Code follows the repository style.
- [x] Local, SSH legacy, SSH Dashboard, and Remote fallback paths were reviewed.
- [x] Full tests, lint, typecheck, formatting, and `lat.md` validation pass.
- [x] No credentials or user data are included.
- [x] AI assistance was used; the change and tests were manually reviewed.


Quine-rq:
Addressed the named-profile scoping issue in 8208a46. The compatibility library now accepts the Dashboard profile query on list/add/update/delete, stores models.json under the requested profile, and scopes load_config through the context-local Hermes home before reading contextLength. Invalid or missing named profiles fail with 400/404, while default/current behavior remains backward-compatible.

Validation includes a profile-isolation execution check and the full suite: 199 test files passed (1,961 tests passed, 3 skipped), plus lint, both typechecks, Prettier, lat.md check, Python syntax compilation, and git diff --check. Please re-review the latest commit.

fathah (commented):

Reviewed profile-specific context windows against upstream home resolution. Fixed explicit default-profile requests reading the named process profile, and verified context restoration after errors. Integrated the reviewed #945 branch and resolved the two overlapping conflicts. Local validation: the combined suite passed 1,972 tests, with 3 skipped. Merge after #945.

Reviewed commit: `a0bbf80b44393a86fd45aa00f9881b3216b80114`.

fathah (approved):

Reviewed commit: `a0bbf80b44393a86fd45aa00f9881b3216b80114`.

PR #947, opened 2026-09-08
fix(sessions): honor native archive visibility locally

### LINKED ISSUE #939: Local session list/cache ignores Hermes Agent native archived flag
## Summary

In **Hermes One 0.7.7**, using a **local connection on macOS**, sessions archived through Hermes Agent's native archive mechanism remain visible in One's normal session list.

The archive operation succeeds: the owning profile's `state.db` contains `sessions.archived = 1`, and Hermes Agent's native default listing excludes those sessions. One's local database queries and desktop session cache do not honor that flag.

## Steps to reproduce

1. Use a local connection in Hermes One and let it populate its session cache.
2. Archive an existing, inactive session in that same profile through Hermes Agent's native archive mechanism. Keep the session and its messages in the database; do not delete them.
3. Confirm that the target row has `archived = 1` and is excluded from Hermes Agent's native default listing.
4. Refresh/synchronize One's normal session list.

**Actual:** the archived session remains in One's local list/cache.

**Expected:** the normal session list excludes native-archived sessions, while retaining their underlying messages. If an archive view is offered, those sessions can be shown there explicitly.

## Verified evidence

I inspected the installed 0.7.7 application's bundled `out/main/index.js` and compared its local query with the same profile's database and `desktop/sessions.json` cache:

- Native archive flags are persisted correctly.
- The desktop cache still contains the archived IDs and does not carry the archive flag.
- Executing the installed local listing query read-only against that database returns archived rows as well.

The same code paths are present on `main` at commit `3f744975f818bbb40ed029e6b3022cd0c5ad7a24`:

1. [`listSessions()` in `src/main/sessions.ts`](https://github.com/fathah/hermes-desktop/blob/3f744975f818bbb40ed029e6b3022cd0c5ad7a24/src/main/sessions.ts#L263-L285) selects sessions without an archive predicate.
2. [`syncSessionCache()` in `src/main/session-cache.ts`](https://github.com/fathah/hermes-desktop/blob/3f744975f818bbb40ed029e6b3022cd0c5ad7a24/src/main/session-cache.ts#L125-L142) queries by `started_at` without selecting/filtering `archived`.
3. [The existing-cache refresh](https://github.com/fathah/hermes-desktop/blob/3f744975f818bbb40ed029e6b3022cd0c5ad7a24/src/main/session-cache.ts#L221-L231) checks row existence and message count, not archive state; [`listCachedSessions()`](https://github.com/fathah/hermes-desktop/blob/3f744975f818bbb40ed029e6b3022cd0c5ad7a24/src/main/session-cache.ts#L265-L271) simply slices the cache.
4. In contrast, [the remote listing path](https://github.com/fathah/hermes-desktop/blob/3f744975f818bbb40ed029e6b3022cd0c5ad7a24/src/main/remote-sessions.ts#L289-L299) explicitly requests `archived=exclude`.

## Suggested fix / regression coverage

- Apply the native archive visibility policy consistently to local direct and cached listings, including removal of already-cached archived sessions.
- Reconcile archive/unarchive changes to old sessions: a change to `archived` does not change their `started_at`, so filtering only the initial/new-session query is insufficient. Unarchived older sessions should reappear too.
- Cover cold-cache loading, warm-cache archive transitions, unarchive transitions, pagination, and profile isolation. Preserve compatibility with older databases if they lack the archive column.
- Preserve session/message data; do not use physical deletion to hide archived sessions.

Related to #412, but this is a narrower compatibility bug: honoring an **already-persisted native archive flag**, not a request to add a new archive-management UI.

No conversation contents, credentials, or private session identifiers are included in this report.

Quine-rq:
## Summary

- Honor the Agent's native `archived` flag in local direct session lists, filtering before pagination while keeping older databases without the column readable.
- Reconcile the complete visible metadata set during cache sync so already-cached archives disappear and restored old sessions reappear without changing `started_at`.
- Allow an empty local quiet refresh to remove the last archived row from the Sessions modal; preserve the SSH/HTTP transient-empty guard and reject stale profile responses.
- Retain messages and linked project folders. Keep the last good cache when database reads fail, and keep the fast cache-read path DB-free.

Fixes #939

## Why

The direct query ignored `archived`, while the cache used session creation time as an incremental cursor and checked only existence/message count for older rows. Filtering only new rows would leave warm-cache archives visible and miss restores of old sessions.

The sync now reads visible metadata in one query and reuses cached titles. It reads message content only when a newly visible row needs a generated title. This replaces the previous creation-time query plus batched refresh of older IDs; it does not scan all message bodies.

This is compatibility with an already-persisted native archive flag, not a new archive-management UI. Search/history behavior and SSH/HTTP listing implementations are unchanged.

## Validation

- `npm test`: 198 test files passed; 1,967 tests passed, 3 skipped.
- `npm run typecheck`, `npm run lint`, Prettier for all changed files, `npx --yes lat.md check`, and `git diff --check`.
- `electron-vite build` passed during local implementation.
- Nine new real-SQLite regression tests cover cold/warm caches, old and never-cached restores, pagination, same-ID profile isolation, legacy schema upgrades, failed-read recovery, and retained messages/folders. Tests execute production SQL through Node 22 SQLite, avoiding a dependency on the Electron native-module ABI.
- Three new renderer tests cover the last local row disappearing/reappearing, failed connection-check recovery, and a delayed empty response after a profile switch. Existing SSH empty-refresh protection remains covered.
- Isolated macOS/Electron 44.1.1 smoke test using the actual built main process, better-sqlite3, IPC, preload, sidebar, and modal: toggle the fixture's native archive flag; both lists and disk cache empty and recover, while message count stays unchanged. Only the installation gate was bypassed; no real model/gateway or private user data was used.

Full end-to-end verification through a real installed Hermes Agent's archive command, and Windows/Linux app checks, are still pending. The initial DB-free cache paint may remain stale until the next successful sync; existing refresh intervals are unchanged.

AI assistance was used for implementation and validation. No credentials or private conversation data are included.

### DIFF
--- a/lat.md/sidebar-navigation.md
+++ b/lat.md/sidebar-navigation.md
@@ -22,6 +22,18 @@ Session titles in the inline list are constrained to the sidebar width and trunc
 
 The native sidebar scrollbar is hidden to avoid layout shifts. [[src/renderer/src/screens/Layout/Layout.tsx#Layout]] measures the chat scroll container and renders an absolutely positioned overlay thumb only while the user is scrolling, so showing or hiding the scrollbar never changes row width.
 
+## Native archive visibility
+
+Local lists follow the Agent's native archive flag. Archiving hides a conversation without deleting its messages or linked project folder; restoring it makes it visible on the next sync.
+
+[[src/main/db.ts#sessionVisibilityPredicate]] detects whether the selected profile's database has an `archived` column. [[src/main/sessions.ts#listSessions]] filters `archived = 0` before pagination, while legacy databases without that column remain readable.
+
+[[src/main/session-cache.ts#syncSessionCache]] reconciles the complete visible metadata set rather than using `started_at` as a change cursor: archive and restore do not change creation time. Cached titles are reused, and message bodies are read only to generate missing titles for newly visible rows. Successful sync replaces the cached membership, including an empty set; unavailable databases or failed reads retain the last good cache until retry. [[src/main/session-cache.ts#listCachedSessions]] remains DB-free, so an initial cached paint can be stale until sync finishes. Message history and search are unchanged.
+
+[[src/renderer/src/screens/Sessions/Sessions.tsx]] accepts an empty quiet-refresh result for the explicitly selected local connection, allowing the last archived row to disappear. Network-backed lists retain their transient-empty guard. Failed connection checks retain visible rows, and request identity prevents a delayed empty result from clearing a newer profile's list.
+
+[[tests/session-archive.test.ts]] executes real SQLite queries for cold and warm caches, archive/restore without timestamp changes, pagination, equal IDs across profiles, legacy schema upgrades, read failure and recovery, and retained history/project folders. [[src/renderer/src/screens/Sessions/Sessions.test.tsx]] covers the last local row disappearing and returning, failed refresh recovery, and stale responses after profile switches.
+
 ## Project grouping
 
 Workspace-linked conversations are grouped under project rows so repository chats stay together without hiding ordinary chats.
--- a/src/main/db.ts
+++ b/src/main/db.ts
@@ -6,6 +6,16 @@ let cachedDb: Database.Database | null = null;
 let cachedDbPath: string | null = null;
 let cachedDbReadonly: boolean | null = null;
 
+/** Older Agent databases predate native archiving; keep their lists readable. */
+export function sessionVisibilityPredicate(db: Database.Database): string {
+  const columns = db.prepare("PRAGMA table_info(sessions)").all() as Array<{
+    name: string;
+  }>;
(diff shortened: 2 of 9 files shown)


fathah (commented):

Reviewed native archive/restore against upstream soft-archive semantics, including filtering before pagination, legacy schemas, complete cache reconciliation, profile isolation, and recovery from failed reads. Local validation: 60 focused tests passed, including real SQLite and renderer coverage.

Reviewed commit: `d6d8f30c21b1effaaa5858f104e0fd8e0aec5cc5`.

fathah (approved):

Reviewed commit: `d6d8f30c21b1effaaa5858f104e0fd8e0aec5cc5`.

PR #960, opened 2026-09-15
fix(chat): render and answer dashboard clarification cards

### LINKED ISSUE #954: ClarifyCard is unreachable for dashboard/WebSocket sessions — clarify renders as a plain text bubble
# ClarifyCard is unreachable for dashboard/WebSocket sessions — clarify renders as a plain text bubble

## Summary

`clarify.request` has **two independent renderer paths**, and only the IPC one builds the interactive card:

| Transport | Handler | Result |
|---|---|---|
| Local IPC (`useChatIPC`) | `window.hermesAPI.onClarifyRequest` | `ClarifyMessage` → **`ClarifyCard` with buttons** |
| Dashboard / WebSocket (`dashboardEventAdapter`) | `case "clarify.request"` → `appendClarifyRequest` | `ChatBubbleMessage` → **plain text, no buttons** |

When the desktop app talks to the agent over the dashboard WebSocket transport (`/api/ws`), every `clarify` prompt arrives as a numbered-list text bubble. The choice buttons, the "Other" textarea, and the "skip — let Hermes decide" control added in #604 are never rendered, because that path never produces a `kind: "clarify"` message.

The feature is fully implemented and well tested — it is simply not wired to the second transport.

## Evidence (current `main`)

**`src/renderer/src/screens/Chat/dashboardEventAdapter.ts:154`** — the WS path builds a bubble, not a clarify message:

```ts
const bubble: ChatBubbleMessage = {
  id,
  role: "agent",
  content,          // question + "1. …\n2. …"
  pending: false,
  localOnly: true,
};
```

Choices are flattened into markdown text at lines 146–151 instead of being passed through as data:

```ts
const content =
  choices.length > 0
    ? `${question}\n\n${choices.map((choice, index) => `${index + 1}. ${choice}`).join("\n")}`
    : question;
```

**`src/renderer/src/screens/Chat/hooks/useChatIPC.ts:245`** — the IPC path, for contrast, emits the card message:

```ts
const cleanupClarify = window.hermesAPI.onClarifyRequest((eventRunId, req) => {
  if (!eventMatchesRun(eventRunId, runId)) return;
  ...
  prev.some((m) => m.kind === "clarify" && m.requestId === req.requestId)
```

`ClarifyCard` itself is present in the shipped bundle and renders correctly — it is just never instantiated on the WS path.

## Complication: the response path is transport-specific too

This is not a one-line fix, which may be why it was missed. The two paths also *answer* differently:

- **`ClarifyCard.tsx:40`** replies over IPC:
  ```ts
  const ok = await window.hermesAPI.respondClarify(msg.requestId, answer);
  ```
  In `src/main`, `respondClarify` → `resolvePendingClarify(requestId, …)`, which only finds a resolver if `registerPendingClarify` ran — and that happens **only** in the IPC `clarify.request` branch.

- **The WS path** replies over the gateway client, using the next composer message as the answer:
  ```ts
  const pendingClarifyRequestId = pendingClarifyRequestIdRef.current;
  if (pendingClarifyRequestId) {
    await client.request("clarify.respond", { request_id: pendingClarifyRequestId, answer: text });
  }
  ```

So simply emitting `kind: "clarify"` from `appendClarifyRequest` would render the card but break its submit: `resolvePendingClarify` returns `false` for an unknown id, and the card shows its error state. A fix needs the card to answer through the dashboard gateway client when the request arrived over WS.

The silver lining: answering **does** work today on the WS path — typing a normal chat message resolves the pending clarify. Only the buttons are missing.

## Reproduction

1. Run the agent with the API server enabled (`platforms.api_server.enabled: true`, port 8642).
2. Connect Hermes One to it so the session uses the dashboard WebSocket transport (gateway log shows `tui_gateway.ws: ws accepted peer=…`).
3. Have the agent call the `clarify` tool with choices.

**Expected:** an inline card with clickable choice buttons.
**Actual:** a plain agent text bubble containing the question and a numbered list; no buttons, nothing clickable.

## Environment

- Hermes One **0.7.7** (`hermes-desktop_0.7.7`, Linux)
- Hermes Agent **0.20.5**
- Arch Linux
- Transport: dashboard WebSocket (`/api/ws`), agent on `127.0.0.1:8642`

## Notes

Worth distinguishing from a separate, agent-side issue I hit while diagnosing this: agent ≥ 0.20.5 emitted single questions in the batch wire shape `{questions: [...]}`, while the desktop reads flat `payload.question` / `payload.choices`. That one is fixed upstream by NousResearch/hermes-agent#96682 ("unwrap one-question clarify batches"), and with that fix applied the payload reaching the app is flat and correct — the card still does not appear, which is what this report is about.

## Suggested direction

Make the transport an implementation detail of the card rather than a fork in the renderer:

1. Have `dashboardEventAdapter` produce a `ClarifyMessage` (`kind: "clarify"`, with `choices` preserved as an array) instead of pre-rendering the list into markdown.
2. Give `ClarifyCard` a submit strategy — IPC `respondClarify` for locally spawned backends, `client.request("clarify.respond", …)` for dashboard/WS sessions — so the existing `pendingClarifyRequestIdRef` bookkeeping stays the WS fallback rather than the only path.

Quine-rq:
Dashboard clarification requests currently appear as numbered text. This change renders interactive question cards and sends choices, free text and skips through the dashboard connection that owns the request. Composer replies use the same path.

Pending questions are scoped to their request and runtime session. Failed replies remain retryable; duplicate submissions are blocked; expired replies cannot appear successful. Connection changes, disconnects, completion and abort invalidate stale cards. Replayed questions preserve newer pending questions and turns whose answers are in flight. Agent `clarify.expire` notifications disable the matching card while tracking the turn that resumes after timeout.

Fixes #954.

Validation on the integrated branch:
- All 2,308 tests passed across 221 files, including regression tests for stale replays, in-flight replays and gateway expiration.
- Node and renderer type checks, ESLint on all changed TypeScript files, production electron-vite build, `lat check` and `git diff --check` passed.
- Cross-checked the protocol against sibling Hermes Agent at `503d863fcd2cbfc0be5a6d6c536fae2e98aa4204` and inspected Hermes One Backend at `343c725911b32b089ff88e768ab421ac4937a120`.
- An isolated loopback WebSocket smoke test used the real desktop gateway client and the Agent's actual `_block` / `_respond` functions: choices, free text, empty skip, timeout notification and late-response expiration passed.

Scope is the existing single-question `question` / `choices` protocol. Multi-question batches and a live model-backed Agent session were not tested in this change.


fathah (approved):
Thanks @Quine-rq for fixing the dashboard clarification flow and for addressing the disconnect cleanup finding.

Reviewed the implementation against the sibling Hermes Agent gateway protocol and checked Hermes One Backend's role. Integrated current main without conflicts, then fixed three additional lifecycle gaps: stale question replays disabling a newer card, in-flight replays clearing the resumed turn, and unhandled `clarify.expire` notifications. Added regression coverage and updated the architecture/test documentation.

Validation on 945ae1a1ab2d817698c97d8384ffdfefea85efc0: 2,308 tests passed across 221 files; type checks, changed-file ESLint, production build, documentation validation and diff checks passed. An isolated real-WebSocket smoke test used the desktop client and the upstream Agent's actual blocking/reply functions for choice, free-text, skip, timeout and expired late replies. CI is green and refreshed Greptile is 5/5 on this exact commit, with no outstanding review threads.

The existing single-question protocol is ready to merge. Multi-question batches and a live model-backed Agent session remain outside this validation.

